PR #56, opened 2013-10-07
Proposed fix for BitBucket Issue #157

hydrogen18:
I have taken a stab at fixing bit bucket issue #157. A couple points

1. This code uses the 'print' function directly. I'm not sure if that is desirable, but I opted not to factor that out at this time.
2. Issue #157 was about unix domain sockets, I believe I have fixed it. I opted to try and make backdoor compatible with IPv6 at the same time.
3. I had to add a catch for BaseException in the call to InteractiveConsole.interact in order to make unit tests pass. The exception is printed when finalize is called. There may be a better way and I welcome any suggestions.
4. I probably broke some coding standard in this. I don't think I let any tabs slip in at least.

### DIFF
--- a/eventlet/backdoor.py
+++ b/eventlet/backdoor.py
@@ -2,6 +2,7 @@ import socket
 import sys
 import errno
 from code import InteractiveConsole
+import traceback
 
 import eventlet
 from eventlet import hubs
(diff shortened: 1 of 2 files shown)


temoto:
I tried to pull this patch again https://github.com/eventlet/eventlet/commit/e536cfaea7616b5e76b311bc0ae87858f3409422 it reliably breaks all CI tests, I guess it's because BaseException again. I tried to change it to catch SystemExit and Exception separately, no luck. Will look again after releasing v0.17.

jstasiak:
Well this is old. :)

Thanks for the patch @hydrogen18, I let myself change the commit message and refactor the patch (it works without the try block you mentioned), please let me know if this is ok and I'll merge it if so. The commit: 24d283c

jstasiak:
Thanks for commenting on 24d283c @hydrogen18, I've merged it into master just now, thank you for the patch once again :)

PR #65, opened 2013-12-02
fix interpreter shutdown message

neilrahilly:
Sometimes, during shutdown, this code is executed when deque no longer exists, which produces an unhelpful error message.


temoto:
This issue is most likely already fixed. Please reopen if it persists.

PR #85, opened 2014-04-05
make sure we return results during killall

davidszotten:
since we are about to delete the _rspq


temoto:
Merged in https://github.com/eventlet/eventlet/commit/52d42dd741fc8da7f1b1ac525ff6c9f2f5739ee7 thank you very much.

PR #115, opened 2014-08-05
Fix AttributeError GreenPipe._rbuf

rutube:
In Python3 GreenPipe is a child of socket.SocketIO class, that doesn't have _rbuf attribute. That is why any attempt to use tell or seek on it is raising AttributeError.

### DIFF
--- a/eventlet/greenio.py
+++ b/eventlet/greenio.py
@@ -505,27 +505,27 @@ class GreenPipe(_fileobject):
         len = self._get_readahead_len()
         if len > 0:
             self.read(len)
+    if six.PY2:
+        def tell(self):
+            self.flush()
+            try:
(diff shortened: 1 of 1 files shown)


jstasiak:
@temoto: This is still an issue

@rutube: I'm in process of changing that code extensively so it uses buffering and decoding/encoding infrastructure from Python 3 io module and we won't need this workaround but thank you for the contribution, I'll leave this open until it's fixed.

temoto:
@jstasiak will you finish the change before v0.17 at 2014-02-20? By the way, thanks for merging other pull requests.

jstasiak:
No problem. Yeah I should have some test-passing code by then.

jstasiak:
This is fixed in master so I'll close.

PR #116, opened 2014-08-08
add separate keepalive_timeout to wsgi server

btmorex:
Right now, connections are kept alive for socket_timeout seconds (or forever if socket_timeout is None). This patch adds a separate option which is useful if you want idle connections to be kept around for a long time, but still have a low socket_timeout.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -6,13 +6,13 @@ import traceback
 import types
 import warnings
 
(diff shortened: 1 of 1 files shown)


briancline:
This is great and would be **extremely** useful to have. Unfortunately it's been a very long time since this was opened. Is this only waiting on a rebase and a unit test?

PR #117, opened 2014-08-08
add option to disable nagle algorithm to wsgi server

btmorex:
This patch adds and option to disable the nagle  algorithm in the wsgi server. It reuses functionality provided by StreamRequestHandler for both setting TCP_NODELAY and setting a socket timeout.


temoto:
Since v0.18.4 we have a better [1] option, TCP_QUICKACK enabled by default. I hope this also satisfies what your patch intended to do. Please reopen if not.

https://news.ycombinator.com/item?id=10607422

Issue #136, opened 2014-09-05
Multiple  coroutine requests semaphores  failure


ljz888666555:
here is test code

``` python
# !/usr/bin/env python

import eventlet
eventlet.monkey_patch(socket=True, select=True, thread=True)

g_mutex = eventlet.semaphore.Semaphore()

def func1(no):
    while True:
        with g_mutex:
            print "i am running  no = %d" % no
            eventlet.sleep(1)


def func2(no):
    while True:
        with g_mutex:
            print "i have no chance to run no = %d" % no
            eventlet.sleep(1)

def test():
    my_pool = eventlet.greenpool.GreenPool()
    my_pool.spawn_n(func1, no=1)
    my_pool.spawn_n(func2, no=2)
    while True:
        eventlet.sleep(100)

if __name__ == "__main__":
    test() 
```

temoto:
This issue is fixed long ago. Please reopen if you feel it persists.

PR #163, opened 2014-11-09
Don't hog a semaphore if someone else is waiting for it

shauns:
https://github.com/eventlet/eventlet/issues/136


temoto:
Thank you, merged in https://github.com/eventlet/eventlet/commit/bab11168093d63db8cd11b47321264b3ae91411c

Issue #114, opened 2014-07-31
 error:1409F07F:SSL routines:SSL3_WRITE_PENDING:bad write retry


remimin:
python 2.6.6
eventlet 0.13 version
```
 self._socket = eventlet.wrap_ssl(self._socket, **ssl_kwargs) 
File "/usr/lib/python2.6/site-packages/eventlet/wsgi.py", line 399, in handle_one_response
    write(''.join(towrite))
  File "/usr/lib/python2.6/site-packages/eventlet/wsgi.py", line 349, in write
    _writelines(towrite)
  File "/usr/lib64/python2.6/socket.py", line 334, in writelines
    self.flush()
  File "/usr/lib64/python2.6/socket.py", line 303, in flush
    self._sock.sendall(buffer(data, write_offset, buffer_size))
  File "/usr/lib/python2.6/site-packages/eventlet/green/ssl.py", line 131, in sendall
    v = self.send(data[count:])
  File "/usr/lib/python2.6/site-packages/eventlet/green/ssl.py", line 107, in send
    super(GreenSSLSocket, self).send, data, flags)
  File "/usr/lib/python2.6/site-packages/eventlet/green/ssl.py", line 77, in _call_trampolining
    return func(*a, **kw)
  File "/usr/lib64/python2.6/ssl.py", line 174, in send
    v = self._sslobj.write(data)
SSLError: [Errno 1] _ssl.c:1276: error:1409F07F:SSL routines:SSL3_WRITE_PENDING:bad write retry
```

liorn:
To reproduce:
1. Create 10 SSL sockets
2. eventlet.spawn 10 functions which continuously send large amount of data to that socket: `s.send(large_string)`, where `len(large_string)` is 1MB for example

Reproduces easily on my machine.

liorn:
I found the bug. Will submit a pull request soon

jstasiak:
@remimin @liorn This is an interesting issue, I spent a chunk of the day trying to reproduce the issue and debugging Python binary to make sure the only piece of code that can cause that error is being executed and it was, but the conditions inside weren't being met.

My system: OS X 10.9.5, OpenSSL 0.9.9 (I think it's 0.9.8y but I don't see the version being specified formally anywhere), Python 2.7.9 and 2.6.8, eventlet 0.15.1 and 0.16.1.

[...]

PR #184, opened 2015-01-04
Make sure SSL retries are done using the exact same data buffer

liorn:
This is a fix for the bug described in https://github.com/eventlet/eventlet/issues/114

OpenSSL requires that read/write retries are made using the exact same buffer.
When using `v = self.send(data[count:])` over multiple retries, the buffer is recreated - and OpenSSL detects it as a bad retry.


liorn:
@temoto can you pull this one in? Thanks!

temoto:
@liorn thank you very much, it's merged in master. pypi release is planned this month, in case you need it.

Issue #185, opened 2015-01-05
threading.RLock (and threading.Condition) doesn't work on Python 3.2+ with monkey-patching


haypo:
The threading.RLock class was reimplemented in C in Python 3.2. threading.RLock doesn't use _thread.get_ident() or threading.get_ident(), but it calls directly the C function. Because of that, RLock doesn't work ("as expected") with eventlet monkey-patching. Example:

```
#!python
import eventlet
eventlet.monkey_patch()

import threading
#threading.RLock = threading._PyRLock

def func(c):
    with c:
        print("thread: owned?", c._is_owned())
        c.notify()
        print("thread: owned?", c._is_owned())

c = threading.Condition()
print("LOCK", threading.RLock)
with c:
    t = threading.Thread(target=func, args=(c,))
    t.start()
    print("main: owned?", c._is_owned())
    c.wait()
    print("main: owned?", c._is_owned())
```

The code hangs with eventlet 0.16 on Python 3.5 (and probably also Python 3.2, 3.3 and 3.4).

I tried to reuse the Python implementation of the RLock class (threading._PyRLock), but in this case c.notify() raises: RuntimeError("cannot notify on un-acquired lock").

I will work on a fix.

Note: I'm working on porting the Oslo Messaging project to Python 3. The initial issue came from a threading.Condition object, threading.Condition uses internally a threading.RLock.

haypo:
Oslo Messaging now works well on Python 3 with eventlet. I don't recall how the issue was fixed, but it now works well :-) So I close this old issue.

PR #187, opened 2015-01-06
Fix threading.Condition with monkey-patching on Python 3.3 and newer

haypo:
It fixes monkey-patching threading.Condition on Python 3.3 and newer:
fix issue #185.

### DIFF
--- a/eventlet/semaphore.py
+++ b/eventlet/semaphore.py
@@ -78,9 +78,18 @@ class Semaphore(object):
         When invoked with blocking set to false, do not block. If a call without
         an argument would block, return false immediately; otherwise, do the
         same thing as when called without arguments, and return true.
+
+        Timeout value must be strictly positive.
         """
-        if not blocking and timeout is not None:
-            raise ValueError("can't specify timeout for non-blocking acquire")
+        if timeout == -1:
+            timeout = None
+        if timeout is not None and timeout < 0:
+            raise ValueError("timeout value must be strictly positive")
+        if not blocking:
+            if timeout is not None:
+                raise ValueError("can't specify timeout "
+                                 "for non-blocking acquire")
+            timeout = 0
         if not blocking and self.locked():
             return False
 
--- a/eventlet/green/thread.py
+++ b/eventlet/green/thread.py
@@ -3,6 +3,7 @@ from eventlet.support.six.moves import _thread as __thread
 from eventlet.support import greenlets as greenlet, six
 from eventlet import greenthread
 from eventlet.semaphore import Semaphore as LockType
+import sys
 
 
 __patched__ = ['get_ident', 'start_new_thread', 'start_new', 'allocate_lock',
@@ -43,6 +44,19 @@ def __thread_body(func, args, kwargs):
 
 
 def start_new_thread(function, args=(), kwargs=None):
+    if (sys.version_info >= (3, 4)
+    and getattr(function, '__module__', '') == 'threading'
+    and hasattr(function, '__self__')):
+        # Since Python 3.4, threading.Thread uses an internal lock
+        # automatically released when the python thread state is deleted.
+        # With monkey patching, eventlet uses green threads without python
+        # thread state, so the lock is not automatically released.
+        #
+        # Disable the thread state lock to avoid dead locks.
+        thread = function.__self__
+        thread._set_tstate_lock = lambda: None
+        thread._wait_for_tstate_lock = lambda *args, **kw: None
+
     kwargs = kwargs or {}
     g = greenthread.spawn_n(__thread_body, function, args, kwargs)
     return get_ident(g)
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -304,6 +304,13 @@ def monkey_patch(**on):
         # importlib must use real thread locks, not eventlet.Semaphore
         importlib._bootstrap._thread = thread
 
+        # Issue #185: Since Python 3.3, threading.RLock is implemented in C and
+        # so call a C function to get the thread identifier, instead of calling
+        # threading.get_ident(). Force the Python implementation of RLock which
+        # calls threading.get_ident() and so is compatible with eventlet.
+        import threading
+        threading.RLock = threading._PyRLock
+
 
 def is_monkey_patched(module):
     """Returns True if the given module is monkeypatched currently, False if
--- a/tests/isolated/patcher_threading_condition.py
+++ b/tests/isolated/patcher_threading_condition.py
@@ -1,0 +1,23 @@
+# Issue #185: test threading.Condition with monkey-patching
+import eventlet
+
+# no standard tests in this file, ignore
+__test__ = False
+
+
+if __name__ == '__main__':
+    eventlet.monkey_patch()
+
+    import threading
+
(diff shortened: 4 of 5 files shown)


haypo:
Is it ok to change Semaphore.acquire() behaviour for negative timeout? Any issue with the backward compatibility? The idea is to have a behaviour to the Python stdlib.

haypo:
Ping? Is someone available to review this change? Is it ok?

jstasiak:
@haypo I'll review this tomorrow, sorry for the wait

jstasiak:
@haypo The negative timeout value seems to be working already, do you have a test case showing a failure?

haypo:
> @haypo The negative timeout value seems to be working already, do you have a test case showing a failure?

My first commit changes the behaviour of eventlet.semaphore.Semaphore.acquire() when you pass a negative timeout. Before, Semaphore.acquire(timeout=-1) was non-blocking. After, Semaphore.acquire(timeout=-1) becomes blocking, same behaviour than Python 3 lock.

In Python 2, threading.Lock.acquire() has no timeout parameter. So RLock only calls acquire() (on its internal lock) with one parameter: blocking.

In Python 3, threading.Lock.acquire() got a new timeout parameter: def acquire(blocking=True, timeout=None). RLock calls aquire() (on its internal lock) with two parameters: blocking and timeout.

In Python 3, threading.Lock.acquire(timeout=-1) blocks, whereas eventlet.semaphore.Semaphore.acquire(timeout=-1) is currently non-blocking.

eventlet monkey-patching is not supposed to modify the behaviour of patched functions. I wasn't the case in Python 2, since Lock.acquire() had no timeout parameter. It's now the case, since Python 3 also uses the second timeout parameter.

In short, my changes fix two different Python 3 bugs:
- fix monkey-patching on threading.Lock to have the same behaviour on Lock.acquire(timeout=-1)
- fix RLock, it didn't work with monkey-patching on Python 3.3+

haypo:
FYI I failed to find a call to Semaphore.acquire() which uses the timeout parameter in eventlet source code. In eventlet, the timeout doesn't look to be used.

The timeout parameter is documented:
http://eventlet.net/doc/modules/semaphore.html#eventlet.semaphore.Semaphore.acquire

The behaviour on negative timeout is not documented.

haypo:
Ping?

temoto:
@haypo I've rebased it on recent master, now `tests.patcher_test:Threading.test_threading` fails with timeout. Could you look at it? https://github.com/eventlet/eventlet/tree/semaphore_timeout

haypo:
> @haypo I've rebased it on recent master,

Cool!

> now tests.patcher_test:Threading.test_threading fails with timeout. Could you look at it? https://github.com/eventlet/eventlet/tree/semaphore_timeout

It's not a regression of recent changes in eventlet. I'm not sure, but I guess that I used Python 3.3 to work on my patch, whereas Python 3.4 requires extra changes. Python 3.4 adds a new threading.Thread._tstate_lock attribute which doesn't work with monkey-patching, because it relies on private C code which releases the new lock at thread exit (when its Python thread state is destroyed).

Well, see the pull request https://github.com/eventlet/eventlet/pull/211 for the details, I proposed a fix.

haypo:
The changeset "Disable the thread state lock" fixes the threading support on Python 3.4 when eventlet is used with monkey-patching.

Python 3.4 introduces threading.Thread._tstate_lock to fix a race condition, see:
- https://hg.python.org/cpython/rev/d52b68edbca6
- http://bugs.python.org/issue18808

I'm not really proud of this change: patching a Thread instance after its creation is not the most elegant fix.

I tried to create a Thread subclass in eventlet.green.threading, but I got a lot of new issues. If the subclass is declared in eventlet.green.threading, the parent Thread class uses functions of the original threading module, instead of using patched functions by eventlet. It's unclear to me how symbols are added and patched in modules. I noticed that the threading module of the standard library is imported multiple times. The implementation of monkey-patching is really complex :-/

Currently, the best solution is to patch Thread instance in the patched thread.start_new_thread() function. The problem is to retrieve the instance. My heuristic checks if the wrapped function is the bounded Thread.run() method, and I retrieve the instance from function.__self__.

I'm not yet sure that I really understood the problem. At least, "tox -e py34-selects" pass.

haypo:
I rewrite the first commit: in fact, we don't need to support negative timeout, we only need to accept timeout=-1 for Semaphore.accept(). If the timeout value is None or -1, the timeout is ignored, only the blocking parameter is used:
- acquire(timeout=None)/acquire(timeout=-1) is the same than acquire(blocking=True): blocking
- acquire(blocking=False, timeout=None)/acquire(blocking=False, timeout=-1) is the same than acquire(blocking=False): non-blocking

eventlet/semaphore.py
```diff
@@ -78,9 +78,18 @@ def acquire(self, blocking=True, timeout=None):
...
         same thing as when called without arguments, and return true.
+
+        Timeout value must be strictly positive.
         """
-        if not blocking and timeout is not None:
-            raise ValueError("can't specify timeout for non-blocking acquire")
+        if timeout == -1:
+            timeout = None
+        if timeout is not None and timeout < 0:
+            raise ValueError("timeout value must be strictly positive")
```
temoto: Exception says "strictly positive" ( x > 0 ), while code allows timeout=0 just fine.

haypo: > Exception says "strictly positive" ( x > 0 ), while code allows timeout=0 just fine.

Hum... yes... I copied code from the C implementation of threading.Lock, to try to be as close as possible to threading.Lock.

Example:

```
$ python3
Python 3.4.1 (default, Nov  3 2014, 14:38:10) 
>>> import threading
>>> l=threading.Lock()
>>> l.acquire(timeout=-2)
...
ValueError: timeout value must be strictly positive
>>> l.acquire(timeout=0)
True
>>> l.acquire(timeout=0)
False
```

haypo:
temodo wrote "Exception says "strictly positive" ( x > 0 ), while code allows timeout=0 just fine." but I copied the error message from Python.

Do you expect that I change my patch, or are you ok to mimick Python lock?

temoto:
Quite surprised that Python code has this code/comment mismatch. Yeah, leave it same as stdlib.

haypo:
> Quite surprised that Python code has this code/comment mismatch. Yeah, leave it same as stdlib.

Oh, so did you review my patch? Does it look ok? Should I change something?

temoto:
LGTM, running last round of tests now.

temoto:
I took the liberty of sqashing PEP-8 fixes into your commits.

haypo:
> I took the liberty of sqashing PEP-8 fixes into your commits.

No problem. I only care of having eventlet running on python 3.3 ;-)

temoto:
It's merged into master 390e71d89883649dd55fbb70dc8686bf99e2a2e2 thank you!
Please test if it works and drop a line.

jstasiak:
@hardys @temoto Can this issue be closed now?

haypo:
> @hardys @temoto Can this issue be closed now?

A change fixing this issue was merged, so yes, the issue can be closed.

jstasiak:
Thanks @haypo!

Issue #145, opened 2014-10-10
ssl error


shivaenigma:
I am using latest git code from python3.3 branch on ubuntu. Let me know if you need any more information

```
  File "/usr/lib/python3.4/urllib/request.py", line 1232, in do_open
    h.request(req.get_method(), req.selector, req.data, headers)
  File "/usr/lib/python3.4/http/client.py", line 1065, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.4/http/client.py", line 1103, in _send_request
    self.endheaders(body)
  File "/usr/lib/python3.4/http/client.py", line 1061, in endheaders
    self._send_output(message_body)
  File "/usr/lib/python3.4/http/client.py", line 906, in _send_output
    self.send(msg)
  File "/usr/lib/python3.4/http/client.py", line 841, in send
    self.connect()
  File "/usr/lib/python3.4/http/client.py", line 1205, in connect
    server_hostname=server_hostname)
  File "/usr/lib/python3.4/ssl.py", line 364, in wrap_socket
    _context=self)
  File "/usr/lib/python3.4/ssl.py", line 578, in __init__
    self.do_handshake()
  File "/usr/lib/python3.4/ssl.py", line 805, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLEOFError: EOF occurred in violation of protocol (_ssl.c:598)
```

temoto:
@jstasiak please comment if this issue is supposed to be fixed with #199 merged?

jstasiak:
To the best of my knowledge, yes.

temoto:
@shivaenigma please, try again with recent master.

bitnumus:
@temoto I will test it this week and let you know

bitnumus:
@temoto I'm now receiving an import error for urllib, what changed? i opened an issue regarding this also.

bitnumus:
@temoto recent master seems to have broken our current code due to import errors with urlopen, so i'll have to fix this and take another look.

However, i wrote another script to test "requests" lib with eventlet and if i try to use a full poll of 1000 requests i get the same error as in this issue. Is it likely there will be full support for requests library in the future rather than through monkey_patch() or however its working at the moment? urllib imho really is an awful library to develop with, requests just seems far cleaner.

temoto:
@bitnumus do you mean "ssl.SSLEOFError: EOF occurred in violation of protocol" ? Can you show the code?

bitnumus:
[...]

Issue #108, opened 2014-07-11
Python3 ImportError: No module named 'urllib2'


temoto:
```
======================================================================
ERROR: Failure: ImportError (No module named 'urllib2')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/eventlet/eventlet/.tox/py33selects/lib/python3.3/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/home/travis/build/eventlet/eventlet/.tox/py33selects/lib/python3.3/site-packages/nose/loader.py", line 411, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/build/eventlet/eventlet/.tox/py33selects/lib/python3.3/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/build/eventlet/eventlet/.tox/py33selects/lib/python3.3/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/eventlet/eventlet/.tox/py33selects/lib/python3.3/imp.py", line 180, in load_module
    return load_source(name, filename, file)
  File "/home/travis/build/eventlet/eventlet/.tox/py33selects/lib/python3.3/imp.py", line 119, in load_source
    _LoadSourceCompatibility(name, pathname, file).load_module(name)
  File "<frozen importlib._bootstrap>", line 584, in _check_name_wrapper
  File "<frozen importlib._bootstrap>", line 1022, in load_module
  File "<frozen importlib._bootstrap>", line 1003, in load_module
  File "<frozen importlib._bootstrap>", line 560, in module_for_loader_wrapper
  File "<frozen importlib._bootstrap>", line 868, in _load_module
  File "<frozen importlib._bootstrap>", line 313, in _call_with_frames_removed
  File "/home/travis/build/eventlet/eventlet/tests/test__greenness.py", line 7, in <module>
    from eventlet.green import urllib2, BaseHTTPServer
  File "/home/travis/build/eventlet/eventlet/eventlet/green/urllib2.py", line 13, in <module>
    ('urllib', urllib))
  File "/home/travis/build/eventlet/eventlet/eventlet/patcher.py", line 88, in inject
    module = __import__(module_name, {}, {}, module_name.split('.')[:-1])
ImportError: No module named 'urllib2'
```

PR #199, opened 2015-02-12
Python 3 compatibility improvement + some random bits

jstasiak:
Although there are still some Python 3 related issues not covered by the test suite it's nice to see the build green.

CC GH #108 #6 #145

@temoto: requesting comments, I'm aware of some of the code being suboptimal (code repetition and the way it's split into modules) but I figure it's good to get it out of the way and have the chance to refactor later


eventlet/green/subprocess.py
```diff
@@ -1,15 +1,22 @@
 import errno
 import time
```
temoto: Remove original time, so you don't have to rename green one.

eventlet/greenio.py
```diff
@@ -1,693 +1,14 @@
...
-            return None
-        raise
-
-
-if sys.platform[:3] == "win":
-    # winsock sometimes throws ENOTCONN
-    SOCKET_BLOCKING = set((errno.EAGAIN, errno.EWOULDBLOCK,))
-    SOCKET_CLOSED = set((errno.ECONNRESET, errno.ENOTCONN, errno.ESHUTDOWN))
+if six.PY2:
+    from eventlet._py2greenio import _fileobject  # noqa
```
temoto: Intuitively, `import *` should be enough, without `_fileobject`. Is there any reason why `_fileobject` is not included in `_pyXgreenio.__all__`?

tests/wsgi_test.py
```diff
@@ -1743,7 +1743,9 @@ def handler(*args):
             fd = self.connect()
             fd.sendall(req.encode())
             fd.close()
-            eventlet.sleep(0.0)
+
+            # TODO changed from 0 to make Python 3 tests pass, not sure if ok
+            eventlet.sleep(0.01)
```
temoto: 0.01 is magic number. Two `sleep(0)` also work and could be explained like "wait for complete send, wait for complete receive". It may be not actually these two events that we yield to with sleep, so it doesn't fit as comment without prior investigation, I only establish a meaningful example.

tests/test__greenness.py
```diff
@@ -4,8 +4,15 @@
 """
 import unittest
 
-from eventlet.green import urllib2, BaseHTTPServer
+from eventlet.green import BaseHTTPServer
 from eventlet import spawn, kill
+from eventlet.support import six
+
+if six.PY2:
+    from eventlet.green.urllib2 import HTTPError, urlopen
```
temoto: `from ...green import urllib2`
importing module level names tends to bring more harm than convenience later on.

eventlet/wsgi.py
```diff
@@ -456,6 +448,8 @@ def start_response(status, response_headers, exc_info=None):
                 minimum_write_chunk_size = int(self.environ.get(
                     'eventlet.minimum_write_chunk_size', self.minimum_chunk_size))
                 for data in result:
+                    if not isinstance(data, six.binary_type):
```
temoto: This is a huge API change and may break existing applications. The only way I see it included in v0.17 is keep previous behavior by default and enable bytes check with environment flag, like `EVENTLET_WSGI_BYTES`.

eventlet/wsgi.py
```diff
@@ -8,12 +8,16 @@
 
 from eventlet.green import BaseHTTPServer
 from eventlet.green import socket
-from eventlet.green import urllib
 from eventlet import greenio
 from eventlet import greenpool
 from eventlet import support
 from eventlet.support import six
 
+if six.PY2:
```
temoto: six does this `if PY2` check for you: `from six.moves.urllib.parse import unquote`

As much as I avoid importing module variables/functions, can't think of better option here.

temoto:
To preserve visual locality for things related to greenio, either make it a subpackage or rename: ```
greenio.py
greenio_base.py
greenio_py2.py
greenio_py3.py
```

File: eventlet/wsgi.py (hunk 5 above)
Comment (temoto) [inline eventlet/wsgi.py]: And you don't need another name, it's `TypeError`.

**Comment by temoto** [comment]
Most importantly, don't let this list upset you. When people critique parts of your job, it means they like everything else and willing to pay for it! Likewise here, I can imagine how much praise you'll get from OpenStack team for doing this. What I think about Python3 tests pass... FUNKY AWESOME! And `verify_hub_empty` got more useful output too, I've just tried it.

File: eventlet/greenio.py (hunk 2 above)
Comment (jstasiak) [inline eventlet/greenio.py]: Yes, the reason is there's no ``__all__`` in ``_pyXgreenio`` :)

File: tests/wsgi_test.py (hunk 3 above)
Comment (jstasiak) [inline tests/wsgi_test.py]: From what I saw 2 * ``sleep(0)`` works here just as well; I'll debug it later on and see what events are actually served there.

File: tests/test__greenness.py (hunk 4 above)
Comment (jstasiak) [inline tests/test__greenness.py]: How do you see it being imported as a module name when in the else block we have to do 2 imports from 2 different modules anyway?

File: eventlet/wsgi.py (hunk 5 above)
Comment (jstasiak) [inline eventlet/wsgi.py]: Yes and yes, I'll just make it only fail when there are non-ascii characters (takes care of compatibility issue), GotUnicode is a remain of my earlier patch that I forgot to change, thanks for pointing it out (also once I restore handling unicode strings it'll become ``ValueError`` rather than ``TypeError`` I think)

**Comment by jstasiak** [comment]
Thanks for the feedback, I'll address it later today and hope we can merge this :)

File: tests/test__greenness.py (hunk 4 above)
Comment (temoto) [inline tests/test__greenness.py]: If we don't have common API in green.urllib/2, then forget about it, too much worry for few lines.

File: tests/wsgi_test.py (hunk 3 above)
Comment (jstasiak) [inline tests/wsgi_test.py]: Ha, I figured it out, gonna update the patch

Comment (temoto) [inline tests/wsgi_test.py]: I'm curious to know what you found.

**Comment by jstasiak** [comment]
@temoto I think all your concerns are addressed in https://github.com/eventlet/eventlet/commit/2a8fd1d13c0cb9f0ff86e838533ac8da20fae233.

File: tests/wsgi_test.py (hunk 3 above)
Comment (jstasiak) [inline tests/wsgi_test.py]: See https://github.com/eventlet/eventlet/commit/2a8fd1d13c0cb9f0ff86e838533ac8da20fae233#diff-a5a1a1f5a38ed138c2ab60f183db9e8bR1761

**Comment by jstasiak** [comment]
What Python version are you using?

```python
Python 3.4.2 (default, Oct 16 2014, 05:21:12) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.51)] on darwin
>>> import http
>>> dir(http)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__']
>>> import sys
>>> sys.modules['http.client']
KeyError: 'http.client'
```

temoto:
IPython 2.1.0, doesn't reproduce in normal shell, I see exactly what you just pasted. Forget it.
```
In [1]: import http

In [2]: dir(http)
Out[2]: 
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'client']
```

temoto:
It's the most green Eventlet Travis I've seen. Would you squash the last commit somewhere and decide which branch should be merged from dev/test2?

jstasiak:
It's dev, I just pushed squashed commits.
PS. Yeah I'm quite excited, there's a chance it'll be all green, I have a lead on how to fix PyPy failures.

temoto:
@jstasiak thank you so much, I've just merged it. Please don't forget to eat and sleep well.

jstasiak:
You're welcome. Don't worry :)

PR #200, opened 2015-02-12
Fix exception raising in create_connection()

donaghmccabe:
The create_connection() function could raise a
socket.error instance, where the errno was the actual
socket.error instance.

Fixes https://bitbucket.org/eventlet/eventlet/issue/168/create_connection-incorrectly-raises

### DIFF
--- a/eventlet/green/socket.py
+++ b/eventlet/green/socket.py
@@ -53,9 +53,9 @@ def create_connection(address,
             sock.connect(sa)
             return sock
 
-        except error:
+        except error as e:
             if sock is not None:
                 sock.close()
-            raise
+            raise e
 
     raise error(msg)
--- a/eventlet/green/socket.py
+++ b/eventlet/green/socket.py
@@ -53,9 +53,9 @@ def create_connection(address,
             sock.connect(sa)
             return sock
 
-        except error as e:
-            msg = e
+        except error:
             if sock is not None:
                 sock.close()
+            raise
 
     raise error(msg)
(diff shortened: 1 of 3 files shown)


donaghmccabe:
Sorry, I filled the bug as https://bitbucket.org/eventlet/eventlet/issue/168/create_connection-incorrectly-raises but submitted the fix through githib -- hence the #168 does not match up. 

I don't understand why the CI is failing. I picked a random failure and checked it passed in my environment.

temoto:
Try to leave `except error as e`.

temoto:
Thank you very much for bug report and patch.

On issue links, please read this short mail, it was sent to mailing list, but will be included in contribution section of documentation. https://gist.github.com/temoto/e5b236d6dd8645d1a95c

donaghmccabe:
Opps. Left some debug in there. Will clean up if the CI passes.

temoto:
Okay, I've found the problem. Sorry for misleading with `except error as e`. It's important but not relevant to CI test failures.

We have the following code structure:
```
default_error = "DNS fail"
for addr in resolve(name):
  try: connect(addr)
  except:
    remember error, cleanup socket
raise last or default error
```
patch in question basically removes for-loop, it doesn't try second/etc address. That's an API change we can't afford. So cool that we have tests that fail on this.

But again, thanks for bringing this issue up. I've pushed fix here https://github.com/eventlet/eventlet/commit/411827efc6357afd99a44d7e0d9b5f922c6e8c23 does it look good for you?

donaghmccabe:
Thanks -- thinking about it overnight, I worked out what was wrong with my path, but thanks to timezones, you beat me to it. I presume you will just go ahead and merge your patch and I can close this pull request?

temoto:
Yes, it's merged in 184dce2104b and will be released in v0.17 this month. Thank you, Donagh.

PR #211, opened 2015-03-06
Disable the thread state lock

haypo:
This change fixes the threading support on Python 3.4 when eventlet is used with monkey-patching.

Python 3.4 introduces threading.Thread._tstate_lock to fix a race condition, see:
- https://hg.python.org/cpython/rev/d52b68edbca6
- http://bugs.python.org/issue18808

I'm not really proud of this change: patching a Thread instance after its creation is not the most elegant fix.

I tried to create a Thread subclass in eventlet.green.threading, but I got a lot of new issues. If the subclass is declared in eventlet.green.threading, the parent Thread class uses functions of the original threading module, instead of using patched functions by eventlet. It's unclear to me how symbols are added and patched in modules. I noticed that the threading module of the standard library is imported multiple times. The implementation of monkey-patching is really complex :-/

Currently, the best solution is to patch Thread instance in the patched thread.start_new_thread() function. The problem is to retrieve the instance. My heuristic checks if the wrapped function is the bounded Thread.run() method, and I retrieve the instance from function.__self__.

I'm not yet sure that I really understood the problem. At least, "tox -e py34-selects" pass.

### DIFF
--- a/eventlet/green/thread.py
+++ b/eventlet/green/thread.py
@@ -3,6 +3,7 @@ from eventlet.support.six.moves import _thread as __thread
 from eventlet.support import greenlets as greenlet, six
 from eventlet import greenthread
 from eventlet.semaphore import Semaphore as LockType
+import sys
 
 
 __patched__ = ['get_ident', 'start_new_thread', 'start_new', 'allocate_lock',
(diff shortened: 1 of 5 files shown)


haypo:
Oh, my commit was already pushed to pull request #187, it's not useful to open a duplicate pull request. I'm closing this one.

PR #212, opened 2015-03-10
Use _socket_nodns and select in dnspython support

TimSimmons:
* I ran into issues where, when passing DNS messages through a proxy with
eventlet 0.17.1 sockets would be prematurely closed, and connections would time out and break. 
This was not the case in 0.16.1, this fixes the issue.

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -38,14 +38,27 @@ from eventlet import patcher
 from eventlet.green import _socket_nodns
 from eventlet.green import os
 from eventlet.green import time
+from eventlet.green import select
 from eventlet.support import six
 
 
-dns = patcher.import_patched('dns')
-dns.resolver = patcher.import_patched('dns.resolver')
+dns = patcher.import_patched('dns',
+                             select=select,
+                             time=time,
+                             os=os,
+                             socket=_socket_nodns)
+dns.resolver = patcher.import_patched('dns.resolver',
+                                      select=select,
+                                      time=time,
+                                      os=os,
+                                      socket=_socket_nodns)
 
 for pkg in ('dns.entropy', 'dns.inet', 'dns.query'):
-    setattr(dns, pkg.split('.')[1], patcher.import_patched(pkg))
+    setattr(dns, pkg.split('.')[1], patcher.import_patched(pkg,
+                                                           select=select,
+                                                           time=time,
+                                                           os=os,
+                                                           socket=_socket_nodns))
 import dns.rdtypes
 for pkg in ['dns.rdtypes.IN', 'dns.rdtypes.ANY']:
     setattr(dns.rdtypes, pkg.split('.')[-1], patcher.import_patched(pkg))


temoto:
Hello, @TimSimmons 
Thanks for bug report and the patch. Please describe failing configuration in more detail, so maybe it's also easy to construct a regression test.

TimSimmons:
hi @temoto 
If I'm honest, I don't really understand the problem in it's completeness. But I'll give it my best shot. 
What's happening is I'm sending DNS NOTIFYs (and other opcodes) through a proxy (HAProxy) over TCP. That NOTIFY goes through fine, but the following AXFR (Zone Transfer) that is triggered from the receiving side seems to fail. I believe this is a totally new TCP connection. 

Logs:

From the original sender of the NOTIFY, and subsequent handler of AXFR:

		ERROR designate.dnsutils [-] Unhandled exception while processing request
		TRACE designate.dnsutils Traceback (most recent call last):
		TRACE designate.dnsutils   File "/opt/designate/local/lib/python2.7/site-packages/designate/dnsutils.py", line 242, in handle
		TRACE designate.dnsutils     client.send(tcp_response)
		TRACE designate.dnsutils   File "/opt/designate/local/lib/python2.7/site-packages/eventlet/greenio/base.py", line 359, in send
		TRACE designate.dnsutils     total_sent += fd.send(data[total_sent:], flags)
		TRACE designate.dnsutils error: [Errno 32] Broken pipe

From the receiver of the initial request, initiate of AXFR:

		ERROR designate.dnsutils [-] There was a problem with the AXFR
		TRACE designate.dnsutils Traceback (most recent call last):
		TRACE designate.dnsutils   File "/opt/designate-0.3.0/designate/designate/dnsutils.py", line 279, in do_axfr
		TRACE designate.dnsutils     raw_zone = dns.zone.from_xfr(xfr, relativize=False)
		TRACE designate.dnsutils   File "/opt/designate-0.3.0/local/lib/python2.7/site-packages/dns/zone.py", line 1026, in from_xfr
		TRACE designate.dnsutils     for r in xfr:
		TRACE designate.dnsutils   File "/opt/designate-0.3.0/local/lib/python2.7/site-packages/dns/query.py", line 434, in xfr
		TRACE designate.dnsutils     ldata = _net_read(s, 2, mexpiration)
		TRACE designate.dnsutils   File "/opt/designate-0.3.0/local/lib/python2.7/site-packages/dns/query.py", line 246, in _net_read
		TRACE designate.dnsutils     raise EOFError
		TRACE designate.dnsutils EOFError

		TRACE designate.dnsutils Traceback (most recent call last):
		TRACE designate.dnsutils   File "/opt/designate/local/lib/python2.7/site-packages/designate/dnsutils.py", line 242, in handle
		TRACE designate.dnsutils     client.send(tcp_response)
		TRACE designate.dnsutils   File "/opt/designate/local/lib/python2.7/site-packages/eventlet/greenio/base.py", line 359, in send
		TRACE designate.dnsutils     total_sent += fd.send(data[total_sent:], flags)
		TRACE designate.dnsutils error: [Errno 32] Broken pipe

It seems that these changes I've made keep the connection open on these queries. 

Is this what you're looking for? Perhaps if I understood a bit more about eventlet I could explain things better. But it seems that the socket I substituted would be important. If I'm honest, I don't really know what "select" is doing, but it's needed :smiley:

TimSimmons:
Ping @temoto :)

temoto:
I should've been more clear. I'm looking for simple instructions how to reproduce your problem. Example:

* haproxy configured like this ...
* DNS server XXX, configured like this ...
* send this message to proxy

TimSimmons:
@temoto It isn't a terribly simple reproduction unfortunately. When you send the request through the proxy, it must be answered by a DNS server that was built leveraging eventlet's dnspython support. Our product is such a thing. But unless I create a small functioning dns server with dnspython + eventlet. I'm afraid that'll be the only way to see the behavior. 

Even that might not be enough, because the conditions I observed this in include a cross-region deployment on 3 different servers. So there was some amount of latency involved. 

Another way to approach this might be for you to help me with what I've actually done here, because if I'm honest, I only partly understand what I'm accomplishing.

temoto:
So your product is a DNS server. Could you describe what it does (network wise) upon receiving NOTIFY?

dnspython package `__init__` (imported as `dns`) module could have `import socket` line. Original `import_patched('dns')` basically does two things: 1) vanilla `import dns` 2) replace dns.socket with eventlet.green.socket. Note: by default, `green.socket` itself replaces name resolving functions with dnspython implementations using eventlet green sockets. Your first change makes `import_patched` to replace `dns.socket` with "mostly green" version, that is eventlet sockets but native name resolution functions. Honestly I could not imagine why dnspython would call any stdlib name resolution functions. Maybe it's some kind of fallback behaviour.

Your second change makes `dns.resolver` use green version of stdlib `select` module. Basically, this module provides one function that is used to wait for events on a set of non-blocking sockets. It makes sense to patch all of `dns.resolver` network functions with green versions, though I'd expect that to be default, not requiring `select=select` kwarg.

TimSimmons:
Sure. 

One clarification, the product in this case is actually two sides of a DNS transaction, both implementing DNS protocol using DNSPython with Eventlet. 

The side sending the NOTIFY is the one that needs this change. Alright, continuing. 

The side receiving the NOTIFY, does some quick logic, and then perfroms an AXFR request. So opening up a new connection to the original notifier, this results in a response with zone information. After the AXFR, the response to the NOTIFY goes back. Where it breaks down with Eventlet 0.17.1 seems to be that the AXFR request, so going to the original notifier, doesn't get there. 

Based on your excellent explanation (thank you!) I can make a couple conclusions.

DNSPython does seem to mess around a bit with the stdlib name resolution functions: https://github.com/rthalley/dnspython/blob/6d52fda5e2d887301e748ca9ac3e7dd6b5ffda25/dns/resolver.py#L1053 perhaps changing them to the eventlet ones, and then having this get reset to the stdlib ones was messing something up?

The select change seems very necessary for this use case, where I'm opening sockets for a period, and then having other ones come around and do other things. That this specific change is needed tells me that ones of the other green modules that is being imported by default might be messing things up. I'll try and figure out what that is.

TimSimmons:
So the problem seemed to be the green socket class. I've updated the PR to patch everything that should be. Out of curiosity, why was this using the nodns socket before, and what prompted the switch? 

https://github.com/eventlet/eventlet/commit/3de49aef1399594b2cab0652d9ff3976d19b5e72#diff-47bd8cdaac88cccf21624af4df7a35c3L54

What are the next steps for this matter? :smile:

temoto:
I'm sorry, have to warn you that this is going to wait one ping.

TimSimmons:
I'm sorry, I don't understand what you mean?

temoto:
I mean I'm very busy, so if that doesn't trouble you, please ping me again later, as you did above.

TimSimmons:
@temoto Ping :)

TimSimmons:
@temoto Whenever you have a moment :)

TimSimmons:
@temoto Ping

temoto:
Thanks for reminders. :-)

temoto:
@TimSimmons your fix is released in v0.17.2, available on PyPi.

PR #215, opened 2015-03-23
greenpool: Make sure that imap/starmap set the correct pool

djipko:
Calling a GreenPool.imap() followed by a GreenPool.waitall() does not in fact
wait on the threads that were spawned, this is because we create a
completely new GreanPool instance for the resulting GreenPile result.

We really want to be spawning threads limited by the GreenPool instance
we call the imap() on, so we should be using it for the resulting GreenPile.

### DIFF
--- a/eventlet/greenpool.py
+++ b/eventlet/greenpool.py
@@ -150,7 +150,7 @@ class GreenPool(object):
         """
         if function is None:
             function = lambda *a: a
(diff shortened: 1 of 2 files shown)


djipko:
Hmmm... the CI failures do not seem to be related to my change.

djipko:
Closed and re-oppened in order to re-kick travis

Issue #223, opened 2015-04-09
threading.Threading.join() broken in Python 3.4 with monkey-patching


haypo:
Hi,

threading.Threading.join() doesn't work in Python 3.4 with monkey-patching. It's related to the new "thread-state lock" introduced in Python 3.4:
https://hg.python.org/cpython/rev/d52b68edbca6
https://bugs.python.org/issue18808

This lock is related internally when the C thread exits. So it doesn't work well with monkey-patching.

My pull request #187 fixed threading.Condition, but it also tried to handle this new lock by setting it to None.

haypo:
I wrote a fix in pull #224. With this fix, tests of the Oslo Messaging project now pass on python 3.4 with eventlet and monkey-patching!

And if my pull request is merged, would it possible to prepare a new release? It would allow me to port more OpenStack projects to python3! Eventlet is blocking more than 10 projects because it doesn't work well on python 3. It looks like this issue is the last known eventlet issue with monkey-patching on Python 3.

temoto:
Congratulations with Oslo/Py3.4! I'll bake 0.17.3 bugfix some time soon. And thanks for letting me know of blockers, I was going to communicate with somebody from OpenStack team to know if there are any eventlet blockers for you guys.

PR #224, opened 2015-04-09
Issue #223: Fix threading monkey-patching on py3.4

haypo:
Fix the monkey-patching of the threading module on Python 3.4. Instead
of removing the thread state lock, patch the _bootstrap_inner() method
to release it.

### DIFF
--- a/tests/patcher_test.py
+++ b/tests/patcher_test.py
@@ -502,3 +502,7 @@ def test_importlib_lock():
 
 def test_threading_condition():
     tests.run_isolated('patcher_threading_condition.py')
+
+
+def test_threading_join():
+    tests.run_isolated('patcher_threading_join.py')
--- a/eventlet/green/thread.py
+++ b/eventlet/green/thread.py
@@ -52,10 +52,20 @@ def start_new_thread(function, args=(), kwargs=None):
         # With monkey patching, eventlet uses green threads without python
         # thread state, so the lock is not automatically released.
         #
-        # Disable the thread state lock to avoid dead locks.
+        # Wrap _bootstrap_inner() to release explicitly the thread state lock
+        # when the thread completes.
         thread = function.__self__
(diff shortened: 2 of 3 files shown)


haypo:
I replaced the assertion with an if to be more safe.

tests/patcher_test.py
```diff
@@ -502,3 +502,7 @@ def test_importlib_lock():
 
 def test_threading_condition():
     tests.run_isolated('patcher_threading_condition.py')
+
+
+def test_threading_join():
+    tests.run_isolated('patched_threading_join.py')
```
temoto: Small typo "patcher_...", after module name. It's not a blocker for merge, I'm just bringing it to attention.

haypo: It's now fixed.

temoto:
Just curious, what assertion?

haypo:
> Just curious, what assertion?

In wrap_bootstrap_inner(), I wrote "assert thread._tstate_lock is not None". But I replaced it with an if, because it looks like the lock can be None in some corner cases. Now the code should work in all cases ;-)

temoto:
Merged. Thanks for quick bug report and fix. More thanks for rebasing, it was a pleasure to handle this.

haypo:
> Merged. Thanks for quick bug report and fix. More thanks for rebasing, it was a pleasure to handle this.

Wow, this merge was really fast! Thanks for the quick review.

Now, would it be possible to release eventlet with this change? It would help OpenStack *a lot* to port the project to Python 3. Currenctly, eventlet blocks porting most OpenStack applications to Python 3 because it doesn't (fully) work on Python 3 (with monkey-patching).

temoto:
Yup, I just released v0.17.3 https://pypi.python.org/pypi/eventlet/0.17.3
Thanks for contributions!

PR #225, opened 2015-04-09
tox 1.8 or newer is required

haypo:
The "py{27,34,py}-dns" syntax in envlist is only supported on tox 1.8 or
newer.


temoto:
Thanks, merged.

Issue #230, opened 2015-04-17
Able to cause '_wait_for_tstate_lock' inability to join in threads


harlowja:
Something interesting that might be some underlying issue with the current eventlet (0.17.3) + python 3.4.

It appears it is very easy to cause ``_wait_for_tstate_lock`` of threads to stop working (and therefore join() to not work) without even using eventlet (but just by late importing it).

An example that seems to reliably cause this to occur (100% of the time?):

```
import collections
import random
import sys
import threading
import time

from six.moves import queue as compat_queue
from six.moves import range as compat_range

from oslo_utils import importutils

# Delay between producing/consuming more work.
WORKER_DELAY, PRODUCER_DELAY = (0.5, 0.5)

# How many workers and producers of work will be created (as threads).
PRODUCERS = 3
WORKERS = 5

PRODUCER_UNITS = 10
EXPECTED_UNITS = PRODUCER_UNITS * PRODUCERS
WORK_TIME = 1.0

# Avoid overlaps (may be not needed...)
STDOUT_LOCK = threading.Lock()


def daemon_thread(target, *args, **kwargs):
    """Makes a daemon thread that calls the given target when started."""
    thread = threading.Thread(target=target, args=args, kwargs=kwargs)
    # NOTE(skudriashev): When the main thread is terminated unexpectedly
    # and thread is still alive - it will prevent main thread from exiting
    # unless the daemon property is set to True.
    thread.daemon = True
    return thread


def safe_print(name, message, prefix=""):
    with STDOUT_LOCK:
        if prefix:
            print("%s %s: %s" % (prefix, name, message))
        else:
            print("%s: %s" % (name, message))


def producer(ident, q):
    name = "P-%s" % (ident)
    safe_print(name, "started")
    for i in compat_range(0, PRODUCER_UNITS):
        details = {
            'color': random.choice(['red', 'blue']),
        }
        q.put(details)
        safe_print(name, "'%s' [posted]" % (details))
        time.sleep(PRODUCER_DELAY)
    safe_print(name, "finished", prefix=">>>")


def worker(ident, q, consumed):
    # Doing these import(s) in the thread seems to cause it to bork...
    eventlet = importutils.try_import('eventlet')
    eventlet_patcher = importutils.try_import('eventlet.patcher')

    name = "W-%s" % (ident)
    safe_print(name, "started")
    while len(consumed) != EXPECTED_UNITS:
        try:
            details = q.get(False, WORKER_DELAY)
[...]

PR #231, opened 2015-04-20
Issue #230: Fix patcher.original()

haypo:
Don't load a module twice when eventlet monkey-patched is not used.

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -175,6 +175,11 @@ def original(modname):
     if original_name in sys.modules:
         return sys.modules.get(original_name)
 
+    if not already_patched and modname in sys.modules:
+        # Issue #230: Don't load a module twice
+        # when eventlet monkey-patched is not used
+        return sys.modules[modname]
+
     # re-import the "pure" module and store it in the global _originals
     # dict; be sure to restore whatever module had that name already
     saver = SysModulesSaver((modname,))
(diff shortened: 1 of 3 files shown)


temoto:
Thanks for fix.

Are you ok with this change in all 3 modified files?

```
@@ -176,7 +176,7 @@ def original(modname):
        return sys.modules.get(original_name)

    if not already_patched and modname in sys.modules:
-       # Issue #230: Don't load a module twice
+      # https://github.com/eventlet/eventlet/issues/230 Don't load a module twice
        # when eventlet monkey-patched is not used
        return sys.modules[modname]
```

temoto:
It fails on selects hub. Please have a look https://travis-ci.org/eventlet/eventlet/jobs/59281880

haypo:
I didn't understand that calling eventlet.monkey_patch() modifies modules already loaded:

```
import threading
import eventlet
eventlet.monkey_patch()
# threading has been modified
```

So my patch doesn't seem correct.

Calling eventlet.monkey_patch() in a thread doesn't work. On Python 2, a KeyError is displayed. On Python 3.4, the thread hangs. Example:

```
import threading

def test():
    import eventlet; eventlet.monkey_patch()

t = threading.Thread(target=test)
t.start()
t.join()
```

Output on Python 2.7:

```
Exception KeyError: KeyError(140222136564944,) in <module 'threading' from '/usr/lib64/python2.7/threading.pyc'> ignored
```

temoto:
Modifying modules, no matter if they were loaded or not, is the whole point of monkey patching. And, yeah, it's a tough issue. Do you have ideas?

haypo:

Issue #226, opened 2015-04-11
HTTPS/SSL failure when using requests library on Python 3.4


mkai:
I use Python 3.4.3. HTTPS requests using the requests library normally work fine:

```python
>>> import requests
>>> requests.get('https://www.github.com/').status_code
200
```

However, if eventlet is imported, I get the following error:

```python
>>> import eventlet
>>> import requests
>>> requests.get('https://www.github.com/').status_code
venv/lib/python3.4/site-packages/requests/packages/urllib3/connection.py:251: SecurityWarning: Certificate has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)
  SecurityWarning
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "venv/lib/python3.4/site-packages/requests/api.py", line 68, in get
    return request('get', url, **kwargs)
  File "venv/lib/python3.4/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "venv/lib/python3.4/site-packages/requests/sessions.py", line 464, in request
    resp = self.send(prep, **send_kwargs)
  File "venv/lib/python3.4/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "venv/lib/python3.4/site-packages/requests/adapters.py", line 370, in send
    timeout=timeout
  File "venv/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py", line 544, in urlopen
    body=body, headers=headers)
  File "venv/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py", line 341, in _make_request
    self._validate_conn(conn)
  File "venv/lib/python3.4/site-packages/requests/packages/urllib3/connectionpool.py", line 762, in _validate_conn
    conn.connect()
  File "venv/lib/python3.4/site-packages/requests/packages/urllib3/connection.py", line 253, in connect
    match_hostname(cert, self.assert_hostname or hostname)
  File ".pyenv/versions/3.4.3/lib/python3.4/ssl.py", line 261, in match_hostname
    raise ValueError("empty or no certificate, match_hostname needs a "
ValueError: empty or no certificate, match_hostname needs a SSL socket or SSL context with either CERT_OPTIONAL or CERT_REQUIRED
```

[...]

PR #233, opened 2015-05-01
Issue #226: HTTPS: Fix empty SSL certificate issue

stuart-mclaren:
On python 3 and python 2.7.9, the requests library would raise the
following value error:

 ValueError: empty or no certificate, match_hostname needs
 a SSL socket or SSL context with either CERT_OPTIONAL or CERT_REQUIRED


temoto:
Thank you, I've pushed a07f836 following your idea and credit but without explicit listing of other kwargs.

stuart-mclaren:
Thanks!

Issue #192, opened 2015-01-21
processes are consuming more than 90% of CPU when calling send method in eventlet/greenio.py


XuhuiMiao:
[Scenario]
1. Start rsyslog server.
2. Run our application.
3. Stop rsyslog server
4. Trigger a syslog in our application.
5. The thread of logging consumes 90+% of cpu.

[Description]
According to debug result, we found an endless loop found in send method, the send method would be blocked when no data was sent as description in the document(greenio.py), but unfortunately the send method returned immediately, and the return value was 107( "'Transport endpoint is not connected'"). The return value will raise an exception in Windows System but not in Linux System. I think It is lack of consideration for all cases of the return value in Linux System. 
```
/usr/lib/python2.6/site-packages/eventlet/greenio.py
if sys.platform[:3] == "win":
    # winsock sometimes throws ENOTCONN
    SOCKET_BLOCKING = set((errno.EAGAIN, errno.EWOULDBLOCK,))
    SOCKET_CLOSED = set((errno.ECONNRESET, errno.ENOTCONN, errno.ESHUTDOWN))
else:
    # oddly, on linux/darwin, an unconnected socket is expected to block,
    # so we treat ENOTCONN the same as EWOULDBLOCK
    SOCKET_BLOCKING = set((errno.EAGAIN, errno.EWOULDBLOCK, errno.ENOTCONN))
    SOCKET_CLOSED = set((errno.ECONNRESET, errno.ESHUTDOWN, errno.EPIPE))

/usr/lib/python2.6/site-packages/eventlet/greenio.py
    def send(self, data, flags=0):
        fd = self.fd
        if self.act_non_blocking:
            return fd.send(data, flags)

        # blocking socket behavior - sends all, blocks if the buffer is full
        total_sent = 0
        len_data = len(data)
        while 1:
            try:
                total_sent += fd.send(data[total_sent:], flags)
            except socket.error as e:
                if get_errno(e) not in SOCKET_BLOCKING:
                    raise

            if total_sent == len_data:
                break

            try:
                self._trampoline(self.fd, write=True, timeout=self.gettimeout(),
                           timeout_exc=socket.timeout("timed out"))
            except IOClosed:
                raise socket.error(errno.ECONNRESET, 'Connection closed by another thread')

        return total_sent

[Debug]
> /usr/lib/python2.6/site-packages/eventlet/greenio.py(346)send()
-> try:
(Pdb) n
[...]

PR #234, opened 2015-05-14
Issue #192:Fix inifinte loop when connection lost

xtrusia:
when "Transport endpoint is not connected" error occurs on send function
it can be catched by except statement but skipped by if statement (because SOCKET_BLOCKING has ENOTCONN)
so comes infinite loop, then hang.
I worried about regression if i remove errno.ENOTCONN on SOCKET_BLOCKING
then I added new line for checking ENOTCONN


temoto:
Thank you very much, I've pushed commit ed06553 with your idea but one get_errno call and == instead of is comparison, all credit is still yours of course.

PR #237, opened 2015-06-29
Port eventlet.green.OpenSSL to Python 3

haypo:
eventlet.green.OpenSSL is not compatible with Python 3:

```
>>> import eventlet.green.OpenSSL
  File "/home/haypo/prog/openstack/eventlet/eventlet/green/OpenSSL/__init__.py", line 1, in <module>
    import rand
ImportError: No module named 'rand'
```

This pull request fix eventlet.green.OpenSSL for Python 3:
* Fix import syntax for Python 3: use relative imports
* Add unit test

### DIFF
--- a/tests/openssl_test.py
+++ b/tests/openssl_test.py
@@ -1,0 +1,20 @@
+from tests import LimitedTestCase, main, SkipTest
+
+class TestOpenSSL(LimitedTestCase):
+    def test_import(self):
+        try:
+            import OpenSSL
+        except:
+            raise SkipTest("need OpenSSL")
+
+        # Ensure that it's possible to import eventlet.green.OpenSSL.
+        # Most basic test to check Python 3 compatibility.
+        import eventlet.green.OpenSSL.SSL
+        import eventlet.green.OpenSSL.crypto
+        import eventlet.green.OpenSSL.rand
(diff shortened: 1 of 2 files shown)


tests/openssl_test.py
```diff
@@ -0,0 +1,15 @@
+from tests import LimitedTestCase, main
+
+class TestOpenSSL(LimitedTestCase):
```
temoto: `import`, `class` and `main` are all redundant. Just `def test_import():` would work just fine. Would you care to remove them? Or I can do it.

haypo: I like the TestCase API (defining methods). I used queue_test.py as a template, this file calls main().

Please rewrite the test as you want, I don't care on how tests are written, I only care of having this bug fixed :-)

temoto:
Unfortunately, it fails when pyopenssl library is not installed. Test with `tox -e py27-dns py34-dns`.

haypo:
> Unfortunately, it fails when pyopenssl library is not installed. Test with tox -e py27-dns py34-dns.

I modified my test to skip it if OpenSSL is missing.

temoto:
Merged into master, thank you very much.

PR #241, opened 2015-08-10
wsgi: better error for chunk read failures

smerritt:
When a client is sending a chunked request body and disconnects
between chunks, wsgi.Input.read() raises a ValueError. That's fine if
you can remember to wrap every single call to read/readline in a
try/except, but it's bad for a higher-level error handler because so
many things raise ValueError.

This adds in a new error class wsgi.ChunkReadError and raises that
instead. It inherits from ValueError so anyone catching ValueError
should continue to work.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -50,6 +50,10 @@ BAD_SOCK = set((errno.EBADF, 10053))
 BROKEN_SOCK = set((errno.EPIPE, errno.ECONNRESET))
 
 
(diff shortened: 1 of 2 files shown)


temoto:
Thank you, merged.

clayg:
Can we catch this when the wsgi server? (i.e. https://github.com/smerritt/eventlet/blob/c3ce3eef0b4d0dfdbfb1ec0186d4bb204fb8ecd5/eventlet/wsgi.py#L515)

PR #244, opened 2015-08-21
wsgi: expose raw HTTP headers in the environment

smerritt:
To explain this commit, I'll need to provide three pieces of
background information.

First, PEP 333 gives us rules for how to convert HTTP headers to WSGI
environment keys: uppercase the names, combine repeated headers into
comma-separated strings, and convert dashes to
underscores. Case-mangling and header combining are things that RFC
7230 (HTTP 1.1) defines, but the dash-to-underscore conversion is
something PEP 333 made up to mimic CGI environment variables, and CGI
has dash-to-underscore conversion since a dash is not a valid
character in a shell variable name.

Second, Amazon S3 lets a client authenticate by including an
"Authorization" header in the request; its value is the HMAC of some
normalized form of the request, including certain headers (notably,
those of the form x-amz-meta-*). That normalization includes sorting
and lowercasing, but it does *not* include dash-to-underscore
conversion.

Third, there is a Python project called swift3; it is a WSGI
middleware that attempts to provide an S3-compatible API atop
OpenStack Swift.

An attentive reader may have already noticed the problem: when signing
an S3 request with a header like "x-amz-meta-wing_nut_spin: leftward"
or "x-amz-meta-wing-nut-spin: leftward" the WSGI environment contains
the key HTTP_X_AMZ_WING_NUT_SPIN, but swift3 cannot tell which
underscores were originally dashes and which were originally
underscores, so it has no way to compute the correct HMAC, and thus
cannot authenticate requests.

(Of course, "x-amz-meta-wing_nut_spin" is made up, but
"x-amz-meta-sts_image_version" is a real header emitted by Symantec's
NetBackup product, and I've seen such headers used by bespoke
proprietary apps as well.)

This commit introduces a new set of keys in the WSGI environment:
RAWHTTP_*. If the client sends a header line "Thing-type-One: two",
then the WSGI environment will have env[b"RAWHTTP_Thing-type-One"] =
b"two". Keys are not case-normalized or underscore-normalized, and
both keys and values are of type bytes. Keys and values should match
the bytes put on the wire by the client.

swift3 can use this to get the raw headers sent by the client and
perform its own normalization, thus arriving at the correct HMAC
value.

This is a little more general than swift3 needs, but my hope is that
it is general enough that (a) no more ugly hacks are needed for other
applications that can't abide by WSGI's header mangling, and (b) it
can eventually be adopted by other WSGI servers.

Thanks to Peter Portante for reviewing an early version of this
change.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -575,9 +575,9 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
         else:
             headers = [h.split(':', 1) for h in headers]
 
-        for k, v in headers:
-            k = k.replace('-', '_').upper()
-            v = v.strip()
+        for raw_k, raw_v in headers:
+            k = raw_k.replace('-', '_').upper()
+            v = raw_v.strip()
             if k in env:
                 continue
             envk = 'HTTP_' + k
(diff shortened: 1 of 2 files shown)


temoto:
Why not dump the whole `HTTP_` prefix idea altogether: `environ['raw_header'] == [('x-amz-meta-sts_image_version', '1.0'), ...]` (list of 2-tuples or multidict), no concatenation either.

smerritt:
that'd work too... really, this whole thing kind of stinks, since it duplicates all the information, so it's all too easy for middlewares to write contradictory things into the environment. I need to think harder about this.

PR #246, opened 2015-09-14
Fix docstring which pointed to wrong Queue

rameshg87:
This commit fixes the docstring which pointed
to the wrong class eventlet.queue.Queue instead
of the one in standard library.


temoto:
Thank you, merged in master 299da78

Issue #248, opened 2015-09-21
eventlet.monkey_patch() on Python 3.4 makes stdout non-blocking


rbtcollins:
We have a test suite which writes it's output to stdout. We started observing short writes on Python 3.4:

http://paste.ubuntu.com/12513141/

```
  File "/tmp/designate/.tox/py34/lib/python3.4/site-packages/subunit/v2.py", line 215, in _write_packet
    raise Exception("Failed to write packet (%d) bytes" % len(buf))
Exception: Failed to write packet (4271) bytes
{3} subunit.parser [0.000328s] ... FAILED
```

Now, I could put a loop in there with a timeout etc - but that is the is synchronous codepath in subunit, because we're writing to a known quantity.

This behaviour doesn't happen on Python2.6 or 2.7 with eventlet, so its just a *bit* of a puzzle.

On Python3.4 we get a binary file object by unwrapping stdout - sys.stdout.buffer is the object we're writing to, and we obtain the handle for that before monkeypatching is done (because the test harness is itself not an eventlet app)

Even more odd, piping the suite output through cat prevents the errors being triggered - I have no idea why!

haypo:
Are you sure that standard streams (stdin, stdout, stderr) are non-blocking? I'm unable to reproduce the issue. Example with Python 3.5:
```
import eventlet
eventlet.monkey_patch()
import os
for fd in range(3):
    print(fd, os.get_blocking(fd))
```

Output:
```
0 True
1 True
2 True
```

How can I try to reproduce the issue? Did you check with "strace -f -o trace" to check that fd 0, 1 or 2 is made non-blocking with fcntl()?

By default (with no redirection), all standard streams are the same device:
```
$ ls -l /proc/self/fd/[012]
lrwx------. 1 haypo haypo 64 22 sept. 10:12 /proc/self/fd/0 -> /dev/pts/0
lrwx------. 1 haypo haypo 64 22 sept. 10:12 /proc/self/fd/1 -> /dev/pts/0
lrwx------. 1 haypo haypo 64 22 sept. 10:12 /proc/self/fd/2 -> /dev/pts/0
```

If you modify one stream, the two other streams are impacted. Example:
```
import os
os.set_blocking(1, False)
for fd in range(3):
    print(fd, os.get_blocking(fd))
```

Output:
```
0 False
1 False
2 False
```

[...]

PR #250, opened 2015-09-24
GH #248: Fix GreenFileIO.write()

haypo:
On Python 3, GreenFileIO.write() now always write all bytes using a
loop, to have the same behaviour than GreenPipe.write() on Python 2.
Before, the write() could be partial, write less bytes than expected.

On Python 2, GreenPipe.write() doesn't allow partial writes, it
always write all bytes, even if the user requested an unbuffered pipe
(ex: by calling os.fdopen(fd, 'wb', 0)).

Modifying Python 2 to be pedantic and allow partial writes for
unbuffered pipes will likely break a lot of applications. It's
simpler to modify Python 3 GreenFileIO to behave the same than Python
2 GreenPipe.

### DIFF
--- a/eventlet/greenio/py3.py
+++ b/eventlet/greenio/py3.py
@@ -132,14 +132,19 @@ class GreenFileIO(_OriginalIOBase):
         self._closed = True
 
     def write(self, data):
-        while True:
+        view = memoryview(data)
+        datalen = len(data)
+        offset = 0
+        while offset < datalen:
             try:
-                return _original_os.write(self._fileno, data)
+                written = _original_os.write(self._fileno, view[offset:])
             except OSError as e:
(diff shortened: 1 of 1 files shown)


haypo:
Hi, any progress on this change?

temoto:
Thank you, the fix is merged in master e3f88347d96f653d65f66a46b6bda618a6cb129d

haypo:
Wow, it's crazy how much I got from Travis because tests failed!

Please review my pull request #254 which fixes eventlet tests on Python 3. It's a regression introduced by the commit 299da78a65c4c0f87070d92098675677eb113199.

temoto:
Yup, sorry, for some reason I wanted a green build this time.

PR #254, opened 2015-10-13
Fix eventlet.queue on Python 3

haypo:
Get the queue module of the Python standard library from
six.moves.queue to be compatible with Python 3.


temoto:
Thanks, but I've already fixed it in master.

haypo:
> Thanks, but I've already fixed it in master.

Great! Thanks a lot :-) Travis flooded my mailbox with errors.

PR #257, opened 2015-10-19
Fix GreenFileIO.readall() for regular file

haypo:
GreenFileIO.readall() calls _trampoline(self, read=True). Problem: on
regular file, epoll_ctl() fails with EPERM on Linux, because regular
files are not supported.

This change skips the call to _trampoline() on regular files. Add
also various unit tests on the various ways to read the whole content
of a file.

### DIFF
--- a/eventlet/greenio/py3.py
+++ b/eventlet/greenio/py3.py
@@ -10,6 +10,7 @@ from io import (
     TextIOWrapper as _OriginalTextIOWrapper,
     IOBase as _OriginalIOBase,
 )
+import stat
 from types import FunctionType
 
 from eventlet.greenio.base import (
@@ -47,6 +48,8 @@ class GreenFileIO(_OriginalIOBase):
         self._closed = False
         set_nonblocking(self)
         self._seekable = None
+        mode = _original_os.fstat(self._fileno).st_mode
+        self._is_regular = stat.S_ISREG(mode)
 
     @property
     def closed(self):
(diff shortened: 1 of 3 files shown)


haypo:
ping?

jstasiak:
Hey @haypo, do you mind moving the code in ``regular_file_readall.py`` into a section guarded by ``if __name__ == '__main__'`` so that it's import-safe?

haypo:
> Hey @haypo, do you mind moving the code in regular_file_readall.py into a section guarded by if __name__ == '__main__' so that it's import-safe?

Sure, it's done.

eventlet/greenio/py3.py
```diff
@@ -98,7 +101,8 @@ def readall(self):
             except OSError as e:
                 if get_errno(e) not in SOCKET_BLOCKING:
                     raise IOError(*e.args)
-            self._trampoline(self, read=True)
+            if not self._is_regular:
```
temoto: According to `stat()` documentation, file descriptor may be a regular file, directory, symlink, block or char device, socket and fifo. What are supported types? Is it possible to use white list instead? Check that fd is of supported type.

davidszotten: hm. looking at this again actually makes me wonder if #314 is a better fix. the problem is triggered when we call `epoll.register` on a regular file, but we should only be doing that when we catch something in `SOCKET_BLOCKING`. if i understand things correctly, this should never be raised from regular files.

davidszotten:
this should be fixed by #314

haypo:

PR #259, opened 2015-10-24
Fixed Python 3 compatibility errors

astawiarski:
Two following exceptions were being thrown when attempting to use this profiler under Python 3:

```
dispatch = dict([(key, ContextWrap(val)) for key, val in six.iteritems(dispatch)])
NameError: name 'ContextWrap' is not defined
```

```
thread = patcher.original('thread') # non-monkeypatched module needed
ImportError: No module named 'thread'
```

### DIFF
--- a/eventlet/green/profile.py
+++ b/eventlet/green/profile.py
@@ -42,7 +42,7 @@ from eventlet import greenthread
 from eventlet import patcher
 from eventlet.support import six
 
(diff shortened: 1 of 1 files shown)


temoto:
Thanks, your fix is merged in master. I know it's embarrassing 16 months, no hate please. :-)

PR #262, opened 2015-11-10
Added default value for max_size in eventlet's wsgi server.

ashutosh-mishra:
Need to add default value for max_size variable in wsgi server.
While debugging some scale issue in our wsgi server, not getting(through description) how much connections server can support simultaneously.


temoto:
Thank you, documentation is very important.

I've changed your patch to avoid too long line. Merging soon.

temoto:
Merged in master 503c584a7d30d7ced1509df2c0c9262be1157589

PR #263, opened 2015-11-17
Avoid set_nonblocking() for GreenPipe on Windows

kvinod1010:
GreenPipe init fails on Windows platform due to set_nonblocking() call, however avoiding set_nonblocking() call on windows does works and GreenPipe can be created and used.


nat-goodspeed:
I wonder if the author encountered #483?

For a nontrivial child-process run, I *want* the pipes passed to that child to be nonblocking.

PR #268, opened 2015-11-25
doc: allow sphinx to create pure text file docs

anthraxx:
A lot of people like simple text file based documentation that can be read on the terminal.
Sphinx allows to create those with the 'text' formatter, this patches includes the according
call in the doc/Makefile.


jstasiak:
Thank you @anthraxx, just merged into master!

PR #279, opened 2016-01-04
Add Python 3.5 to the Tox and Travis test envs

pjdelport:
This should help catch things like da7f9732.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,32 +1,39 @@
 language: python
-python: 2.7
-env:
-  matrix:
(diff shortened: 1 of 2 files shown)


pjdelport:
Hmm, it looks like there may be a hoop or two to jump through to get Python 3.5 working on Travis, currently: travis-ci/travis-ci#4794

temoto:
This is still open by mistake, your patch or derivative has been merged long ago, we even test against Python3.6 now. Thanks a lot.

PR #280, opened 2016-01-07
Update __init__.py

caot:
trace exception that help fix potential issues.

### DIFF
--- a/eventlet/__init__.py
+++ b/eventlet/__init__.py
@@ -40,6 +40,10 @@ try:
     exc_after = greenthread.exc_after
     call_after_global = greenthread.call_after_global
 except ImportError as e:
+    import sys
+    import logging
+    print >> sys.stderr, logging.traceback.format_exc()
+    
     # This is to make Debian packaging easier, it ignores import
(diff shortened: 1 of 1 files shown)


temoto:
```
    # This is to make Debian packaging easier, it ignores import
    # errors of greenlet so that the packager can still at least
    # access the version.  Also this makes easy_install a little quieter
    if 'greenlet' not in str(e):
        # any other exception should be printed
        import traceback
        traceback.print_exc()
```
Does this make what you wanted?

caot:
The exception was raised due to statement monkey_patch = patcher.monkey_patch, so it's good to traceback any exception, and to remove the statement "if 'greenlet' not in str(e):". I ran into the issue while I was trying on question "Using Popen in a thread blocks every incoming Flask-SocketIO request" from stackoverflow at link http://stackoverflow.com/questions/34599578/using-popen-in-a-thread-blocks-every-incoming-flask-socketio-request.

temoto:
Now `except ImportError` is gone and any exceptions are propagated as in normal Python code. I hope it helps with your problem too. Please reopen if not.

Issue #282, opened 2016-01-12
eventlet.green.zmq ignores RCVTIMEO setting


talwrii:
RCVTIMEO is documented here: http://api.zeromq.org/2-2:zmq-setsockopt

Here is a repro for you: 

```
from eventlet.green import zmq # broken with this
#import zmq # works with this

sock = zmq.Context().socket(zmq.SUB)
sock.connect('tcp://127.0.0.1:10432')
sock.setsockopt(zmq.RCVTIMEO, 1000)
sock.recv() # hangs with eventlet zmq, times out with zmq
```

talwrii:
I've got a patch that fixes this...

Just writing some unit tetsts now (I wanted a bug number)

talwrii:
Mergey Merge Merge:

https://github.com/eventlet/eventlet/pull/283

temoto:
Thank you. Slightly modified version f657c22eff384439ed19ea61720dbe678de4cd51 of your patch is merged into master. It'll be released on PyPI soon.

PR #283, opened 2016-01-13
Resolve bug 282

talwrii:
See https://github.com/eventlet/eventlet/issues/282 for details.

### DIFF
--- a/eventlet/green/zmq.py
+++ b/eventlet/green/zmq.py
@@ -9,6 +9,7 @@ __zmq__ = __import__('zmq')
 from eventlet import hubs
 from eventlet.patcher import slurp_properties
 from eventlet.support import greenlets as greenlet
+from eventlet.green import time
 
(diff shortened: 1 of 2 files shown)


talwrii:
Well that's a little sad!

Getting tox to run successfully on my machine seems to be taking considerably longer than it took to write the patch in the first place. And getting old version of python working on new versions of ubuntu is more effort than it should be.

Perhaps a regular developer might be able to quickly find out what the issues are. The patch is very small after all.

temoto:
Sorry a whole year has passed. Thanks for patch.

Please, check out slight modification to your patch here c40759b887, authorship kept intact. Some time in future we'll translate hub timers from relative offsets to absolute deadline time. There were no API changes.

temoto:
Maybe patch author is angry for delay. Patch was merged into master f657c22eff384439ed19ea61720dbe678de4cd51

PR #75, opened 2014-01-23
Provide python logging compatibility in eventlet.wsgi

sdague:
Provide a compatibility layer in eventlet.wsgi that allows you
to hand a python logger as the log parameter instead of a
filehandle. This is done via an object wrapper so the logging
methods will only be used if the contents of the log parameter
look like a log, otherwise the previous behavior of just writing
to the logger is done instead.

(This is work in progress patch to make sure the approach is
ok. Can add unit tests in future revision.)

PR #302, opened 2016-02-22
wsgi: document compatibility with python `logging`

lbonn:
I found this capability by chance by looking at the source and tracking back the change (#75)
It would be nice if this behavior was better documented.

Feel free to re-formulate the docstring if needed.


temoto:
Thanks for improving documentation.

Issue #270, opened 2015-12-03
call sleep function and change system time


shoppon:
```python
from eventlet import sleep


def execute():
    while True:
        print("execute")
        sleep(10)


execute()

```

After the first time print "execute", change the system time manually, it won't print "execute" after 10 second.

it's because that the sleep function based on system time, not the cpu time.

How can I make it work after changing system time?

temoto:
Monotonic time patch was merged into master f92b49f77e073dad40a71858ef9149b543e07d67, should fix this issue.

Issue #132, opened 2014-08-27
Modification time forward, the timer failure


ljz888666555:
To take effect, for example, modify the local time forward one hour timer to wait one hour after

haypo:
FYI I backported the PEP 418 to Python 2 using ctypes:
https://bitbucket.org/haypo/misc/src/tip/python/pep418.py

temoto:
@haypo great, thank you.

temoto:
Monotonic time patch was merged into master f92b49f77e073dad40a71858ef9149b543e07d67, should fix this issue.

PR #303, opened 2016-03-03
Add hubs.default_clock and use time.monotonic

haypo:
- Add a new eventlet.hubs.default_clock variable to be able to choose
  the default clock for hubs.
- Modify hub classes to use default_clock as the default clock
- default_clock uses time.monotonic if available (Python 3.3 or
  newer), or time.time

The asyncio project also uses time.monotonic() by default.

For example information of monotonic clock, see the PEP 418:
https://www.python.org/dev/peps/pep-0418/

Note: I wrote and implemented the PEP 418 in Python 3.3.

### DIFF
--- a/doc/hubs.rst
+++ b/doc/hubs.rst
@@ -48,6 +48,16 @@ MAINLOOP is launched only when the first I/O operation happens, and it is not th
 More Hub-Related Functions
 ---------------------------
 
+.. data:: default_clock
+
+   Default clock function used by hubs. On Python 3.3 and newer, use
+   :func:`time.monotonic` by default, otherwise use :func:`time.time`.
+
+   A monotonic clock is recommended.
+
+   .. versionadded:: 0.18.5
+
+
 .. autofunction:: eventlet.hubs.get_hub
 .. autofunction:: eventlet.hubs.get_default_hub
 .. autofunction:: eventlet.hubs.trampoline
--- a/eventlet/hubs/__init__.py
+++ b/eventlet/hubs/__init__.py
@@ -6,8 +6,15 @@ from eventlet.support import greenlets as greenlet, six
 
 __all__ = ["use_hub", "get_hub", "get_default_hub", "trampoline"]
 
+time = patcher.original('time')
 threading = patcher.original('threading')
 _threadlocal = threading.local()
+
+
+if hasattr(time, 'monotonic'):
+    default_clock = time.monotonic
+else:
+    default_clock = time.time
 
 
 def get_default_hub():
--- a/eventlet/hubs/epolls.py
+++ b/eventlet/hubs/epolls.py
@@ -25,6 +25,7 @@ else:
                     " NOT http://pypi.python.org/pypi/pyepoll/. "
                     " easy_install pyepoll installs the wrong version.")
 
+from eventlet import hubs
 from eventlet.hubs.hub import BaseHub
 from eventlet.hubs import poll
 from eventlet.hubs.poll import READ, WRITE
@@ -34,7 +35,9 @@ from eventlet.hubs.poll import READ, WRITE
 
 
 class Hub(poll.Hub):
-    def __init__(self, clock=time.time):
+    def __init__(self, clock=None):
+        if clock is None:
+            clock = hubs.default_clock
         BaseHub.__init__(self, clock)
         self.poll = epoll()
         try:
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -20,6 +20,7 @@ else:
         arm_alarm = alarm_signal
 
 from eventlet import patcher
+from eventlet import hubs
 from eventlet.hubs import timer, IOClosed
 from eventlet.support import greenlets as greenlet, clear_sys_exc_info
 time = patcher.original('time')
@@ -113,7 +114,9 @@ class BaseHub(object):
     READ = READ
     WRITE = WRITE
 
-    def __init__(self, clock=time.time):
+    def __init__(self, clock=None):
+        if clock is None:
+            clock = hubs.default_clock
         self.listeners = {READ: {}, WRITE: {}}
         self.secondaries = {READ: {}, WRITE: {}}
         self.closed = []
--- a/eventlet/hubs/kqueue.py
+++ b/eventlet/hubs/kqueue.py
@@ -7,6 +7,7 @@ time = patcher.original('time')
 sleep = time.sleep
 
 from eventlet.support import clear_sys_exc_info
+from eventlet import hubs
 from eventlet.hubs.hub import BaseHub, READ, WRITE, noop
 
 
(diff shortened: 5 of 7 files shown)


haypo:
Workaround proposed in OpenStack (in oslo.service): https://review.openstack.org/286838

temoto:
:+1: 

Relevant issues: #270 #132 

Do you have info about monotonic for 2.7? I only found these:

- http://stackoverflow.com/a/1205762/73957
- https://pypi.python.org/pypi/Monoclock

The idea to use monotonic clock is of course the only way to go. Huge thanks for bringing this up, and patch.
I disagree with `default_clock` though, it's an additional proxy for Python kwarg default value `Hub(clock=X)`. I will rewrite it, please don't be angry. :-)

haypo:
> Do you have info about monotonic for 2.7?

OpenStack uses https://pypi.python.org/pypi/monotonic

I backported the PEP 418 to Python 2, but I didn't publish it on PyPI, since most users only care of time.monotonic() and there are already projects providing a monotonic clock for Python 2.
https://bitbucket.org/haypo/misc/src/tip/python/pep418.py

> I disagree with default_clock though, it's an additional proxy for Python kwarg default value Hub(clock=X).

Sorry, I don't understand. The idea of default_clock is to not add a new dependency to eventlet, and let users choose the clock.

Right now, there is no way to choose the clock for all threads. See for example https://review.openstack.org/#/c/286838/ which looks a little bit to a hack to me, that's why I proposed to add default_clock option.

My default_clock option is a compromise for a practical issue, the need of an external dependency.

Well. Another option is to integrate a monotonic clock to eventlet directly. But I have no preference, they are various implementations which advantages and drawbacks. So I prefer to continue to use a 3rd party module for that.

haypo:
Python libraries providing a monotonic clock:

* https://github.com/gavinbeatty/python-monotonic-time
* https://github.com/ludios/Monoclock
* https://pypi.python.org/pypi/monotonic

temoto:
There is a way to choose clock for all threads with a small risk.
```
# somewhere close to program initialization, where they usually put monkey_patch()
import time, eventlet.hubs
if hasattr(time, 'monotonic') and eventlet.hubs.get_hub().clock is time.time:
  eventlet.hubs.get_hub().clock = time.monotonic
```
There's no way you could set `eventlet.hubs.default_clock` and be sure that incorrect clock wasn't used already. So you always take risk, but putting runtime assign workaround at top or in `__main__` is lowest risk.

Another way to restate my previous attack on `default_clock`: it gives programmer a way to **customise** **default** clock function. Immediately, you can see oxymoron here. You don't customise defaults. It works backwards. Default is used when you didn't customise.

Better option to let users choose clock, without risk, with hard guarantee that incorrect clock wasn't used is environment variable.
```
EVENTLET_CLOCK=time.monotonic python ...
```
And your version with `EVENTLET_HUB=service_hub` in Openstack is exactly this, working great. It uses documented environment API to set hub's public attribute. Perfectly correct code. And kudos for `service_hub` being just a function.

Thanks for monotonic libraries list. I will insert one into Eventlet source, the practice is already established with `mock` and `six`.

temoto:
Do you practical experience on `CLOCK_MONOTONIC` vs `CLOCK_MONOTONIC_RAW`? From what I read on stackoverflow, the RAW version should be used for timers.

haypo:
> Do you practical experience on CLOCK_MONOTONIC vs CLOCK_MONOTONIC_RAW? From what I read on stackoverflow, the RAW version should be used for timers.

Please don't use CLOCK_MONOTONIC_RAW. You really want CLOCK_MONOTONIC which is adjusted to use *seconds*. For time.monotonic(), it was important that the unit is second, since users expect this unit. It's a good thing to adjust CLOCK_MONOTONIC with NTP to really measure *seconds* and not an arbitrary unit (slower or faster than seconds).

CLOCK_MONOTONIC_RAW must be reserved for low-level applications synchronizing time between multiple computers.

haypo:
When I wrote this issue, I was thinking at asyncio where the event loop is created explicitly. It looks like the creation of the eventlet is more implicit, and it's harder to control when it is created. So I'm not sure that my design (default_clock) works to configure a clock.

The problem is to set the clock function *before* the first event loop is created.

temoto:
Where did you got this anti-raw information?

haypo:
Linux clock_gettime() manual page:

```
       CLOCK_MONOTONIC_RAW (since Linux 2.6.28; Linux-specific)
              Similar to CLOCK_MONOTONIC, but provides access to a raw
              hardware-based time that is not subject to NTP adjustments or
              the incremental adjustments performed by adjtime(3).
```

NTP is used to *fix* (adjust) clocks to get seconds instead of a clock slower or faster than atomic clocks using seconds.

Other explanations: http://stackoverflow.com/questions/14270300/what-is-the-difference-between-clock-monotonic-clock-monotonic-raw

NTP adjustement is a *good* thing ;-)

Or: "[CLOCK_MONOTONIC] ... the timestamps from the hardware clock source are being scaled, per adjustments made via adjtime/NTP, to correct for imprecision in the hardware clock rate that would otherwise make it drift relative to real time." http://stackoverflow.com/questions/25583498/clock-monotonic-vs-clock-monotonic-raw-truncated-values

temoto:
Seems true, thank you for educating me on this matter.

On Mon, Apr 25, 2016 at 9:01 PM, Victor Stinner <notifications@github.com>
wrote:

> Linux clock_gettime() manual page:
>
>        CLOCK_MONOTONIC_RAW (since Linux 2.6.28; Linux-specific)
>               Similar to CLOCK_MONOTONIC, but provides access to a raw
>               hardware-based time that is not subject to NTP adjustments or
>               the incremental adjustments performed by adjtime(3).
>
> NTP is used to *fix* (adjust) clocks to get seconds instead of a clock
> slower or faster than atomic clocks using seconds.
>
> Other explanations:
> http://stackoverflow.com/questions/14270300/what-is-the-difference-between-clock-monotonic-clock-monotonic-raw
>
> NTP adjustement is a *good* thing ;-)
>
> Or: "[CLOCK_MONOTONIC] ... the timestamps from the hardware clock source
> are being scaled, per adjustments made via adjtime/NTP, to correct for
> imprecision in the hardware clock rate that would otherwise make it drift
> relative to real time."
> http://stackoverflow.com/questions/25583498/clock-monotonic-vs-clock-monotonic-raw-truncated-values
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/eventlet/eventlet/pull/303#issuecomment-214417439>
>



-- 
Sergey Shepelev
Skype: sergey.shepelev
+79996126031
https://temoto.ru/

haypo:

temoto:
@haypo similar patch was merged, Eventlet is monotonic by default now. I hope it can shred some burden of patching hubs in OpenStack too.

haypo:
Very good news! I don't really care of the exact implementation, I only
care that monotonic clock is used by default to prevent bad surprises on
whiter/summer timezone switch (DST).

PR #309, opened 2016-04-07
Patch logging._lock

lxsli:
In projects which dynamically determine whether to activate eventlet,
it can be hard not to import a low level module like logging before
eventlet. When logging is imported it initialises a threading.RLock
which it uses to protect the logging configuration. If two
greenthreads attempt to claim this lock, the second one will block the
/native/ thread not just itself.

Replace the unsafe RLock with one which relies on a green Semaphore
while monkey-patching.

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -330,26 +330,49 @@ def _green_existing_locks():
     blocks the native thread. We need to replace these with green Locks.
 
     This was originally noticed in the stdlib logging module."""
-    if sys.version[0] != '2':
-        return
     import gc
     import threading
     import eventlet.green.thread
-    import eventlet.green.threading
     lock_type = type(threading.Lock())
     rlock_type = type(threading.RLock())
+    if sys.version_info[0] >= 3:
+        pyrlock_type = type(threading._PyRLock())
     # We're monkey-patching so there can't be any greenlets yet, ergo our thread
     # ID is the only valid owner possible.
     tid = eventlet.green.thread.get_ident()
-    for o in gc.get_objects():
-        if isinstance(o, rlock_type) and isinstance(o._RLock__block, lock_type):
-            orig = o._RLock__block
-            new = eventlet.green.threading.Lock()
-            o._RLock__block = new
-            if orig.locked():
-                new.acquire()
-            o._RLock__owner = tid
-            # TODO test re-locking and owner
+    for obj in gc.get_objects():
+        if isinstance(obj, rlock_type):
+            if (sys.version_info[0] == 2 and
+                    isinstance(obj._RLock__block, lock_type)):
+                _fix_py2_rlock(obj, tid)
+            elif (sys.version_info[0] >= 3 and
+                    not isinstance(obj, pyrlock_type)):
+                _fix_py3_rlock(obj)
+
+
+def _fix_py2_rlock(rlock, tid):
+    import eventlet.green.threading
+    old = rlock._RLock__block
+    new = eventlet.green.threading.Lock()
+    rlock._RLock__block = new
+    if old.locked():
+        new.acquire()
+    rlock._RLock__owner = tid
+
+
+def _fix_py3_rlock(old):
+    import gc
+    import threading
+    new = threading._PyRLock()
+    while old._is_owned():
+        old.release()
+        new.acquire()
+    if old._is_owned():
+        new.acquire()
+    for ref in gc.get_referrers(old):
+        for k, v in vars(ref):
+            if v == old:
+                setattr(ref, k, new)
 
 
 def _green_os_modules():
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -250,6 +250,9 @@ def monkey_patch(**on):
             on.setdefault(modname, False)
         on.setdefault(modname, default_on)
 
+    if on['thread'] and not already_patched.get('thread'):
+        _green_existing_locks()
+
     modules_to_patch = []
     for name, modules_function in [
         ('os', _green_os_modules),
@@ -318,6 +321,19 @@ def is_monkey_patched(module):
     module."""
     return module in already_patched or \
         getattr(module, '__name__', None) in already_patched
+
+
+def _green_existing_locks():
+    if sys.version[0] != '2':
+        return
+    import threading
+    import eventlet.green.threading
+    lock_type = type(threading.Lock())
+    rlock_type = type(threading.RLock())
+    gc = __import__('gc')
+    for o in gc.get_objects():
+        if isinstance(o, rlock_type) and isinstance(o._RLock__block, lock_type):
+            o._RLock__block = eventlet.green.threading.Lock()
 
 
 def _green_os_modules():
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -334,14 +334,22 @@ def _green_existing_locks():
         return
     import gc
     import threading
+    import eventlet.green.thread
     import eventlet.green.threading
     lock_type = type(threading.Lock())
     rlock_type = type(threading.RLock())
+    # We're monkey-patching so there can't be any greenlets yet, ergo our thread
+    # ID is the only valid owner possible.
+    tid = eventlet.green.thread.get_ident()
     for o in gc.get_objects():
         if isinstance(o, rlock_type) and isinstance(o._RLock__block, lock_type):
-            o._RLock__block = eventlet.green.threading.Lock()
-            # TODO: lock this lock if the original was locked
-            # TODO: Set o._RLock__owner
+            orig = o._RLock__block
+            new = eventlet.green.threading.Lock()
+            o._RLock__block = new
+            if orig.locked():
+                new.acquire()
+            o._RLock__owner = tid
+            # TODO test re-locking and owner
 
 
 def _green_os_modules():
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -369,6 +369,7 @@ def _fix_py3_rlock(old):
         new.acquire()
     if old._is_owned():
         new.acquire()
+    gc.collect()
     for ref in gc.get_referrers(old):
         for k, v in vars(ref):
             if v == old:
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -324,16 +324,24 @@ def is_monkey_patched(module):
 
 
 def _green_existing_locks():
+    """Make locks created before monkey-patching safe.
+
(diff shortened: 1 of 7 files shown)


tests/patcher_test.py
```diff
@@ -498,6 +498,32 @@ def test_is_daemon(self):
...
 
+class Logging(ProcessBase):
+    def test_logging_early(self):
+        new_mod = """import eventlet
+eventlet.monkey_patch()
+import logging
+print(logging._lock._RLock__block.__class__)
+"""
+        self.write_to_tempfile("newmod", new_mod)
+        output, lines = self.launch_subprocess('newmod')
```
temoto: Please, forget `write_to_tempfile` and `launch_subprocess` ever existed. Current supported testing API is `tests.run_python` and `tests.run_isolated`.

temoto:
1. Thanks for patch, it's a good idea to help crippled modules
2. Right now main obstacle from merging is broken tests
3. I highly recommend avoid using `logging`. Because this lock is not the only problem you will face with `logging`. Go with `logbook` if you think you need fancy features. But in the end, the best logger is this:

```
def log(fmt, *args, **kwargs):
  fmt = datetime.utcnow().replace(microsecond=0).isoformat() + '\t' + fmt + '\n'
  sys.stderr.write(fmt.format(*args, **kwargs))
```

lxsli:
Thanks for the fast response! I've updated the tests to use run_isolated. I'm working on OpenStack, a very large open source project, so switching our log library would be a massive change. What other problems do you expect we'll see with logging and eventlet please? Anything you can tell me would be very helpful!

temoto:
It's not related to Eventlet. Next problem with logging is speed, next is sending big messages to syslog.

lxsli:
OK, that's good to know, thank you. I hope this version will pass CI!

temoto:
It doesn't with regard to Python3, you can run all tests locally using tox.

temoto:
Our CI configuration is purposedly complicated to run the one exact configuration local and on Travis.

lxsli:
The version before last failed on Py3, the latest just passed \o/

temoto:
I'm sorry to say about it so late, but your tests verify condition with knowledge (hope) of how internals work. I'm sure it's possible to verify actual locking problem by forging two synchronised log writes from different green threads. You can defer this task to me for later.

lxsli:
That's fair feedback, let me make them less intrusive and see what you think.

lxsli:
I hope this is OK, although I still call private methods of logging (_acquireLock and _releaseLock) this is far less intrusive than the old approach and guarantees the lock will be re-entered. Achieving such a guarantee with public methods would require equally intrusive mocking to force a yield.

temoto:
Now there's no testing at all.

`spawn ; sleep ; pass`  is semantically equivalent to  `sleep ; pass`  which is just `pass` after 1 second. Sleeping some time is not a method of synchronisation. It could technically only run spawned functions after `pass` and it would be strange but valid behavior.

lxsli:
The sleep causes a yield which lets the greenlets execute. You're right in that another process could hog all CPU so that after a second, instead of B waking, main wakes. I'll swap the time.sleep(1) for "a.wait(); b.wait()" to fix this.

temoto:
There's still no intersync between greenthreads. It's valid to schedule them as A.start - A.wait - B.start - B.wait. And you're still verifying locks, not the behavior.

lxsli:
I thought eventlet provided a cooperative yielding model, where greenlets cannot be interrupted unless they yield? If I read your example correctly, after A starts it is interrupted without yielding. Please can you explain this seeming contradiction?

lxsli:
Also logging messages has nothing to do with eventlet. What we need to ensure is that the contract of the lock this module relies on is preserved in the presence of eventlet and that is exactly what I am testing.

temoto:
1. Cooperative is exactly what you said, but less strict than what you expect in test. `spawn(f, *a, **kw)` is a contract to call `f(*a, **kw)` will be called in a new greenthread. There is no contract about time or order of greenthread execution for several `spawn`.

```
a = spawn(f1)
b = spawn(f2)
a.wait()  # you did not wait for f2 to begin
b.wait()
```

2. Locks is what you found after exploring internals of logging. How did the problem manifest in the first place?

lxsli:
Morning, hope you had a good weekend.

1. Ah, it works if neither are started immediately but not if one is. Fixed.
2. Forget about logging, that was just the breadcrumb trail that led to the problem. Eventlet can't test every piece of code it transitively impacts. I've generalised the solution to make this clear, using some inspiration from gevent.

eventlet/patcher.py
```diff
@@ -320,6 +323,18 @@ def is_monkey_patched(module):
         getattr(module, '__name__', None) in already_patched
 
 
+def _green_existing_locks(threading):
+    if sys.version[0] != '2':
```
temoto: Why is this only for Python2?

eventlet/patcher.py
```diff
@@ -320,6 +323,18 @@ def is_monkey_patched(module):
...
+        return
+    import eventlet.green.threading
+    lock_type = type(threading.Lock())
+    rlock_type = type(threading.RLock())
+    gc = __import__('gc')
```
temoto: Two different ways to import in single function. Please, unify to `import` or leave a comment in code why it was necessary.

tests/__init__.py
```diff
@@ -12,7 +12,10 @@
 except ImportError:
     resource = None
 import signal
-import subprocess
+try:
+    import subprocess32 as subprocess  # py2
+except:
```
temoto: `except:` in Python2 is only allowed together with comment why you need to catch `SystemExit`, `KeyboardInterrupt`, `GeneratorExit `. Otherwise in Python2 it must be `except Exception`. In this particular case, only `except ImportError`.

tests/__init__.py
```diff
@@ -12,7 +12,10 @@
 except ImportError:
     resource = None
 import signal
-import subprocess
+try:
+    import subprocess32 as subprocess  # py2
```
temoto: While `run_python` would definitely benefit from timeout, as of now it breaks Python2.6 tests. I suggest to retract subprocess32 for now, if a simple solution is not possible, to avoid blocking of merging this threading patch.

If you really insist on subprocess32, it would safely work backwards:
```
import subprocess
if six.PY2:
  # temporary workaround `run_python` needs `.communicate()` with timeout
  import subprocess32
elif six.PY3:
  subprocess32 = subprocess
```

tests/__init__.py
```diff
@@ -320,7 +323,12 @@ def run_python(path, env=None, args=None):
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
     )
-    output, _ = p.communicate()
+    try:
+        output, _ = p.communicate(timeout=10)
+    except subprocess.TimeoutExpired:
+        p.kill()
+        output, _ = p.communicate(timeout=10)
+        return "%s\nFAIL - timed out" % output
```
temoto: Please, avoid `%` interpolation when possible. Use `str.format`, e.g. `'{0}\n\nfail - timeout'.format(output)`. Besides new features it also doesn't have string/tuple ambiguity. `%` single argument must be enclosed in tuple when it comes from external code.

tests/isolated/patcher_existing_locks_early.py
```diff
@@ -0,0 +1,30 @@
+import threading
```
temoto: It would be more clear to keep all imports together if possible.

tests/patcher_test.py
```diff
@@ -498,6 +498,14 @@ def test_is_daemon(self):
         self.assertEqual(lines[1], "True", lines[1])
 
 
+class ExistingLocks(ProcessBase):
```
temoto: Please, avoid `ProcessBase` and class based tests in general in new code.

eventlet/patcher.py
```diff
@@ -320,6 +323,18 @@ def is_monkey_patched(module):
...
+    for o in gc.get_objects():
+        if isinstance(o, rlock_type) and isinstance(o._RLock__block, lock_type):
+            o._RLock__block = eventlet.green.threading.Lock()
```
temoto: Not sure, but this seems like a disaster if somebody holds a lock when `monkey_patch()` is called.

temoto:
Thanks for starting a new epoch in monkey patching. This gc-enumerate approach will be useful for other symbols too.

It would help to add description of original problem, why `green_existing_locks` is useful and why it's useful only on Python2. Whether it's described in patcher or tests is not important.

lxsli:
"Why is this only for Python2?" - because the test passes on Py3 without the change :/

"timeout, as of now it breaks Python2.6 tests" - that was a different issue, Py2.6 tests are passing now.

"this seems like a disaster if somebody holds a lock when monkey_patch() is called." - yes, that's definitely a concern. I'll work on that tomorrow. I'll add some description too.

Other comments will be fixed in next patchset.

temoto:
> "Why is this only for Python2?" - because the test passes on Py3 without the change :/

Yeah test pass is good, but do you have idea why?

lxsli:
No, I can look into that later. For now it's not my primary concern since it's no more broken than it used to be.

lxsli:
Needs tests

tests/__init__.py
jstasiak: Relevant flake8 plugin: https://pypi.python.org/pypi/flake8-blind-except

temoto:
Thanks for `flake8-blind-except` reference, that's good to know.

lxsli:
Well these test failures are a mystery to me, any hints appreciated!

temoto:
We still have a little of sporadic errors depending on Travis hardware load. Ignore those.

temoto:
Tests have become too complicated for quick glance understanding, so please explain this:
```
def take(lock, sync1, sync2):
    sync2.acquire()
    sync1.release()
    with lock:
        sync2.release()
# ...
    sync1 = threading.Lock()
    sync2 = threading.Lock()
    sync1.acquire()
    eventlet.spawn(take, lock, sync1, sync2)
    # Ensure sync2 has been taken
    with sync1:
        pass
```

Comment says ensure sync2 taken, but next code takes and immediately releases sync1, possibly without switching to greenthread `take`.

lxsli:
sync1 is a non-reentrant Lock which is acquired before the spawn. So to
re-acquire it, 'take' must have released it, which doesn't happen until
after sync2 has been taken.

temoto:
Thanks, it makes sense now.

lxsli:
How's this doing please?

temoto:
Merged into master, thank you.

lxsli:
Fantastic, thank you!

Issue #256, opened 2015-10-19
Need a new release for fix io.open().read() on Python 3


haypo:
Hi,

The pull request #239 adds a missing readall() method to GreenFileIO. This is required to port the OpenStack Swift project to Python 3.

It would also be nice to have the a release with the fix #248.

I created a Git branch greenfileio but then I saw that my issue was already fixed :-) It's already possible to call GreenFileIO.read() without argument: readall() is called in this case, as io.FileIO.read() does (without eventlet monkey patching).

Thanks.

haypo:
Hum, it looks like eventlet master branch doesn't work for my use case in fact :-/ readall() uses trampoline(self, read=True) but the file is a regular file. In this case, epoll_ctl() fails with EPERM because regular files cannot be handled by epoll on Linux.

haypo:
> Hum, it looks like eventlet master branch doesn't work for my use case in fact :-/

Here is a fix: pull request #257: " Fix GreenFileIO.readall() for regular file".

haypo:
Many eventlet versions were released with the fix, so this issue can now be fixed ;-)

PR #314, opened 2016-05-16
only trampoline when we block

davidszotten:
### update

bugfix: only trampoline after catching a "blocking" ioerror. trampolining might otherwise fail, e.g. if using epoll and the fd happens to be a regular file (would never block)

test thanks to @haypo from #257

### original description, kept for posterity

wasn't able to figure out how to reproduce in a test, but can be trigged with a pytest test:

```
import eventlet
eventlet.monkey_patch()

pytest_plugins = "pytester"


def test_foo(testdir):

    testdir.makepyfile(
        """
        def test_foo():
            print('foo')
        """
    )
    result = testdir.runpytest()
    assert result.ret == 0
```

hangs on py3 without this patch

### DIFF
--- a/eventlet/greenio/py3.py
+++ b/eventlet/greenio/py3.py
@@ -85,7 +85,7 @@ class GreenFileIO(_OriginalIOBase):
             except OSError as e:
                 if get_errno(e) not in SOCKET_BLOCKING:
                     raise IOError(*e.args)
-            self._trampoline(self, read=True)
+                self._trampoline(self, read=True)
 
     def readall(self):
         buf = []
@@ -98,7 +98,7 @@ class GreenFileIO(_OriginalIOBase):
             except OSError as e:
                 if get_errno(e) not in SOCKET_BLOCKING:
                     raise IOError(*e.args)
-            self._trampoline(self, read=True)
+                self._trampoline(self, read=True)
 
     def readinto(self, b):
         up_to = len(b)
--- a/tests/isolated/regular_file_readall.py
+++ b/tests/isolated/regular_file_readall.py
@@ -1,0 +1,41 @@
+import eventlet; eventlet.monkey_patch()
+
+from eventlet.support import six
+import io
+import os
+import tempfile
+
+
+if __name__ == "__main__":
+    with tempfile.NamedTemporaryFile() as tmp:
+        with io.open(tmp.name, "wb") as fp:
+            fp.write(b"content")
+
+        # test BufferedReader.read()
+        fd = os.open(tmp.name, os.O_RDONLY)
+        fp = os.fdopen(fd, "rb")
+        with fp:
+            content = fp.read()
+        assert content == b'content'
+
+        # test FileIO.read()
+        fd = os.open(tmp.name, os.O_RDONLY)
+        fp = os.fdopen(fd, "rb", 0)
+        with fp:
+            content = fp.read()
+        assert content == b'content'
+
+        if six.PY3:
+            # test FileIO.readall()
+            fd = os.open(tmp.name, os.O_RDONLY)
+            fp = os.fdopen(fd, "rb", 0)
+            with fp:
+                content = fp.readall()
+            assert content == b'content'
+
+        # test FileIO.readall() (for Python 2 and Python 3)
+        with io.open(tmp.name, "rb", 0) as fp:
+            content = fp.readall()
+        assert content == b'content'
+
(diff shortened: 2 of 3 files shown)


eventlet/greenio/py3.py
```diff
@@ -85,7 +85,7 @@ def read(self, size=-1):
             except OSError as e:
                 if get_errno(e) not in SOCKET_BLOCKING:
                     raise IOError(*e.args)
-            self._trampoline(self, read=True)
```
davidszotten: this works either way, but i think having it inside the `except` clause is more semantically correct (and copying this is what caused the error below)

temoto:
Thanks for patch.
How to read this test? I fail to understand what exactly does it verify.

davidszotten:
do you mean `test_foo`? That's distilled from a test testing a pytest plugin. Without the patch in this pr, running that test the runner just hangs. The patch stops the hanging.

Tried but failed to trigger the same issue from the eventlet test suite. Not suggesting the above should be added, just included it for reference if anyone else wants to try to to reproduce it.

davidszotten:
do you need anything else here?

temoto:
Not with code, but field testing would be nice.

davidszotten:
is that something you are looking for from me?

jstasiak:
I think we should have a reliable unit test included in the same patch as the issue seems nontrivial.

davidszotten:
@jstasiak as i mentioned, i haven't been able to reproduce in a test scenario. i can trigger the bug using pytest and the minimal example in the description. any help writing a test would be appreciated

temoto:
I take test.

Anybody willing to run this version on their systems - much appreciated.

davidszotten:
To add another datapoint: I had been testing this on os x, which is where pytest hangs without this patch. On linux i instead get a `Second simultaneous read` exception.

davidszotten:
So if i understand correctly, this change just disturbs things enough (in my particular case) to avoid a bug elsewhere. In that case, this is a simple performance improvement, and the existing test should be sufficient.

davidszotten:
ok. some more digging: the simultaneous read might be a red herring (or an effect of something). hadn't noticed there was another exception being raised earlier: 

```
>                   self.poll.register(fileno, mask)
E                   PermissionError: [Errno 1] Operation not permitted
```

looks like we are getting

>  EPERM  The target file fd does not support epoll.  This error occur if fd refers to, for example, a regular file or directory.

travis run:
https://travis-ci.org/mattbennett/eventlet-pytest-bug/jobs/133923901

do you know what causes this?

davidszotten:
ok, now that i have something better to look for, i guess i'm just hitting #256, fixed by #257 (merge conflicts are just that multiple tests were added in `tests/patcher_test.py`.  happy to rebase that branch and raise a new pr if you prefer?

could still (separately) take the slight perf improvement in this pr, though that's less important

temoto:
Thanks for investigation. Let's put this trampoline optimisation on hold for a while and continue with #257.

davidszotten:
so i'm back to thinking this is the proper fix. apologies for the amount of back and forth, but i think my understanding of what's going on is slowly improving. i've incorporated the test from #257 which passes with this fix (and fails without it).

davidszotten:
hi. do you need anything else from me to move this forward?

mattbennett:
It would be great to see a fix for this, or #257

tests/isolated/regular_file_readall.py
```diff
@@ -0,0 +1,42 @@
+import eventlet
+eventlet.monkey_patch()
```
temoto: Please, follow this pattern for isolated tests:
```
__test__ = False

if __name__ == '__main__':
    import eventlet
    # and everything else
```

`monkey_patch()` at module level is a hidden bomb in a file that is potentially imported by test runner collector.

davidszotten: thanks, hadn't noticed that the standards had moved since that test was written. updated now

temoto:
Field testing is very welcome.

Thanks, @davidszotten

davidszotten:
thanks. [nameko](https://github.com/onefinestay/nameko/) is a decent sized open source library that makes heavy use of eventlet. there is already a travis job that runs against the latest released version of eventlet, but if it would help we could probably add one against eventlet master (with failure allowed)

temoto:
Yes, please! CI of different code against master would be super awesome.

Issue #293, opened 2016-01-24
Green select.poll is required for subprocess


temoto:
https://github.com/eventlet/eventlet/commit/bc4d1b5d362e5baaeded35b1e339b9db08172dd2 deletes blocking version of `select.poll` and that is good.
But that breaks `subprocess .communicate()` under certain valid order of import and `monkey_patch()`. #290 
I think proper solution is to have green version of `select.poll()`.

PR #317, opened 2016-05-18
Remove select.poll and improve subprocess

jstasiak:
@temoto Could you have a look at this? I'm not comfortable pushing to master directly as the subprocess module wasn't monkey patched at all until this change, there may be a reason for that.

This closes https://github.com/eventlet/eventlet/issues/293

### DIFF
--- a/eventlet/green/subprocess.py
+++ b/eventlet/green/subprocess.py
@@ -115,7 +115,17 @@ class Popen(subprocess_orig.Popen):
         except AttributeError:
             pass
 
+
 # Borrow subprocess.call() and check_call(), but patch them so they reference
 # OUR Popen class rather than subprocess.Popen.
-call = FunctionType(six.get_function_code(subprocess_orig.call), globals())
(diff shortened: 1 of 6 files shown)


jstasiak:
Ignore the broken build, I forgot about few things, will fix soon.

eventlet/green/subprocess.py
```diff
@@ -9,14 +9,15 @@
...
+__patched__ = ['call', 'check_call', 'check_open', 'Popen']
 to_patch = [('select', select), ('threading', threading), ('time', time)]
 
 if sys.version_info > (3, 4):
     from eventlet.green import selectors
     to_patch.append(('selectors', selectors))
 
 patcher.inject('subprocess', globals(), *to_patch)
-subprocess_orig = __import__("subprocess")
+subprocess_orig = patcher.original("subprocess")
```
temoto: 👍

jstasiak:
After few iterations this is in a mergeable state now I believe.

temoto:
Thank you very much.

PR #322, opened 2016-06-02
Fix IPv6 LLA handling

jjohnson42:
getaddrinfo would fail with LLA with scope index.
Discard any '%' delimited suffix.


jjohnson42:
I have no idea how the indicated test failures could relate to this tiny change...

temoto:
Thanks.
I've added test to your commit.
It's merged into master ac4e28f

PR #324, opened 2016-06-08
Add get_server_info to db_pool.py

ernstp:
Django uses this for example.

### DIFF
--- a/eventlet/db_pool.py
+++ b/eventlet/db_pool.py
@@ -336,6 +336,7 @@ class GenericConnectionWrapper:
         'errno',
         'error',
         'errorhandler',
+        'get_server_info',
         'insert_id',
         'literal',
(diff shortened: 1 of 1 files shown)


4383 (approved):
Hello,

Is this patch still relevant?
AFAICS these changes make sense.

ernstp:
Wow, that was a long time ago! :-D

I guess step one is to rebase this! Feel free to do that and push and so on if you have time... I might have time to look at it in the weekend otherwise.

4383:
hm... something went wrong during your rebase. All tests are failing and your addition does not seems in the right place in the code like before. Please check

ernstp:
I guess it should be like this now.. ?

4383:
Looks better, thanks

PR #326, opened 2016-06-10
Commiting changes to eventlet/greenio/base.py to fix the pypy bad fil…

ysingh7:
…e descriptor error

### DIFF
--- a/eventlet/greenio/base.py
+++ b/eventlet/greenio/base.py
@@ -299,10 +299,10 @@ class GreenSocket(object):
             res = _python2_fileobject(dupped, *args, **kwargs)
             if hasattr(dupped, "_drop"):
                 dupped._drop()
-                """ Making the close function of dupped None so that when garbage collector
-                kicks in and tries to call del, which will ultimately call close, _drop
-                doesn't get called on dupped twice as it has been already explicitly called in
-                previous line"""
+                # Making the close function of dupped None so that when garbage collector
+                # kicks in and tries to call del, which will ultimately call close, _drop
+                # doesn't get called on dupped twice as it has been already explicitly called in
+                # previous line
                 dupped.close = None
             return res
 
--- a/eventlet/greenio/base.py
+++ b/eventlet/greenio/base.py
@@ -299,8 +299,9 @@ class GreenSocket(object):
             res = _python2_fileobject(dupped, *args, **kwargs)
             if hasattr(dupped, "_drop"):
                 dupped._drop()
-                """ Making the close function of dupped None so that when garbage collector kicks in and tries to call del, which
-                will ultimately call close, _drop doesn't get called on dupped twice as it has been already explicitly called in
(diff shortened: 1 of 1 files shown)


clayg:
I don't think this is the right approach.  Neutering the dupped sockets close method will obviously prevent the garbage collector from calling `close` on the fd, but that's just papering over the issue.

I don't think pypy generally expects you to use `_drop` without `_reuse` - the two go hand-in-hand

Take a look here:

https://github.com/gevent/gevent/blob/80f014eed210d01b60802863d3cb0d3d38cc61ff/src/gevent/_socket2.py#L119

Anytime gevent creates a new reference to a socket from another socket that was passed in it increments the pypy refcount with `_reuse` - therefore when the pypy runtime gc's a gevent `socket.socket` and calls it's `_drop` implementation the refcount won't drop to zero until the *other* reference(s) to real socket (who incremented the ref count with `_reuse`) *also* get's gc'd.

FWIW, I don't think that's the best approach either - the gevent code also has to be *very* careful to call `_drop` anytime it returns a gevent `socket.socket` that was created from a real socket in order to keep up with the refcount increment'd with `_reuse` when it was created [1].

IMHO, the better solution to *this* bug I think is to just not call `dup` when we call makefile.  I really can't see how to make that API work correctly.

 * The CPython implementation doesn't call `dup` [2]
 * The py3 implementation in eventlet doesn't call `dup` [3]
 * gevent doesn't call `dup` [4] 
 * PyPy itself doesn't do it [5]

It's not entirely clear to me why @alex deduced the call to `_drop` was correct?  It was probably *necessary* because `eventlet.wsgi` never calls `close` on the `rfile` and `wfile` - and the fact that eventlet's `makefile` calls `dup` on the underlying real socket means the refcount was too high to actually close the socket when it called `connection.close`?  But it seems to me calling `_drop` on the socket we give to makefile leaves it as a bit of a weak reference, meaning you'd have to hold onto the socket you `makefiled`'d from to keep your connection from being closed from underneath you.

1. https://github.com/gevent/gevent/blob/80f014eed210d01b60802863d3cb0d3d38cc61ff/src/gevent/_socket2.py#L200
2. https://github.com/python/cpython/blob/2243a7c789a5029610ee1349aa353ee59a9c3805/Lib/socket.py#L216
3. https://github.com/eventlet/eventlet/blob/863a1b7605b5c90ba470e597666c61a3261b706c/eventlet/greenio/base.py#L294
4. but again, it *does* have to be careful to call `_drop` because it creates a new `type(self)(sock)` which always calls `_reuse` - gah - https://github.com/gevent/gevent/blob/80f014eed210d01b60802863d3cb0d3d38cc61ff/src/gevent/_socket2.py#L261
5. https://bitbucket.org/pypy/pypy/src/5a0e839de8bae43a856cdce2d5dace5bb9850c95/lib-python/2.7/socket.py?at=default&fileviewer=file-view-default#socket.py-267

PR #327, opened 2016-06-10
wsgi: only skip Content-Type and Content-Length headers

tipabu:
Previously, some HTTP headers would not be transcribed to the WSGI environment provided to applications. This could be happen if either:

  * the header closely mirrors another CGI variable, such as `Request-Method` or `Path-Info`, or

  * the header closely mirrors a previous header but prefixed with "Http-", such as `Http-Foo` when a `Foo` header has already been received.

Now, only `Content-Type` and `Content-Length` will be skipped, and only because they were previously already added as `CONTENT_TYPE` and `CONTENT_LENGTH` keys.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -604,7 +604,8 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
         env['headers_raw'] = headers_raw = tuple((k, v.strip()) for k, v in headers)
         for k, v in headers_raw:
             k = k.replace('-', '_').upper()
-            if k in env:
+            if k in ('CONTENT_TYPE', 'CONTENT_LENGTH'):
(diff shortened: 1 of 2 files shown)


temoto:
Thanks for patch. I think there's nothing special in content headers either. Care to remove whole 'if'?

tipabu:
I'm not sure that's a good idea, as it places the same data in two locations. As an application writer, which should I use? As a middleware writer, which should I update? How should I resolve differences between the two?

By only putting this data in one location, we keep things unambiguous: to access the Content-Type header, use the CONTENT_TYPE env key. Yes, this is different from how you access most other headers, but it's what the PEP says to use.

temoto:
I'd say WSGI is a bad spec in first place so don't care about two locations, but maybe I'm getting old.

Argument seems reasonable. Merged in master. Thank you.

Issue #316, opened 2016-05-18
green package (more than one module level) weirdness


jstasiak:
Extracted from https://github.com/eventlet/eventlet/issues/267#issuecomment-204391271

A minimal code to reproduce (latest Eventlet and httplib2, Python 3.5.1):

```python
import httplib2
import eventlet.green.urllib

http = httplib2.Http()
http.request('http://example.org')
```

The result:
```
Traceback (most recent call last):
  File "test.py", line 5, in <module>
    print(http.request('http://example.org'))
  File "/Users/me/ve/35/lib/python3.5/site-packages/httplib2/__init__.py", line 1314, in request
    (response, content) = self._request(conn, authority, uri, request_uri, method, body, headers, redirections, cachekey)
  File "/Users/me/ve/35/lib/python3.5/site-packages/httplib2/__init__.py", line 1064, in _request
    (response, content) = self._conn_request(conn, request_uri, method, body, headers)
  File "/Users/me/ve/35/lib/python3.5/site-packages/httplib2/__init__.py", line 1047, in _conn_request
    response = Response(response)
  File "/Users/me/ve/35/lib/python3.5/site-packages/httplib2/__init__.py", line 1382, in __init__
    for key, value in info.items():
AttributeError: 'HTTPResponse' object has no attribute 'items'
```

The relevant bit of code from ``httplib2/__init__.py``:

```python
        if isinstance(info, http.client.HTTPResponse):  # <- this condition fails although it should not
            for key, value in info.getheaders():
                key = key.lower()
                prev = self.get(key)
                if prev is not None:
                    value = ', '.join((prev, value))
                self[key] = value
            self.status = info.status
            self['status'] = str(self.status)
            self.reason = info.reason
            self.version = info.version
        elif isinstance(info, email.message.Message):
            for key, value in list(info.items()):
                self[key.lower()] = value
            self.status = int(self['status'])
        else:
            for key, value in info.items():  # <- the first condition doesn't match so this is executed and fails
                self[key.lower()] = value
            self.status = int(self.get('status', self.status))
```

I think that's because existing, already imported ``http.client`` module is indirectly replaced when ``eventlet.green.urllib`` is imported and this breaks things that already saw the original ``http.client`` module, see this:

```python
import http.client
print(['before green import', id(http.client)])
import eventlet.green.urllib
print(['after green import', id(http.client)])
```

The result:
```
['before green import', 4491451592]
['after green import', 4500059704]
```

[...]

PR #329, opened 2016-06-15
Provide green http without undesired side effects

jstasiak:
The Eventlet patcher and the way we were patching multi-level http
package don't work well[1][2]. I spent a lot of time trying to make it
work but in the end every solution I came up with was breaking something
else and made the patching and providing green http even more
complicated - I wouldn't envy anyone having to debug it in the future.

After a lot of thinking I decided having our own copy of http with the
necessary modifications applied seems like the most straightforward and
the most reliable solution, even considering its downsides (we need to
keep it up to date ourselves and the API won't 100 % match the regular
http module API on older Python 3 versions as our bundled version is the
most recent one and has bug fixes and extra features implemented).

The code introduces by this commit comes from the following Python
commit (development branch):

```
commit 6251d66ba9a692d3adf5d2e6818b29ac44130787
Author: Xavier de Gaye <xdegaye@users.sourceforge.net>
Date:   2016-06-15 11:35:29 +0200

    Issue #26862: SYS_getdents64 does not need to be defined on android
    API 21.
```

Changes to the originl http package code involve:

* Removing unnecessary import(s)
* Replacing some regular imports with eventlet.green imports
* Replacing fullmatch()[3] usage with match() so we stay Python 3.3
  compatible

I left urllib.parse imports intact as nothing there performs IO.

[1] https://github.com/getsentry/raven-python/issues/703
[2] https://github.com/eventlet/eventlet/issues/316
[3] https://docs.python.org/3/library/re.html#re.fullmatch

This addresses https://github.com/getsentry/raven-python/issues/703 fully and https://github.com/eventlet/eventlet/issues/316 partially.

@temoto Can you have a look? This addresses an issue that blocks us from porting one of our applications to Python 3.

### DIFF
--- a/eventlet/green/http/__init__.py
+++ b/eventlet/green/http/__init__.py
@@ -1,2 +1,191 @@
+# This is part of Python source code with Eventlet-specific modifications.
+#
+# Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+# 2011, 2012, 2013, 2014, 2015, 2016 Python Software Foundation; All Rights
+# Reserved
+#
+# PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+# --------------------------------------------
+#
+# 1. This LICENSE AGREEMENT is between the Python Software Foundation
+# ("PSF"), and the Individual or Organization ("Licensee") accessing and
+# otherwise using this software ("Python") in source or binary form and
+# its associated documentation.
+#
+# 2. Subject to the terms and conditions of this License Agreement, PSF hereby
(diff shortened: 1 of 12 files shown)


jstasiak:
There are some minor issues with the patch but I'm sorting them out.

jstasiak:
I think that's it, I just tested this patch against one of our applications and it works!

jstasiak:
@temoto Do you have any strong feelings either way here? I'm inclined to merge this sooner rather than later unless there are reasons not to.

jstasiak:
Just pushed to master as c93279086c21b9f838e1f9c9b232d60317fcfc71, if there are any concerns about this please let me know.

temoto:
LGTM. Hooray to more stdlib code inside ours.

PR #331, opened 2016-07-04
Added compatibility for websockets for gunicorn eventlet worker

catroot:
Now you able to upgrade connection with eventlet.websocket.WebSocketsWSGI when app runs under `gunicorn application:app -k eventlet`

### DIFF
--- a/eventlet/websocket.py
+++ b/eventlet/websocket.py
@@ -135,7 +135,12 @@ class WebSocketWSGI(object):
         return wsgi.ALREADY_HANDLED
 
     def _handle_legacy_request(self, environ):
-        sock = environ['eventlet.input'].get_socket()
+        if 'eventlet.input' in environ.keys():
+            sock = environ['eventlet.input'].get_socket()
+        elif 'gunicorn.socket' in environ.keys():
+            sock = environ["gunicorn.socket"]
+        else:
+            raise("No knows sockets availabe.")
 
         if 'HTTP_SEC_WEBSOCKET_KEY1' in environ:
             self.protocol_version = 76
@@ -192,7 +197,13 @@ class WebSocketWSGI(object):
         return WebSocket(sock, environ, self.protocol_version)
 
     def _handle_hybi_request(self, environ):
-        sock = environ['eventlet.input'].get_socket()
+        if 'eventlet.input' in environ.keys():
+            sock = environ['eventlet.input'].get_socket()
+        elif 'gunicorn.socket' in environ.keys():
+            sock = environ["gunicorn.socket"]
+        else:
+            raise("No knows sockets availabe.")
+
         hybi_version = environ['HTTP_SEC_WEBSOCKET_VERSION']
         if hybi_version not in ('8', '13', ):
             raise BadRequest(status='426 Upgrade Required',
--- a/eventlet/websocket.py
+++ b/eventlet/websocket.py
@@ -135,12 +135,12 @@ class WebSocketWSGI(object):
         return wsgi.ALREADY_HANDLED
 
     def _handle_legacy_request(self, environ):
-        if 'eventlet.input' in environ.keys():
+        if 'eventlet.input' in environ:
             sock = environ['eventlet.input'].get_socket()
-        elif 'gunicorn.socket' in environ.keys():
+        elif 'gunicorn.socket' in environ:
             sock = environ["gunicorn.socket"]
         else:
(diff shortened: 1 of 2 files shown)


eventlet/websocket.py
```diff
@@ -135,7 +135,12 @@ def __call__(self, environ, start_response):
         return wsgi.ALREADY_HANDLED
 
     def _handle_legacy_request(self, environ):
-        sock = environ['eventlet.input'].get_socket()
+        if 'eventlet.input' in environ.keys():
```
temoto: Please, do not use `in dict.keys()` or `dict.has_key()`.

```
d = {'key': 'value'}
assert ('key' in d) == ('key' in d.keys()) == True

%timeit 'key' in d.keys()
10000000 loops, best of 3: 145 ns per loop

%timeit 'key' in d
10000000 loops, best of 3: 57.8 ns per loop
```

eventlet/websocket.py
```diff
@@ -135,7 +135,12 @@ def __call__(self, environ, start_response):
...
+            sock = environ['eventlet.input'].get_socket()
+        elif 'gunicorn.socket' in environ.keys():
+            sock = environ["gunicorn.socket"]
+        else:
+            raise("No knows sockets availabe.")
```
temoto: - Use one kind of quotes (single).
- `raise` needs exception type. I don't have a better idea than `raise Exception('helpful text')`.
- Provide helpful message in exception. Something along `environ must contain either 'eventlet.input' object with .get_socket() method or 'guniron.socket'`.

temoto:
@catroot thank you for patch. Could you share a minimal test case?

catroot:
Of sure. Here it is:

`# -*- coding: utf-8 -

from eventlet import wsgi, websocket, sleep
from gunicorn import __version__


def app(environ, start_response):
    """Simplest possible application object"""
    @websocket.WebSocketWSGI
    def hello_world(ws):
        from random import choice
        while 1:
            ws.send(choice(("red","green","blue")))
            sleep(seconds=0.2)
        ws.close
    body=b"""<html>
  <head><title>pushdemo</title>
    <script src="http://ajax.googleapis.com/ajax/libs/jquery/1.4.1/jquery.min.js"></script>
    <style type="text/css">
      .bar {width: 20px; height: 20px;}
    </style>
    <script>
      window.onload = function() {
        ws = new WebSocket("ws://localhost:5001/");
        ws.onmessage = function(msg) {
          bar = $('#' + msg.data);
          bar.width(bar.width() + 10);
        }
      }
    </script>
  </head>
  <body>
    <div style="width: 400px;">
      <div id="red" class="bar"
          style="background-color: red;">&nbsp;</div>
      <div id="green" class="bar"
          style="background-color: green;">&nbsp;</div>
      <div id="blue" class="bar"
          style="background-color: blue;">&nbsp;</div>
    </div>
  </body>
</html>
"""
    if "HTTP_UPGRADE" in environ and environ['HTTP_UPGRADE']== 'websocket':
            hello_world(environ, start_response)
    status = '200 OK'
    response_headers = [
        ('Content-type', 'text/html'),
        ('Content-Length', str(len(body))),
        ('X-Gunicorn-Version', __version__),
    ]
    start_response(status, response_headers)
    return iter([body])
`

temoto:
6de8d47 this is squashed commit for merge. Changes:
- author name root -> catroot
- exception string did not fit our PEP-8 checker max line length (101)
- test code rewritten (see tests/manual/websocket-gunicorn)

I'm waiting few days for your approval before merging.

catroot:
There are some workaround needed, because of exception after closing websocket:
```
gunicorn/http/wsgi.py", line 238, in should_close
    if self.status_code < 200 or self.status_code in (204, 304):
AttributeError: 'Response' object has no attribute 'status_code'
```
self.status_code initializes from start_response, which is never called if we pass socket to ws handler.

temoto:
Yes, I've seen that exception, but didn't pay much attention to it because everything works.

Adding `start_response` in upgrade branch results in `[Errno 9] Bad file descriptor` in `send_headers`. If you can fix it - great. If not, I don't see merge stoppers.

temoto:
Forget about `greenio` :-)

catroot:
why to forget?
sending data over slosed socked happens there.. imo there must be a try to send and no need to raise any excettions if we just can`t.

temoto:
Because it's not related to Gunicorn.

Your patch is merged into master 6de8d47.

PR #334, opened 2016-08-02
wsgi: 400 on blank Content-Length headers

tipabu:
Previously, a client sending a blank Content-Length header would trigger a ValueError in Input.\_\_init\_\_, preventing a response from being sent.

Now, blank Content-Length headers will be treated like non-integer values, and a `400 Bad Request` response will be sent.


temoto:
Thanks, merged in master

PR #337, opened 2016-08-11
tests: Use upstream mock (2.0.0) and not embedded one

onovy:


temoto:
Does it fix any problem?

onovy:
none i'm aware.

but it's good practice not embed libs but use them as deps :). And this new deps is needed only during developing/testing.

temoto:
I believe a rule has to solve a problem to be a good practice.
Otherwise it's a religion, which I try to stay away from.
Praise critical thinking. :-)

PR #338, opened 2016-08-11
ssl: Disable SSL v2 and v3 protocols

onovy:
TLS v1 is widely supported and SSL v2 and v3 have security flaws.

### DIFF
--- a/doc/ssl.rst
+++ b/doc/ssl.rst
@@ -37,6 +37,7 @@ Here's an example of a server::
 
     # insecure context, only for example purposes
     context = SSL.Context(SSL.SSLv23_METHOD)
+    context.set_options(SSL.OP_NO_SSLv2 | SSL.OP_NO_SSLv3)
     context.set_verify(SSL.VERIFY_NONE, lambda *x: True))
 
     # create underlying green socket and wrap it in ssl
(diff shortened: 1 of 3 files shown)


temoto:
AFAIK, this patch disables TLSv1.1+.

onovy:
yep, sorry. Fixed

onovy:
To be honest, i'm not sure if we should change SSL options inside eventlet.

there is function:
https://docs.python.org/2/library/ssl.html#ssl.create_default_context

which sets reasonable secure defaults and users should use it.

What do you think?

temoto:
I hope it sets secure defaults. If you encounter a problem with those defaults, I'm very open to override SSL options.

General contribution guide:
- encounter a problem (I'm not sure this pull request passes)
- construct a minimal code to reproduce it, preferably in form of test case, but I'm happy to help with that
- construct a bug fix patch, again, I'm happy to help with that

*This is critical scenario*, it's not automated yet, I do it manually:
- ensure test fails against master
- apply patch
- ensure test passes

Strategy to change important pieces, such as security defaults:
- leave previous behavior as default
- add `EVENTLET_OPTION` environment variable handler to let users try it with comfort

PR #339, opened 2016-08-12
greenio: only trampoline when we block (py2)

tipabu:
This is the py2 equivalent of 863a1b7.

Note that `send` was already fixed in 6cbd484, but this makes the code consistent between all versions.


temoto:
Can you think of a way to test this?

itamarst:
Python 2 is no longer supported, so closing this.

PR #342, opened 2016-08-25
Tweak tests.run_python() to run with Python 2.7 subprocess.

nat-goodspeed:
Python 2.7 subprocess doesn't define subprocess.TimeoutExpired, and its
Popen.communicate() doesn't support a timeout parameter. The present
run_python() function in tests/__init__.py therefore produces massive spurious
failures when run with Python 2.7. Work around both so timeout is passed, and
TimeoutExpired caught, if supported; but if not things should still work.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -25,6 +25,35 @@ from nose.plugins.skip import SkipTest
 import eventlet
 from eventlet import tpool
 
+# Python 2.7 compatibility
+
+try:
+    TimeoutExpired = subprocess.TimeoutExpired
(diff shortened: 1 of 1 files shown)


temoto:
We already use subprocess32 compatibility layer for that:
```
# tests/__init__
try:
    import subprocess32 as subprocess  # py2
except ImportError:
    import subprocess  # py3

# tox.ini
deps =
    py{26,27}: subprocess32==3.2.7
```

Run tests in one of tox defined environments, or install `subprocess32` explicitly.

PR #343, opened 2016-08-27
[green.zmq] Support other convenient send/recv wrappers

ozamiatin:
The following wrappers added:

* send_string/recv_string
* send_json/recv_json
* send_pyobj/recv_pyobj

This change is also needed to make these methods thread-safe.


temoto:
Thanks, looks useful to raise popularity of eventlet/zeromq bond.

PR #346, opened 2016-09-15
Update ssl error messages to match builtin ssl

benkuhn:
Some libraries, e.g. `urllib3`, [look at the message string of an `SSLError`](https://github.com/shazow/urllib3/blob/master/urllib3/response.py#L307) to determine their behavior (in particular, whether to translate some exceptions into their own exception types). This causes a bad interaction with eventlet where running with eventlet causes an `SSLError` to be thrown where previously there was a `urllib3.ReadTimeoutError` (or a `requests` error), breaking exception handling.

While this behavior is a somewhat unfortunate choice on urllib3's part, it seems nice for eventlet not to break compatibility with the Python ssl module here. So this PR would change the error messages to match those set in [_ssl.c](https://github.com/python/cpython/blob/master/Modules/_ssl.c#L2168). It looks like the error messages are [consistent between 2.6 and master](https://github.com/python/cpython/blob/2.6/Modules/_ssl.c#L1333) so this should work on all versions.

EDIT NOTE: I don't really know what I'm doing with eventlet, so please let me know if there's a better way to accomplish this or if I changed any error messages incorrectly :)

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -116,12 +116,12 @@ class GreenSSLSocket(_original_sslsocket):
                         trampoline(self,
                                    read=True,
                                    timeout=self.gettimeout(),
-                                   timeout_exc=timeout_exc('timed out'))
+                                   timeout_exc=timeout_exc('The read operation timed out'))
                     elif get_errno(exc) == SSL_ERROR_WANT_WRITE:
                         trampoline(self,
                                    write=True,
                                    timeout=self.gettimeout(),
-                                   timeout_exc=timeout_exc('timed out'))
+                                   timeout_exc=timeout_exc('The write operation timed out'))
                     else:
                         raise
 
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -145,7 +145,7 @@ class GreenSSLSocket(_original_sslsocket):
             return self._call_trampolining(
                 super(GreenSSLSocket, self).send, data, flags)
         else:
-            trampoline(self, write=True, timeout_exc=timeout_exc('timed out'))
+            trampoline(self, write=True, timeout_exc=timeout_exc('The write operation timed out'))
             return socket.send(self, data, flags)
 
     def sendto(self, data, addr, flags=0):
@@ -154,7 +154,7 @@ class GreenSSLSocket(_original_sslsocket):
             raise ValueError("sendto not allowed on instances of %s" %
                              self.__class__)
         else:
-            trampoline(self, write=True, timeout_exc=timeout_exc('timed out'))
+            trampoline(self, write=True, timeout_exc=timeout_exc('The write operation timed out'))
             return socket.sendto(self, data, addr, flags)
 
     def sendall(self, data, flags=0):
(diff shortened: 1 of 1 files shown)


drpoggi:
@benkuhn I'd consider submitting the exact same changes you submitted to requests, to urllib3. Because you're just changing checking the `timed out` message in the string to a less restrictive check, it *shouldn't* break anything (I'm a novice on these libs though, so it could potentially break everything). You're also likely to get quicker movement on a PR to urllib3(@Lukasa isn't human)

temoto:
> While this behavior is a somewhat unfortunate choice on urllib3's part

I deeply agree.

Lukasa:
@temoto We are happy to stop looking at exception strings if we can have another way to tell the difference between timeouts and other SSL-related errors.

temoto:
@Lukasa how about `getattr(ex, 'is_timeout', False)`?

Lukasa:
Does that object generally exist on timeout exceptions?

temoto:
No, generally there's https://docs.python.org/2/library/socket.html#socket.timeout

```
exception socket.timeout
This exception is raised when a timeout occurs on a socket which has had timeouts enabled via a prior call to settimeout(). The accompanying value is a string whose value is currently always “timed out”.

New in version 2.3.
```

we will introduce a new property and hope others see benefits and adopt this practice.

Lukasa:
Yes, so annoyingly `socket.timeout` is sometimes not raised by the standard library during SSL operations, which is why we have done all this work with the exceptions.

temoto:
Please evaluate API in this branch: https://github.com/eventlet/eventlet/tree/is-timeout 6359fd3

```
try:
  sock.recv(1)
except Exception as e:
  if eventlet.is_timeout(e):
    # retry logic
  else:
    raise

CustomTimeout = eventlet.wrap_is_timeout(Exception)
```

Lukasa:
So this API is a bit tricky for us to use because it requires us to conditionally import eventlet, which is not a dependency of ours. Ideally we'd have a property on the exception itself to look for.

temoto:
You can just copy, it's a one line function:

```
def is_timeout(obj):
    return bool(getattr(obj, 'is_timeout', False))
```

temoto:
@Lukasa ping

BTW I found even better API was adopted in Python 3.something: root namespace TimeoutError exception.

Lukasa:
Yeah, it would generally be better to have a Python-3-only world where there is a more sensible exception hierarchy. However, we sadly still need a solution that works in Python 2 as well.

temoto:
I see how you can't catch custom exception without importing eventlet.

Anyway, pinged you to get feedback on this https://github.com/eventlet/eventlet/pull/346#issuecomment-270104540

I think it was misunderstanding, I'm not advocating to use function to test if something is timeout error, quite contrary, testing for is_timeout attribute is the public API, those functions are only shortcuts. Is it good to merge now? Can we use this new attribute instead of looking at exception message?

Lukasa:
I'd be happy for that new attribute.

temoto:
Cool, thanks. I have to fix tests before merging, though.

temoto:
@Lukasa it's merged in master. I think next release is possible closer to february, but no definite schedule.

Thanks for bringing this up.

Highly appreciate spreading the word about this API, I hope together we make overall Python2 timeouts handling a little more sane.

PR #347, opened 2016-09-19
Add DAGPool, a dependency-driven greenthread pool.

nat-goodspeed:
Add DAGPool tests.

### DIFF
--- a/eventlet/dagpool.py
+++ b/eventlet/dagpool.py
@@ -324,7 +324,7 @@ class DAGPool(object):
         # OWN key.
         coro = self.coros.get(key, _MISSING)
         if coro is not _MISSING \
-            and coro.greenthread is not greenthread.getcurrent():
+         and coro.greenthread is not greenthread.getcurrent():
             # oh oh, trying to post a value for running greenthread from
             # some other greenthread
             raise Collision(key)
@@ -335,7 +335,7 @@ class DAGPool(object):
         # Has somebody already post()ed a value for this key?
         # Unless replace == True, this is a problem.
         if self.values.get(key, _MISSING) is not _MISSING \
-            and not replace:
+         and not replace:
             raise Collision(key)
 
         # Either we've never before posted a value for this key, or we're
--- a/eventlet/dagpool.py
+++ b/eventlet/dagpool.py
@@ -30,6 +30,7 @@ $/LicenseInfo$
 
 from eventlet.event import Event
 from eventlet import greenthread
+import collections
 
 
 class Collision(Exception):
@@ -84,13 +85,9 @@ class DAGPool(object):
     GreenPool: it may be hard to provably avoid deadlock.
     """
 
-    class _Coro(object):
-        """
-        Internal object used to track running greenthreads
-        """
-        def __init__(self, greenthread, pending):
-            self.greenthread = greenthread
-            self.pending = pending
+
+    _Coro = collections.namedtuple("_Coro", ("greenthread", "pending"))
+
 
     def __init__(self, preload={}):
         """
--- a/eventlet/dagpool.py
+++ b/eventlet/dagpool.py
@@ -40,7 +40,7 @@ class PropagateError(Exception):
         self.exc = exc
 
     def __str__(self):
-        return "PropagateError({}): {}: {}" \
+        return "PropagateError({0}): {1}: {2}" \
                .format(self.key, self.exc.__class__.__name__, self.exc)
 
 
--- a/eventlet/dagpool.py
+++ b/eventlet/dagpool.py
@@ -357,11 +357,11 @@ class DAGPool(object):
         __getitem__(key) (aka dagpool[key]) blocks until 'key' has a value,
         then delivers that value.
         """
-        # This is a degenerate case of wait_each(). Construct a list
-        # containing only this 'key'. list(wait_each()) will make a list
-        # containing exactly one (key, value) pair. Extract that pair, then
-        # extract just its value.
-        return list(self.wait_each([key]))[0][1]
+        # This is a degenerate case of wait_each(). Construct a tuple
+        # containing only this 'key'. wait_each() will yield exactly one (key,
+        # value) pair. Return just its value.
+        for _, value in self.wait_each( (key,) ):
+            return value
 
     def get(self, key, default=None):
         """
--- a/eventlet/dagpool.py
+++ b/eventlet/dagpool.py
@@ -31,6 +31,7 @@ $/LicenseInfo$
 from eventlet.event import Event
 from eventlet import greenthread
 
+
 class Collision(Exception):
     """
     DAGPool raises Collision when you try to launch two greenthreads with the
@@ -39,6 +40,7 @@ class Collision(Exception):
     key in question.
     """
     pass
+
 
 class DAGPool(object):
     """
--- a/eventlet/dagpool.py
+++ b/eventlet/dagpool.py
@@ -1,0 +1,484 @@
+#!/usr/bin/python
+"""\
+@file   dagpool.py
+@author Nat Goodspeed
+@date   2016-08-08
+@brief  Provide DAGPool class
+
+$LicenseInfo:firstyear=2016&license=mit$
+Copyright (c) 2016, Linden Research, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+$/LicenseInfo$
+"""
+
+from eventlet.event import Event
+from eventlet import greenthread
+
+class Collision(Exception):
+    """
+    DAGPool raises Collision when you try to launch two greenthreads with the
+    same key, or post() a result for a key corresponding to a greenthread, or
+    post() twice for the same key. As with KeyError, str(collision) names the
+    key in question.
+    """
+    pass
+
+class DAGPool(object):
+    """
+    A DAGPool is a pool that constrains greenthreads, not by max concurrency,
+    but by data dependencies.
+
+    This is a way to implement general DAG dependencies. A simple dependency
+    tree (flowing in either direction) can straightforwardly be implemented
+    using recursion and (e.g.) GreenThread.imap(). What gets complicated is
+    when a given node depends on several other nodes as well as contributing
+    to several other nodes.
+
+    With DAGPool, you concurrently launch all applicable greenthreads; each
+    will proceed as soon as it has all required inputs. The DAG is implicit in
+    which items are required by each greenthread.
+
+    Each greenthread is launched in a DAGPool with a key: any value that can
+    serve as a Python dict key. The caller also specifies an iterable of other
+    keys on which this greenthread depends. This iterable may be empty.
+
+    The greenthread callable must accept (key, results), where:
+
+    key is its own key
+    results is an iterable of (key, value) pairs.
+
+    A newly-launched DAGPool greenthread is entered immediately, and can
+    perform any necessary setup work. At some point it will iterate over the
+    (key, value) pairs from the passed 'results' iterable. Doing so blocks the
+    greenthread until a value is available for each of the keys specified in
+    its initial dependencies iterable. These (key, value) pairs are delivered
+    in chronological order, NOT the order in which they are initially
+    specified: each value will be delivered as soon as it becomes available.
+
+    The value returned by a DAGPool greenthread becomes the value for its
+    key, which unblocks any other greenthreads waiting on that key.
+
+    External greenthreads may also interact with a DAGPool. See wait_each(),
+    waitall(), post().
+
+    It is not recommended to constrain DAGPool producer greenthreads in a
+    GreenPool: it may be hard to provably avoid deadlock.
+    """
+
+    class _Coro(object):
+        """
+        Internal object used to track running greenthreads
+        """
+        def __init__(self, greenthread, pending):
+            self.greenthread = greenthread
+            self.pending = pending
+
+    def __init__(self, preload={}):
+        """
+        DAGPool can be prepopulated with an initial dict or iterable of (key,
+        value) pairs. These (key, value) pairs are of course immediately
+        available for any greenthread that depends on any of those keys.
+        """
+        try:
+            # If a dict is passed, copy it. Don't risk a subsequent
+            # modification to passed dict affecting our internal state.
+            iteritems = preload.iteritems()
+        except AttributeError:
+            # Not a dict, just an iterable of (key, value) pairs
+            iteritems = preload
+
+        # Load the initial dict
+        self.values = dict(iteritems)
+
+        # track greenthreads
+        self.coros = {}
+
+        # The key to blocking greenthreads is the Event.
+        self.event = Event()
+
+    def wait_each(self, keys):
+        """
+        keys is an iterable of keys.
+
+        wait_each(keys) is a generator producing (key, value) pairs as a value
+        becomes available for each requested key. wait_each() blocks the
+        calling greenthread until values become available. If the DAGPool was
+        prepopulated with any values, of course those can be delivered
+        immediately without waiting.
+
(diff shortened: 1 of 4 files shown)


eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+        """
+        spawn_many() accepts a single function whose parameters are the same
+        as for spawn().
+
+        The difference is that spawn_many() accepts a dependency dict. A new
+        greenthread is spawned for each key in the dict. That dict key's value
+        should be an iterable of other keys on which this greenthread depends.
+        """
+        # Iterate over a snapshot of 'depends' items
+        for key, deps in depends.items():
```
temoto: In Python3 `dict.items()` is not a snapshot, instead it's a view that sees changes in dict.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+        """
+        while True:
+            # Before even waiting, show caller any (key, value) pairs that
+            # are already available. Copy 'pending' because we want to be able
+            # to remove items from the original set while iterating.
+            for key in pending.copy():
+                try:
+                    value = self.values[key]
+                except KeyError:
+                    pass
```
temoto: It's your call, but I really think that easier to read (also faster but it's a joke in Python)

```
# somewhere in top
_missing = object()


value = self.values.get(key, _missing)
if value is not _missing:
  ...
```

Same applies to all following blocks of code that catch KeyError.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+        """
+        return self.values.get(key, default)
+
+    def keys(self):
+        """
+        Return the list of keys for which we currently have values. Explicitly
+        return a copy rather than an iterator: don't assume our caller will
+        finish iterating before new values are posted.
+        """
+        return self.values.keys()
```
temoto: Again, Python2-3 different semantics for `dict.keys`, `dict.items`, `dict.values`.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+    def __getitem__(self, key):
+        """
+        __getitem__(key) (aka dagpool[key]) blocks until 'key' has a value,
+        then delivers that value.
+        """
+        # This is a degenerate case of wait_each(). Construct a list
+        # containing only this 'key'. list(wait_each()) will make a list
+        # containing exactly one (key, value) pair. Extract that pair, then
+        # extract just its value.
+        return list(self.wait_each([key]))[0][1]
```
temoto: If you don't modify the list, then `tuple` seems a better fit.
And then you only use first item, so no need to construct any sequence storage, just fetch first pair from iterator using `next()` or `for-break`.

```
for _, value in self.wait_each( (key,) ):
  return value
```

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+        """
+        return self.values.keys()
+
+    def items(self):
+        """
+        Return a snapshot list of currently-available (key, value) pairs.
+        Don't assume our caller will finish iterating before new values are
+        posted.
+        """
+        return self.values.items()
```
temoto: Again, Python2-3 different semantics for `dict.keys`, `dict.items`, `dict.values`.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def running_keys(self):
+        """
+        Return keys for running greenthreads. This includes greenthreads
+        blocked while iterating through their 'results' iterable, that is,
+        greenthreads waiting on values from other keys.
+        """
+        # return snapshot; don't assume caller will finish iterating before we
+        # next modify self.coros
+        return self.coros.keys()
```
temoto: Python2-3 different semantics for `dict.keys`, `dict.items`, `dict.values`.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+        waiting on values from other keys. This explicitly does NOT include
+        external greenthreads waiting on wait(), waitall(), wait_each().
+        """
+        # n.b. if Event would provide a count of its waiters, we could say
+        # something about external greenthreads as well.
+        # The logic to determine this count is exactly the same as the general
+        # waiting_for() call.
+        return len(self.waiting_for())
+
+    class _Omitted(object):
```
temoto: `_omitted = object()` would do the same without class entity.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+        This method allows diagnosing a 'hung' DAGPool. If certain
+        greenthreads are making no progress, it's possible that they are
+        waiting on keys for which there is no greenthread and no post() data.
+        """
+        # We may have greenthreads whose 'pending' entry indicates they're
+        # waiting on some keys even though values have now been posted for
+        # some or all of those keys, because those greenthreads have not yet
+        # regained control since values were posted. So make a point of
+        # excluding values that are now available.
+        available = set(self.values.iterkeys())
```
temoto: `six.iterkeys`

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+
+        # Start by iterating over (key, coro) pairs in self.coros. Generate
+        # (key, pending) pairs in which 'pending' is the set of keys on which
+        # the greenthread believes it's waiting, minus the set of keys that
+        # are now available. Filter out any pair in which 'pending' is empty,
+        # that is, that greenthread will be unblocked next time it resumes.
+        # Make a dict from those pairs.
+        return dict((key, pending)
+                    for key, pending in ((key, (coro.pending - available))
+                                         for key, coro in self.coros.iteritems())
```
temoto: `six.iteritems`

temoto:
See Travis https://travis-ci.org/eventlet/eventlet/builds/160940365
also failed PEP-8 and Python2.6 doesn't support multiple assignments in `with`.

nat-goodspeed:
I saw the Travis failures a few days ago and have just committed changes that I hope will address the previous round.

Thank you for your personal review! I'm abashed to admit that I've gotten used to relying on the 2to3 converter promise that it will "do the right thing" with (e.g.) dict.iteritems(). I'm not used to coding in a style that can be executed unmodified by either version of the language. Will address.

eventlet/dagpool.py
```diff
@@ -0,0 +1,484 @@
...
+    def __init__(self, preload={}):
+        """
+        DAGPool can be prepopulated with an initial dict or iterable of (key,
+        value) pairs. These (key, value) pairs are of course immediately
+        available for any greenthread that depends on any of those keys.
+        """
+        try:
+            # If a dict is passed, copy it. Don't risk a subsequent
+            # modification to passed dict affecting our internal state.
+            iteritems = preload.iteritems()
```
nat-goodspeed: I will fix the Python 2/3 compatibility issue. But the use of iteritems() here isn't a mistake: the copy happens a few lines below, when iteritems is consumed by the dict() constructor.

eventlet/dagpool.py
nat-goodspeed: Hmm! The Python Cookbook claimed that when you really do need alternative control flow, catching KeyError is more idiomatic. Admittedly that was a few revisions ago, of the book and the language. If the get(key, _missing) construct is more consistent with the rest of the library, I'll use that.

eventlet/dagpool.py
temoto: `six.items` returns view on both Python2/3. For this `dict()` copy case, it's like `iteritems` but even better.

temoto:
Yeah, multiversion compatibility is hard.

eventlet/dagpool.py
temoto: It is consistently used in Eventlet, although I'm ashamed to admit it has random names. Maybe upper case `_MISSING` for "kind of constant" would be good.

nat-goodspeed:
I'm curious, though: do you support the DAGPool design? If you hate this functionality, I wouldn't bother fixing it for inclusion.

temoto:
Honestly, I didn't understand the use case. But definitely no negative feelings, lest hate. I'm glad to merge it. Eventually some documentation would help other people to use it.

So far I gathered DAG pool consists of two distinct features:
- "prefork" limited number of long lived worker greenthreads. This is very new, I can't imagine how one could come to this design, given that greenlets are extremely lightweight. Maybe this part was copied from process/thread implementation.
- by the book map-reduce pipeline. This is so awesome, it's a shame we didn't have it yet.

nat-goodspeed:
Documentation: I wrote fairly verbose docstrings for everything. Please direct me how to improve?

I wasn't thinking of this as map-reduce. It's just that I have some processing (rebuilding lots of libraries) in which I want to launch each rebuild as soon as I have in hand all the other libraries on which it depends. It occurred to me that I could get pleasing concurrency in eventlet with a dict and an Event. From that I extracted this DAGPool design, which I hope is general enough to serve a wide variety of needs.

n.b. Delivering each result from a group of greenthreads as soon as it becomes available is a degenerate but imo valuable use of DAGPool.  :-)

temoto:
On documentation, I mean this http://eventlet.net/doc/
it's partly generated from docstrings, but I think that part more for checking subtleties of API, the other part is hand written text (hopefully) helping people who are not sure what to do.

Is your library building thing open source?

nat-goodspeed:
The library building script isn't yet written! When I realized I will want DAGPool functionality to overlap our builds as efficiently as I can manage, I detoured to design the DAGPool API and implementation.

The script I have in mind is intended to work with [autobuild](https://bitbucket.org/lindenlab/autobuild-1.0) packages and our internal [TeamCity](https://www.jetbrains.com/teamcity/) instance. `autobuild` implements cross-package dependencies by baking the upstream package's built URL into the consuming package's `autobuild.xml` file. Therefore, to propagate a change to a given package through the whole dependency graph, you must:

1. run that first package build
1. obtain the URL of the `autobuild` package tarball
1. fetch or update the repository of each immediate consumer library
1. edit the new URL and hash into each consumer's `autobuild.xml`
1. commit (!) and rebuild each consumer, producing a new package URL for each
1. lather, rinse, repeat.

If you tell me this will be generally useful to the whole wide world, I will try to abstract away specific details such as accessing https://bitbucket.org and our TeamCity instance and the `autobuild` command to update the consuming repository.

But initially it looks pretty specific to our needs.

temoto:
About generally useful, in short, try and see. I'm not competent in complex dependency builds.

Last build failures:
- Python2.6 `with assert_raises(Collision): TypeError: failUnlessRaises() takes at least 3 arguments (2 given)`
- Python3.* `dict has not attribute iterkeys`

nat-goodspeed:
Yes, those were the same as the previous round of failures. My last push wasn't trying to address those; it added a dagpool.rst analogous to some of the other .rst files, to try to incorporate DAGPool into the generated documentation.

eventlet/dagpool.py
nat-goodspeed: I may be missing something, but I find no items attribute in eventlet.support.six. I will use six.iteritems() instead.

nat-goodspeed:
I am frankly confused by the remaining pep8 complaints I'm seeing. In two places I have an `if ... and` test that I want broken onto two lines for clarity.

[My first attempt](https://github.com/eventlet/eventlet/pull/347/commits/16443ca4e115be68387744a14722990fa40e2782#diff-a7713f1429e7f4dbcb8d68667ce7e9fcR328) aligned the `and` with the `if`, which [got scolded](https://travis-ci.org/eventlet/eventlet/jobs/162162077):

    eventlet/dagpool.py:329:9: E122 continuation line missing indentation or outdented
        and coro.greenthread is not greenthread.getcurrent():
        ^
    eventlet/dagpool.py:340:9: E122 continuation line missing indentation or outdented
        and not replace:
        ^

I [changed both places](https://github.com/eventlet/eventlet/pull/347/commits/19ea31961da0c75943e44b3f22415efb915dcbf3#diff-a7713f1429e7f4dbcb8d68667ce7e9fcL329) to indent the `and`, which got [a different scolding](https://travis-ci.org/eventlet/eventlet/jobs/162169869):

    eventlet/dagpool.py:327:13: E125 continuation line with same indent as next logical line
            and coro.greenthread is not greenthread.getcurrent():
            ^
    eventlet/dagpool.py:338:13: E125 continuation line with same indent as next logical line
            and not replace:
            ^

So I [reindented](https://github.com/eventlet/eventlet/pull/347/commits/505b19435e9c75de590b9e33045c3b1ada9971ed#diff-a7713f1429e7f4dbcb8d68667ce7e9fcL327) to align with the `if` plus one space, which produced [yet another complaint](https://travis-ci.org/eventlet/eventlet/jobs/162170254):

    eventlet/dagpool.py:327:10: E127 continuation line over-indented for visual indent
         and coro.greenthread is not greenthread.getcurrent():
         ^
    eventlet/dagpool.py:338:10: E127 continuation line over-indented for visual indent
         and not replace:
         ^

Before I conclude that your pep8 filter will not let me use line continuation with an `if` statement at all, do you have any hints? Or can you mark those "failures" as acceptable?

nat-goodspeed:
I would also like to point out that [the](https://travis-ci.org/eventlet/eventlet/jobs/162170262) [other](https://travis-ci.org/eventlet/eventlet/jobs/162170265) [failures](https://travis-ci.org/eventlet/eventlet/jobs/162170266) in the most recent Travis CI run have nothing to do with this PR:

    ERROR: test_suspend_doesnt_crash (tests.hub_test.TestSuspend)
      File "/home/travis/build/eventlet/eventlet/tests/hub_test.py", line 277, in test_suspend_doesnt_crash
        eventlet.sleep(0.4)  # wait for process to hit accept
      File "/home/travis/build/eventlet/eventlet/eventlet/greenthread.py", line 34, in sleep
        hub.switch()
      File "/home/travis/build/eventlet/eventlet/eventlet/hubs/hub.py", line 294, in switch
        return self.greenlet.switch()
    TestIsTakingTooLong: 4

    ERROR: tests.tpool_test.test_isolate_from_socket_default_timeout
      File "/home/travis/build/eventlet/eventlet/.tox/py33-epolls/lib/python3.3/site-packages/nose/case.py", line 198, in runTest
        self.test(*self.arg)
      File "/home/travis/build/eventlet/eventlet/tests/tpool_test.py", line 366, in test_isolate_from_socket_default_timeout
        tests.run_isolated('tpool_isolate_socket_default_timeout.py', timeout=1)
      File "/home/travis/build/eventlet/eventlet/tests/__init__.py", line 339, in run_isolated
        if output.startswith(b'skip'):
    TypeError: startswith first arg must be str or a tuple of str, not bytes

    ERROR: test_pipe (tests.greenio_test.TestGreenPipe)
      File "/home/travis/build/eventlet/eventlet/tests/greenio_test.py", line 673, in test_pipe
        line = rf.readline()
      File "/home/travis/build/eventlet/eventlet/eventlet/greenio/py3.py", line 105, in readinto
        data = self.read(up_to)
      File "/home/travis/build/eventlet/eventlet/eventlet/greenio/py3.py", line 88, in read
        self._trampoline(self, read=True)
      File "/home/travis/build/eventlet/eventlet/eventlet/greenio/py3.py", line 123, in _trampoline
        mark_as_closed=self._mark_as_closed)
      File "/home/travis/build/eventlet/eventlet/eventlet/hubs/__init__.py", line 162, in trampoline
        return hub.switch()
      File "/home/travis/build/eventlet/eventlet/eventlet/hubs/hub.py", line 294, in switch
        return self.greenlet.switch()
    tests.TestIsTakingTooLong: 1

temoto:
On PEP-8, I have two suggestions:
- use autopep8, make robots do the dirty work. Make sure it reads our extended max-line-length setting from configs or pass it explicitly.
- think shell style `\ newline` is evil, always use explicit techniques for multiline expressions. For long strings: triple quote or break into shorter strings and explicit concatenate with `+`. For long expressions: surround with parens and break lines for most readability (autopep8 could care less [1] about humans).

```
          if ((something_long or not_so_long or extremely_long('expression'))
                and this("should be a little more right than next block")
                and repeat_same_indent):
              then_do_the_right_thing()

# or there's always a way to make it short ;-)

          if key in self.values and not replace:
```

[1] https://xkcd.com/1576/

temoto:
Those other failures are indeed irrelevant, do not worry, be happy.

nat-goodspeed:
Please re-review; I think I have addressed all your review comments.

I also claim that the only two remaining Travis CI failures are unrelated.

Is there any further action I must take?

temoto:
Shebang `#!/usr/bin/python` is wrong (should be /usr/bin/env python to support virtualenvs) and should be removed because it's not used.

Can you remove current module docstrings and optionally add useful information there?
I understand you may wish to put your name and date in there (although we have separate AUTHORS file).
But license info really doesn't add any value because it duplicates LICENSE file, which also attributes to Linden.

nat-goodspeed:
I will do that. I am embarrassed to admit I am belatedly adding code to propagate an uncaught exception from a DAGPool greenthread. I acknowledge that the ball is back in my court.

temoto:
> "greenlet %s did not yet produce a value" % k

I can only insist on `%` with tuples, but suggest using `str.format(), it's` is better in every way.

temoto:
> AttributeError: 'generator' object has no attribute 'next'

`next(gen)` works in both Python2/3.

nat-goodspeed:
I will use str.format().

I think you mean six.next(iterator).

nat-goodspeed:
Still working on exception propagation. Thanks for your patience with all this!

temoto:
No I really meant just `next`.
https://docs.python.org/2/library/functions.html#next
https://docs.python.org/3/library/functions.html#next

temoto:
Some issues and pull requests wait for months, relax, there's no hurry.

temoto:
> +        return "PropagateError({}): {}: {}" \
 +               .format(self.key, self.exc.__class__.__name__, self.exc)

We still support Python2.6 which didn't yet allow unnamed references in str.format. Should be either `('{0} {1} {2}'.format(a, b, c)` or `'{key} {klass} {ex}'.format(key=..., klass=..., ex=...)`

nat-goodspeed:
At this point I feel pretty happy with DAGPool and its tests.

temoto:
Thanks.

nat-goodspeed:
Thank _you!_ Again, I appreciate your patience in working through this with me.

PR #348, opened 2016-09-27
Add Rationale and Tutorial example for DAGPool class.

nat-goodspeed:
Also mark up docstrings for Sphinx cross-references and such.

### DIFF
--- a/doc/modules/dagpool.rst
+++ b/doc/modules/dagpool.rst
@@ -1,6 +1,493 @@
 :mod:`dagpool` -- Dependency-Driven Greenthreads
 ================================================
 
+Rationale
+*********
+
+The dagpool module provides the :class:`DAGPool <eventlet.dagpool.DAGPool>`
+class, which addresses situations in which the value produced by one
+greenthread might be consumed by several others -- while at the same time a
+consuming greenthread might depend on the output from several different
+greenthreads.
+
+If you have a tree with strict many-to-one dependencies -- each producer
+greenthread provides results to exactly one consumer, though a given consumer
+may depend on multiple producers -- that could be addressed by recursively
+constructing a :class:`GreenPool <eventlet.greenpool.GreenPool>` of producers
+for each consumer, then :meth:`waiting <eventlet.greenpool.GreenPool.waitall>`
+for all producers.
+
+If you have a tree with strict one-to-many dependencies -- each consumer
+greenthread depends on exactly one producer, though a given producer may
+provide results to multiple consumers -- that could be addressed by causing
+each producer to finish by launching a :class:`GreenPool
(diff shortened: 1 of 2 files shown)


doc/modules/dagpool.rst
```diff
@@ -1,6 +1,493 @@
...
+is stuck!
+
+You could change waitall() to :meth:`wait_each()
+<eventlet.dagpool.DAGPool.wait_each>`, and print each key as it becomes
+available:
+
+::
+
+    for key, result in pool.wait_each():
+        print "got result for {0}".format(key)
```
temoto: `print` is a function.

doc/modules/dagpool.rst
```diff
@@ -1,6 +1,493 @@
...
+    c failed with PropagateError
+    b failed with PropagateError
+    e failed with PropagateError
+    d failed with PropagateError
+    zlib failed with OriginalError
+
+wait_each_exception() yields each PropagateError wrapper as if it were the
+result, rather than raising it as an exception.
+
+Notice that we print err.exc.\_\_class\_\_.\_\_name\_\_ because
```
temoto: Inline code backquotes would save you from escaping hell.

harlowja:
Interesting idea to have this in eventlet itself vs external.

I've created something similiar before @ http://docs.openstack.org/developer/taskflow/engines.html#compiling (that uses https://networkx.github.io/ for the graph structure) so I think its useful to have this, just not so sure it needs to be in eventlet itself to achieve the goal here (but to each there own).

nat-goodspeed:
On Wed, Sep 28, 2016 at 11:32 AM, Joshua Harlow <notifications@github.com>
wrote:

Interesting idea to have this in eventlet itself vs external.
>
> I've created something similiar before @ http://docs.openstack.org/
> developer/taskflow/engines.html#compiling (that uses
> https://networkx.github.io/ for the graph structure) so I think its
> useful to have this, just not so sure it needs to be in eventlet itself to
> achieve the goal here (but to each there own).
>

I thought of submitting it because the problem is to manage concurrency
across a dependency graph. You want to run each piece as soon as possible,
but no sooner. That feels fairly broadly applicable.

harlowja:
Right, I get the applicability of it, though to me it feels like something that doesn't need to tie itself to eventlet to achieve. Just my 2 cents, still useful IMHO and I guess we'll see where people use it :)

temoto:
Graph is a great datastructure and nice to have.
I think the problem is tight coupling between graph and pool. If I were to solve something similar, I'd implement minimalistic graph and run it over existing GreenPool. Greenthreads are dirt cheap.

Of course, all these comments have nothing to do with the documentation patch. But maybe it's good to have some place for discussion.

PR #352, opened 2016-10-13
Server socket SO_REUSEPORT option

gaozhengwei:
From linux kernel >= 3.9, it supports the SO_REUSEPORT option that will
allow multiple server socket to listen the same port


temoto:
Thanks, merged in master f9a3074a3b75f17f76cc04a693dc48a367b99861

PR #353, opened 2016-10-27
_recv_loop behaviour with recv_into on closed sock

davidszotten:
make sure `_recv_loop` returns the correct value when `recv_meth` is of
foo_into type (fills a buffer and returns number of bytes, not the data)


davidszotten:
failing test looks unrelated but i'm not allowed to re-trigger the build :(

temoto:
Thank you.

PR #354, opened 2016-11-02
greendns: Return answers from /etc/hosts despite nameserver errors

tipabu:
When calling `getaddrinfo` with an unspecified address family, we:

* try to get an IPv6 address for the host, first trying /etc/hosts and failing that, using any external nameserver, then
* try to get an IPv4 address for it, with the same /etc/hosts -> external nameserver priority.

Note that if the host can be resolved using /etc/hosts but only to one of the two address families, we still attempt to contact the external nameserver to resolve it in the other family.

Previously, if this timed out or raised a `DNSException` other than `NXDOMAIN`, `EAI_EAGAIN_ERROR` or `EAI_NODATA_ERROR` would be raised even though we could resolve the host. Now, those errors will only be raised if neither attempt at resolution succeeded.

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -402,10 +402,19 @@ def _getaddrinfo_lookup(host, family, flags):
         raise EAI_NONAME_ERROR
     addrs = []
     if family == socket.AF_UNSPEC:
+        err = None
         for qfamily in [socket.AF_INET6, socket.AF_INET]:
-            answer = resolve(host, qfamily, False)
-            if answer.rrset:
-                addrs.extend([rr.address for rr in answer.rrset])
+            try:
+                answer = resolve(host, qfamily, False)
+            except socket.gaierror as e:
+                if e.errno not in (socket.EAI_AGAIN, socket.EAI_NODATA):
+                    raise
+                err = e
+            else:
+                if answer.rrset:
+                    addrs.extend(rr.address for rr in answer.rrset)
+        if err is not None and not addrs:
+            raise err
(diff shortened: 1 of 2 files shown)


eventlet/support/greendns.py
```diff
@@ -402,10 +402,24 @@ def _getaddrinfo_lookup(host, family, flags):
...
+            except socket.gaierror as e:
+                if e.errno == socket.EAI_AGAIN:
+                    EAGAIN_caught = True
+                elif e.errno == socket.EAI_NODATA:
+                    NODATA_caught = True
+                else:
+                    raise
+            else:
+                if answer.rrset:
+                    addrs.extend([rr.address for rr in answer.rrset])
```
temoto: `list.extend` takes iterable. If there were no good reason for materialising list argument, then `extend(rr.a for rr in rrset)` is equally good.

eventlet/support/greendns.py
```diff
@@ -402,10 +402,24 @@ def _getaddrinfo_lookup(host, family, flags):
...
+        if EAGAIN_caught and not addrs:
+            raise EAI_EAGAIN_ERROR
```
temoto: Saving original exception seems better than raising a new one. Like

```
err = None
try:
  ...
except socket.gaierror as e:
  if e.errno in (again, nodata):
    err = e
...
if err is not None and not addrs:
  raise err
```

tipabu:
Good call on the `extend` call; I was just moving code, but that makes it much better.

I also fixed up the error trapping. Seems much cleaner, good recommendation.

temoto:
Thank you.

PR #356, opened 2016-11-15
Fix enumeration of refs

wbyoung:
Introduced in #309, this enumeration should be over the items of the dict in order to work correctly.

Of note: I'm running this via Gunicorn 19.6.0 and PyPy 5.5.0-alpha0 & an exception occurs nearly immediately (it seems that PyPy takes a lock on something before executing any Python).

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -372,9 +372,14 @@ def _fix_py3_rlock(old):
         new.acquire()
     gc.collect()
     for ref in gc.get_referrers(old):
-        for k, v in vars(ref):
-            if v == old:
-                setattr(ref, k, new)
+        try:
+            ref_vars = vars(ref)
+        except TypeError:
+            pass
+        else:
+            for k, v in ref_vars.items():
+                if v == old:
+                    setattr(ref, k, new)
 
 
 def _green_os_modules():


temoto:
Please describe in more detail what does it fix.

wbyoung:
@temoto one enumerates the pairs of a dict in Python by using `items`. Not doing so causes an exception since only one item is available to unpack. If memory serves me, I believe that this was the case in Python 2 as well, but I know it to be true for Python 3.

Also, `vars` cannot be called on objects that lack a `__dict__`, so this accounts for that as well.

temoto:
Thanks. I agree that `vars()` result doesn't contain tuples for unpacking. But I thought that we have that covered by tests and obviously it's not covered enough. Do you have an idea how to reproduce this error?

wbyoung:
@temoto running Gunicorn w/ eventlet under pypy3 will do it.

But it's possible that even importing eventlet w/ pypy3 will trigger the exception.

I don't think this code would be covered by tests. It'd be guaranteed to fail. `gc.get_referrers` must be returning nothing for the test case.

temoto:
Okay, the test is only triggering error on pypy3.

temoto:
Thank you very much.

Note for readers: current tests in patcher_existing_locks_* do not cover this change, my best guess is - because `test_lock` is not tracked by garbage collector. I've tried additional search via `globals()` without success.

wbyoung:
Thanks for incorporating it so quickly!

PR #359, opened 2016-11-29
Make eventlet.dagpool.PropagateError a good Exception subclass.

nat-goodspeed:
Previous `PropagateError.__init__()` failed to call `Exception.__init__()` at all.
Moreover, as described in https://bugs.python.org/issue1692335, setting
self.args is important for an `Exception` subclass with nonstandard constructor
arguments.


temoto:
ping

nat-goodspeed:
Yes, sorry, I'm very embarrassed, but I'm being snowed under with lots of year-end stuff. I do believe the change should be made. I will fix the tests.

Issue #374, opened 2017-01-10
python3.6 socket_test.test_dns_methods_are_green fails on poll/epolls hub


temoto:
Python3.6 continues to bring new exciting test failures, I guess it's good because they clearly are changing a lot of internals untouched for years.

This blocks v0.21 release.

```
tests.socket_test.test_dns_methods_are_green ... Program /Users/temoto/dev/github.com/eventlet/eventlet/tests/isolated/socket_resolve_green.py output:
---
Traceback (most recent call last):
  File "/Users/temoto/dev/github.com/eventlet/eventlet/tests/isolated/socket_resolve_green.py", line 41, in <module>
    assert 0.1 <= td < 0.3, 'Resolve time expected: ~0.1s, real: {0:.3f}'.format(td)
AssertionError: Resolve time expected: ~0.1s, real: 0.497

---
FAIL

======================================================================
FAIL: tests.socket_test.test_dns_methods_are_green
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/temoto/dev/github.com/eventlet/eventlet/venv-36/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/temoto/dev/github.com/eventlet/eventlet/tests/socket_test.py", line 75, in test_dns_methods_are_green
    tests.run_isolated('socket_resolve_green.py', pythonpath_extend=[mock_sys_pkg_dir])
  File "/Users/temoto/dev/github.com/eventlet/eventlet/tests/__init__.py", line 359, in run_isolated
    run_python(prefix + path, **kwargs)
  File "/Users/temoto/dev/github.com/eventlet/eventlet/tests/__init__.py", line 352, in run_python
    assert ok, 'Expected single line "pass" in stdout'
AssertionError: Expected single line "pass" in stdout

----------------------------------------------------------------------
```

temoto:
Okay this sporadic test failure is fixed in master 0fed9e52ed84f690cd20495375fc06203889f189

But it revealed a problem with GreenPool slow first `.spawn()` that is not fixed as of 2017-01-18: #379

PR #375, opened 2017-01-12
Fixed tests.greendns_test.TestGetaddrinfo

Kobla:
Asserting udp for ssh introduce test failures under debian unstable
because netbase 5.4 removed ssh (22/udp).
This fix changes ssh -> domain, which better suit for test name and
allows tcp and udp simultaneously.

### DIFF
--- a/tests/greendns_test.py
+++ b/tests/greendns_test.py
@@ -458,11 +458,11 @@ class TestGetaddrinfo(tests.LimitedTestCase):
         greendns.resolve = _make_mock_resolve()
         greendns.resolve.add('example.com', '127.0.0.2')
         greendns.resolve.add('example.com', '::1')
-        res = greendns.getaddrinfo('example.com', 'ssh')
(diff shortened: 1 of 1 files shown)


onovy:
CI fails is not related to this change, see https://github.com/eventlet/eventlet/issues/374

temoto:
Just out of curiosity, are you saying new netbase package leaves Debian with `/etc/protocols` without ssh-22 line?

temoto:
Thanks a lot.

Kobla:
I dont know whether the /etc/protocols will contain `ssh-22 line`.
I think `ssh 22/udp` will be removed from /etc/services. (`ssh 22/tcp` persists)

temoto:
Yeah, I meant services. Ah, now I get it. It's tcp/udp difference. That's clearly mistake on Eventlet side, thanks for fixing.

PR #377, opened 2017-01-16
Full IPv6 for machines with no IPv4 127.0.0.1

grantma:
Detects if ::1 is present, uses it by preference.  If getaddrinfo()
on hostname returns IPv6 address, turn on full IPv6 defaults.

Submitting for comment about approach.  Test coverage needs writing

### DIFF
--- a/eventlet/backdoor.py
+++ b/eventlet/backdoor.py
@@ -10,6 +10,7 @@ import traceback
 import eventlet
 from eventlet import hubs
 from eventlet.support import greenlets, get_errno
+from eventlet.support.afinet import ip_defaults
 
 try:
     sys.ps1
@@ -133,4 +134,4 @@ def backdoor(conn_info, locals=None):
 
 
 if __name__ == '__main__':
-    backdoor_server(eventlet.listen(('127.0.0.1', 9000)), {})
+    backdoor_server(eventlet.listen((ip_defaults['core_lo_addr'], 9000)), {})
--- a/eventlet/support/afinet.py
+++ b/eventlet/support/afinet.py
@@ -1,0 +1,46 @@
+"""
+Test availabilty of AF_INET6 on a machine
+
+Uses socket API to determine configuration
+"""
+
+from socket import *
+
+def check_ipv6_lo_addr():
+    try:
+        sock = socket(AF_INET6, SOCK_STREAM)
+        sock.bind(('::1', 0))
+        return True
+    except:
+        return False
+
+def check_ipv6_hosts_file():
+    if not check_ipv6_lo_addr():
+        return False
+    addrinfo = getaddrinfo(gethostname(), 0)
+    if (addrinfo[0][0] == AF_INET6):
+        return True
+    return False
+
+core_use_ipv6_lo_addr = check_ipv6_lo_addr()
+use_ipv6_by_default = check_ipv6_hosts_file()
+
+ip_defaults = {}
+
+if core_use_ipv6_lo_addr:
+    ip_defaults['core_lo_addr'] = '::1'
+    ip_defaults['core_af_inet'] = AF_INET6
+else:
+    ip_defaults['core_lo_addr'] = '127.0.0.1'
+    ip_defaults['core_af_inet'] = AF_INET
+
+if use_ipv6_by_default:
+    ip_defaults['af_inet'] = AF_INET6
+    ip_defaults['null_addr'] = '::'
+    ip_defaults['null_cidr'] = '::/0'
+    ip_defaults['lo_addr'] = '::1'
+else:
+    ip_defaults['af_inet'] = AF_INET
+    ip_defaults['null_addr'] = '0.0.0.0'
+    ip_defaults['null_cidr'] = '0.0.0.0/0'
+    ip_defaults['lo_addr'] = '127.0.0.1'
--- a/eventlet/convenience.py
+++ b/eventlet/convenience.py
@@ -5,9 +5,10 @@ from eventlet import greenpool
 from eventlet import greenthread
 from eventlet.green import socket
 from eventlet.support import greenlets as greenlet
+from eventlet.support.afinet import ip_defaults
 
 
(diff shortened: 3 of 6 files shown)


temoto:
Uhm... general approach is fine. Implementation needs few changes, I've marked them in review, not very important right now.

What's important is what problem does it solve? How do you get machine with Python and without IPv4?

eventlet/backdoor.py
```diff
@@ -10,6 +10,7 @@
 import eventlet
 from eventlet import hubs
 from eventlet.support import greenlets, get_errno
+from eventlet.support.afinet import ip_defaults
```
temoto: Only module imports are fine and safe. `from module import name` creates multiple references to same object which are impossible to patch and harder to understand while reading code. I know this may sound hypocritical with previous line importing `get_errno`, please ignore that's legacy code.

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
+"""
+Test availabilty of AF_INET6 on a machine
+
+Uses socket API to determine configuration
+"""
+
+from socket import *
```
temoto: `import socket`.

Again, only module imports are fine, `import *` is double error.

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+
+def check_ipv6_lo_addr():
+    try:
+        sock = socket(AF_INET6, SOCK_STREAM)
+        sock.bind(('::1', 0))
+        return True
+    except:
```
temoto: `except:` is absolutely prohibited. We do not hide unexpected errors. Consider there was `socket.bindwrwongname(...)`, this code would hide error. On Python2 this is double error, because `except:` also catches `GeneratorExit`, `SystemExit`, `KeyboardInterrupt `.

https://docs.python.org/2/library/exceptions.html

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+        return False
+
+def check_ipv6_hosts_file():
+    if not check_ipv6_lo_addr():
```
temoto: One `bind()` to just `import eventlet` is already too much. Executing this function twice is excessive.

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+        return False
+    addrinfo = getaddrinfo(gethostname(), 0)
+    if (addrinfo[0][0] == AF_INET6):
```
temoto: - `if (a) -> if a`, Python doesn't need parens.
- `if a: return True else: return False -> return bool(a)`. In this case it's even simpler `return ...==AFINET6`.
- It needs bounds check. There may be no elements in list.
- This check only considers first address. `not all(...)` construct is consistent with promise of "for machines with no IPv4".

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+        return True
+    return False
+
+core_use_ipv6_lo_addr = check_ipv6_lo_addr()
+use_ipv6_by_default = check_ipv6_hosts_file()
+
+ip_defaults = {}
```
temoto: Absolutely no reason to use mutable, string indexed dictionary here. Go with namedtuple or object with cached lazy property.

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+    ip_defaults['core_lo_addr'] = '::1'
+    ip_defaults['core_af_inet'] = AF_INET6
+else:
+    ip_defaults['core_lo_addr'] = '127.0.0.1'
+    ip_defaults['core_af_inet'] = AF_INET
+
+if use_ipv6_by_default:
+    ip_defaults['af_inet'] = AF_INET6
+    ip_defaults['null_addr'] = '::'
+    ip_defaults['null_cidr'] = '::/0'
```
temoto: null_cidr wasn't used. Is it valuable?

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
+"""
```
temoto: Overall, I think it belongs to greenio/base, probably no need to create another module.

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+
+def check_ipv6_lo_addr():
+    try:
+        sock = socket(AF_INET6, SOCK_STREAM)
+        sock.bind(('::1', 0))
+        return True
+    except:
+        return False
+
+def check_ipv6_hosts_file():
```
temoto: Function name is misleading. It doesn't check hosts file.

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
...
+    return False
+
+core_use_ipv6_lo_addr = check_ipv6_lo_addr()
+use_ipv6_by_default = check_ipv6_hosts_file()
+
+ip_defaults = {}
+
+if core_use_ipv6_lo_addr:
+    ip_defaults['core_lo_addr'] = '::1'
+    ip_defaults['core_af_inet'] = AF_INET6
```
temoto: `ipdefaults[core_af_inet]` vs `ipdefaults[af_inet]` distinction is perplexing. Overall, I think word `core` in this file doesn't clarify things. Maybe I'm not familiar with some common knowledge?

eventlet/support/afinet.py
```diff
@@ -0,0 +1,46 @@
+"""
+Test availabilty of AF_INET6 on a machine
+
+Uses socket API to determine configuration
+"""
+
+from socket import *
+
+def check_ipv6_lo_addr():
```
temoto: What's the purpose of this check anyway? The big picture, I mean. Suppose my machine is able to listen `::1`, why would you hack Eventlet to prefer IPv6 while rest of system doesn't?

temoto:
@grantma ping

temoto:
Closing as conversation is not continued.

PR #382, opened 2017-01-26
wsgi: move write access log info to own function

scarchik:
Modify wsgi.py: 
move write access log info to own function for better way to override and modify access log.

Due to necessity to write this info log in json format and add to log additional information from headers.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -554,17 +554,12 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
                 hook(self.environ, *args, **kwargs)
 
             if self.server.log_output:
-                client_host, client_port = self.get_client_address()
-
-                self.server.log.info(self.server.log_format % {
-                    'client_ip': client_host,
-                    'client_port': client_port,
-                    'date_time': self.log_date_time_string(),
-                    'request_line': self.requestline,
-                    'status_code': status_code[0],
-                    'body_length': length[0],
-                    'wall_seconds': finish - start,
-                })
+                self.write_access_log(
+                    status_code=status_code[0],
+                    body_length=length[0],
+                    wall_seconds=finish - start,
+                    headers=headers_set[1]
+                )
 
     def get_client_address(self):
         host, port = addr_to_host_port(self.client_address)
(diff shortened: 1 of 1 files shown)


temoto:
Can you show how this is used?

temoto:
It seems overriden function doesn't get a lot of useful information and has to resort to copying code.

scarchik:
Actually I want override this function instead of override **handle_one_response**.
I use python-json-logger to report logs in json format to GrayLog and I need add to this log request_id from headers.

My code will be like this:
`
class HttpProtocolWithRequestID(wsgi.HttpProtocol):
    def write_access_log(self, **kwargs):
        client_host, client_port = self.get_client_address()
        extra = {
                    'client_ip': self.get_client_ip(),
                    'client_port': self.client_address[1],
                    'date_time': self.log_date_time_string(),
                    'request_line': self.requestline,
                    'status_code': status_code[0],
                    'body_length': length[0],
                    'wall_seconds': finish - start,
                    'request_id': _get_request_id()
        }

        self.server.log.info(self.server.log_format % extra, extra=extra)
`

Log example:
{"asctime": "2017-01-26 16:08:50,110", "msecs": 110.32581329345703, "levelname": "INFO", "module": "eventlet_http_protocol", "lineno": 184, "args": [], "msg": "10.10.11.29 - - [26/Jan/2017 16:08:50] \"PATCH /admin/user/3 HTTP/1.1\" 500 279 0.067530", "request_id": "admin.user_update.patch.482d89a5-a13f-4822-8cf6-a580c5f4e489", "client_ip": "10.10.11.29", "client_port": 61258, "request_line": "PATCH /admin/wallet/3 HTTP/1.1", "status_code": "500", "date_time": "26/Jan/2017 16:08:50", "wall_seconds": 0.06752991676330566, "body_length": 279}

From eventlet versions 0.19 and 0.20.1 function **handle_one_response** changed that why I also need change my code. This change help resolve this problem in future.

temoto:
Okay, thanks, I understand problem and use case now. I'll modify your patch to avoid copying code like `get_client_address() call` and let you review.

Ping and angry reminders are welcome.

scarchik:
Ok. Any changes is up to you.

PR #388, opened 2017-02-15
Make hubs use a monotonic clock by default

malor:
Note, that this is essentially a simplified version of https://github.com/eventlet/eventlet/pull/303 . The idea here is to make eventlet do "the right thing" by default in the expense of adding a new dependency on monotonic.

### DIFF
--- a/eventlet/hubs/epolls.py
+++ b/eventlet/hubs/epolls.py
@@ -1,7 +1,6 @@
 import errno
 from eventlet.support import get_errno
 from eventlet import patcher
-time = patcher.original('time')
 select = patcher.original("select")
 if hasattr(select, 'epoll'):
     epoll = select.epoll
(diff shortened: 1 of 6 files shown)


temoto:
Thanks for bringing this up.

I agree on library choice, let's give people still using Python2.7 another stir in right direction, namely ctypes overhead on every clock read! :-)

Please, have a look at this commit 4c578c8a0cb85a75c22265b69d6fa72523bf84d7 it's mostly the same code with few modifications, like fewer monotonic imports. If you want to keep yourself as commit author that can be arranged.

Since it's just a pure python library there's no need in external dependency, we can just bundle it like six. And your test does not perform a useful check, so removed. Frankly, I have no idea how to construct a test simulating non-monotonic time.

malor:
Thanks, Sergey! No problem - it works for me as is as long as it's in eventlet master ;)

On the test case... Agreed. We'd probably need a functional test that would adjust the system time, but it'd be very intrusive.

temoto:
It's merged into master f92b49f77e073dad40a71858ef9149b543e07d67 and will be released on PyPI soon.

I think libfaketime may work https://github.com/wolfcw/libfaketime it seems to have "eidt rc file API" for on fly changes. Have not ever tried it.

PR #391, opened 2017-02-17
be explicit about expecting bytes from sock.recv

mattbennett:
On Python3, calling`eventlet.support.greendns._net_read` will raise "TypeError: Can't convert 'bytes' object to str implicitly".

This happens when resolution is requested over TCP, or if resolution falls back to TCP when the UDP response is truncated (which is how we found it).

The bug was fixed in dnspython [here](https://github.com/rthalley/dnspython/pull/159).

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -597,13 +597,13 @@ def getnameinfo(sockaddr, flags):
 
 
 def _net_read(sock, count, expiration):
-    """coro friendly replacement for dns.query._net_write
+    """coro friendly replacement for dns.query._net_read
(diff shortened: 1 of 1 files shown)


temoto:
Thanks, I want to have a test case on this.
If there's no test in reasonable time, LGTM from me, I'll merge or Jakub also has access.

gugu:
@temoto can you please merge it? There is no response from anyone, and greendns is buggy in python 3

temoto:
Привет, Андрей. 👍 

Fix is merged in master fa87d413751740c2c4b0b881623dc984c640b565

gugu:
Спасибо!

Issue #409, opened 2017-05-09
Patching existing locks causes RLocks being acquired by different threads


tianquan23:
For Python 2, patching existing locks replaces RLock._RLock__owner with currentthread ID no matter the old lock is locked or not and thus an unlocked RLock would have a non None owner (e.g. <_RLock owner=140106584489808 count=0>).
Then if we acquire the RLock in the same thread, the method RLock.acquire would not invoke the _RLock__block.acquire() since it treats this is a recursive acquire by checking current thread ID. And then the following RLock.release would invoke the _RLock__block.release method resulting in the counter of Semaphore being improved to 2. Thus, the RLock could be acquired by two dirrerent threads.

 ## How to reproduce:
### Code
```
import eventlet
from eventlet import greenpool
import threading

lock = threading.RLock()
eventlet.monkey_patch()

def task(a):
    with lock:
        print "[%s] Acquired lock" % a
        eventlet.sleep(1)
        print "[%s] Releasing lock" % a

# This will improve lock._RLock__block.couter to 2
lock.acquire()
lock.release()
print 'lock._RLock__block: %s' % lock._RLock__block

pool = greenpool.GreenPool()
pool.spawn(task, 'a')
pool.spawn(task, 'b')
pool.waitall()
```
### Output
```
lock._RLock__block: <Semaphore c=2 _w[0]>
[a] Acquired lock
[b] Acquired lock
[a] Releasing lock
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/hub.py", line 457, in fire_timers
    timer()
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/timer.py", line 58, in __call__
    cb(*args, **kw)
  File "/usr/local/lib/python2.7/dist-packages/eventlet/greenthread.py", line 214, in main
    result = function(*args, **kwargs)
  File "test-lock2.py", line 12, in task
    print "[%s] Releasing lock" % a
  File "/usr/lib/python2.7/threading.py", line 216, in __exit__
    self.release()
  File "/usr/lib/python2.7/threading.py", line 204, in release
    raise RuntimeError("cannot release un-acquired lock")
RuntimeError: cannot release un-acquired lock
[b] Releasing lock
```

temoto:
Fix was merged in master f72cc96a70f9e9d23fc6e00d994263829b672be0

Thank you. Have a great time.

PR #410, opened 2017-05-09
patcher: set locked RLocks' owner only when patching existing locks

tianquan23:
For Python 2, patching existing locks replaces RLock._RLock__owner with current
thread ID no matter the old lock is locked or not and thus an unlocked RLock
would have a non None owner (e.g. <_RLock owner=140106584489808 count=0>).
Then if we acquire the RLock in the same thread, the method RLock.acquire would
not invoke the _RLock__block.acquire() since it treats this is a recursive
acquire by checking current thread ID. And then the following RLock.release
would invoke the _RLock__block.release method resulting in the counter of
Semaphore being improved to 2.

There should be only two states being expected for RLock:
1. owner != None and count > 0
2. owner == None and count == 0

This patch fixs it by only setting locked RLocks' owner during patching.

#409

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -358,7 +358,7 @@ def _fix_py2_rlock(rlock, tid):
     rlock._RLock__block = new
     if old.locked():
         new.acquire()
-    rlock._RLock__owner = tid
+        rlock._RLock__owner = tid
 
 
 def _fix_py3_rlock(old):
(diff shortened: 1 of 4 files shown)


temoto:
Thank you.

New test is skipped on CPython 2.7.10. Do you know what versions of Python would actually run this test?

tianquan23:
Sorry, I made a mistake. I was supposed to skip the test when there is _thread.RLock on Python 3, since _thread.RLock() wouldn't be tracked by gc, resulting in _green_existing_locks not working. I was wondering if existing RLocks would really be green on Python 3. However, I thought it might be unrelated to this patch, maybe have another patch addressing the issue.

Issue #413, opened 2017-05-13
patch module calls raise Error changes


Linbing:
when using eventlet.monkey_patch(), it's will call patch module raise Error change, linke subprocess.check_output or check_call, when somtthing error happened, it's will raise CalledProcessError(retcode, cmd), and this CalledProcessError maybe change in eventlet.green.subprocess, so, we can not except this error like subprocess.CalledProcessError, instead of, we must except this error eventlet.green.subprocess.CalledProcessError. 

example:

#!/usr/bin/python

import eventlet
import subprocess

eventlet.monkey_patch()

try:
        subprocess.check_output(["xxx"],shell=True)
except eventlet.green.subprocess.CalledProcessError as e:
    print ">> except in eventlet.green.subprocess.CalledProcessError"
except subprocess.CalledProcessError as e:
    print ">> except in subprocess.CalledProcessError "

temoto:
Thank you, the problem is clear.

temoto:
This revealed the ugly side of monkey patching, when original stdlib modules are saved for `eventlet.patcher.original()` function, they are re-imported from scratch, which creates another copy of exception classes in memory.

I'm not sure where things start to get wrong, don't have a good idea how to fix this right now. Stay tuned.

Linbing:
@temoto see pull/414, inject will return new reimport module. so the exception classes will take the same.

Linbing:
CalledProcessError has been set in globals(), and he is from the new import subprocess module, so when after inject. sys.modules.subprocess keep older, that's why  we can't using subprocess.CalledProcessError(sys.modules) to except the error.

temoto:
Yeah I understand what's going on, so far the best solution I can do is to rebind certain names from stdlib to green module. Please, try this patch eac49d80bbdb2245173f46d4cf3e1f19b33f5462

`pip install https://github.com/eventlet/eventlet/archive/eac49d80bbdb2245173f46d4cf3e1f19b33f5462.zip`

Linbing:
@temoto i have modify like your patch before, but it's not resolved.  we can't do that like 

CalledProcessError = subprocess_orig.CalledProcessError
or
CalledProcessError = sys.modules['subprocess'].CalledProcessError

both or that will change CalledProcessError in eventlet.green.subprocess.

so we can't define CalledProcessError in eventlet.green.subprocess

Linbing:
[...]

PR #414, opened 2017-05-14
Issue $413 patcher diverts subprocess exceptions instead of keeping c…

Linbing:
when patch inject, it's will reimport module, this will make copy of class in memory, so make the new reimport module in sys.modules, this will keey the same sys.modules with eventlet.green.module

reset new reimport module in sys.modules.

https://github.com/eventlet/eventlet/issues/413


temoto:
I'm sorry this is not acceptable solution because it makes eventlet contagious: just `import eventlet.green.subprocess` also makes previous and next `import subprocess` green.

PR #417, opened 2017-05-30
Support for the permessage-deflate websocket extension

costasgambit:
Added support for this very useful extension, as supported by chrome, firefox, and most other popular browsers.

Tested with Python 2.7 and 3.4.

### DIFF
--- a/eventlet/websocket.py
+++ b/eventlet/websocket.py
@@ -552,7 +552,7 @@ class RFC6455WebSocket(WebSocket):
                                                    zlib.MAX_WBITS))
 
         if options.get("server_no_context_takeover" if self.client
-                       else"client_no_context_takeover"):
+                       else "client_no_context_takeover"):
             # This option means we have to make a new one every time
             return _make()
         else:
@@ -651,12 +651,16 @@ class RFC6455WebSocket(WebSocket):
 
     def _recv_frame(self, message=None):
         recv = self._get_bytes
+
+        # Unpacking the frame described in Section 5.2 of RFC6455
+        # (https://tools.ietf.org/html/rfc6455#section-5.2)
         header = recv(2)
         a, b = struct.unpack('!BB', header)
         finished = a >> 7 == 1
         rsv123 = a >> 4 & 7
+        rsv1 = rsv123 & 4
         if rsv123:
-            if rsv123 & 0x04 and "permessage-deflate" not in self.extensions:
+            if rsv1 and "permessage-deflate" not in self.extensions:
                 # must be zero - unless it's compressed then rsv1 is true
                 raise FailedConnectionError(
                     1002,
@@ -698,7 +702,7 @@ class RFC6455WebSocket(WebSocket):
         received = 0
         if not message or opcode & 8:
             decoder = self.UTF8Decoder() if opcode == 1 else None
-            decompressor = self._get_permessage_deflate_dec(rsv123 & 0x04)
+            decompressor = self._get_permessage_deflate_dec(rsv1)
             message = self.Message(opcode, decoder=decoder, decompressor=decompressor)
         if not length:
             message.push(b'', final=finished)
--- a/eventlet/websocket.py
+++ b/eventlet/websocket.py
@@ -9,6 +9,8 @@ import struct
 import sys
 import time
 
+import zlib
+
 try:
     from hashlib import md5, sha1
 except ImportError:  # pragma NO COVER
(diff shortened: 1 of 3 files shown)


Faboor:
Relative code coverage for this PR is with the addition of test at above 90%, however codevoc.io seems to not like rebasing PRs.

Not sure what the failed test is about, nor how to deal with. Can't see how these changes could have caused it.

eventlet/websocket.py
```diff
@@ -528,25 +528,31 @@ def _get_permessage_deflate_enc(self):
...
-    def _get_permessage_deflate_dec(self):
+    def _get_permessage_deflate_dec(self, rsv1):
         options = self.extensions.get("permessage-deflate")
-        if options is None:
+        if options is None or not rsv1:
             return None
 
         def _make():
-            return zlib.decompressobj(-options.get("client_max_window_bits", zlib.MAX_WBITS))
+            return zlib.decompressobj(-options.get("server_max_window_bits" if self.client
```
temoto: It says `server_... if self.client`. Is that correct?

eventlet/websocket.py
```diff
@@ -692,7 +698,7 @@ def _recv_frame(self, message=None):
         received = 0
         if not message or opcode & 8:
             decoder = self.UTF8Decoder() if opcode == 1 else None
-            decompressor = self._get_permessage_deflate_dec()
+            decompressor = self._get_permessage_deflate_dec(rsv123 & 0x04)
```
temoto: please leave a comment about magic number mask

tests/websocket_new_test.py
```diff
@@ -228,3 +229,245 @@ def error_detector(environ, start_response):
...
+        ]:
+            sock = eventlet.connect(self.server_addr)
+
+            sock.sendall(six.b(self.connect % extension))
+            result = sock.recv(1024)
+
+            # The server responds the correct Websocket handshake
+            # print('Extension offer: %r' % extension)
+            match = re.match(self.handshake_re, result)
+            self.assertIsNotNone(match)
```
temoto: `assert match is not None`
`assert len(match.groups()) == 1`

I know we still have other code using Java style asserts, please use assert statement in new code.

tests/websocket_new_test.py
```diff
@@ -228,3 +229,245 @@ def error_detector(environ, start_response):
...
+        ws.send(b'Hello')
+        msg1 = self.get_deflated_reply(ws)
+        self.assertEqual(msg1, b'\xf2\x48\xcd\xc9\xc9\x07\x00')
+
+        ws.send(b'Hello')
+        msg2 = self.get_deflated_reply(ws)
+        self.assertEqual(msg2, b'\xf2\x00\x11\x00\x00')
+
+        ws.close()
+        eventlet.sleep(0.01)
```
temoto: I fear this small sleep at end of test. What is it for?

eventlet/websocket.py
```diff
@@ -528,25 +528,31 @@ def _get_permessage_deflate_enc(self):
...
             return None
 
         def _make():
-            return zlib.decompressobj(-options.get("client_max_window_bits", zlib.MAX_WBITS))
+            return zlib.decompressobj(-options.get("server_max_window_bits" if self.client
+                                                   else "client_max_window_bits",
+                                                   zlib.MAX_WBITS))
 
-        if options.get("client_no_context_takeover"):
+        if options.get("server_no_context_takeover" if self.client
```
temoto: Again, `server_... if client`. Is it intended?

eventlet/websocket.py
```diff
@@ -528,25 +528,31 @@ def _get_permessage_deflate_enc(self):
...
+                       else"client_no_context_takeover"):
```
temoto: Keyword glued to quoted string. Definitely doesn't pass PEP-8.

temoto:
Don't worry about irrelevant Travis failures. We're making slow progress towards deterministic builds. Not there yet.

tests/websocket_new_test.py
Faboor: Honestly, I don't know. When creating these tests, I started from [`test_send_recv_13()`](https://github.com/costasgambit/eventlet/blob/ab2300a5eded021c6b4dd07bcf07f032526d4557/tests/websocket_new_test.py#L113), which I assumed to be correct and uses the same [short sleep](https://github.com/costasgambit/eventlet/blob/ab2300a5eded021c6b4dd07bcf07f032526d4557/tests/websocket_new_test.py#L134). I just copied it, thinking it might be important to allow to finish the cleanup after the test.

eventlet/websocket.py
Faboor: Surprisingly, pep8 (when run with tox) didn't complained when run locally, nor on the [Travis build](https://travis-ci.org/eventlet/eventlet/jobs/273255441#L595).

eventlet/websocket.py
Faboor: Yes, this is correct. During decompression, we want to be using the same settings as our counterparty. So if the websocket is acting as it client, it should use the negotiated server settings during decompression and vice versa.

Technically, during decompression, either party could be using a larger LZ77 sliding window and always allow context takeover, but limiting it to the agreed setting of the other party helps save on memory needed for the decompression.

eventlet/websocket.py
Faboor: Intended, see above

Faboor:
Thank you very much for the feedback. Is it safe to squash changes into the last commit, or will it corrupt the comments requesting changes? (and do we even care about those?)

eventlet/websocket.py
Faboor: A similar magic number thing happens around [line 657](https://github.com/costasgambit/eventlet/blob/ab2300a5eded021c6b4dd07bcf07f032526d4557/eventlet/websocket.py#L657) with shifts. Would it be better add `rsv1 = rsv123 & 4` or `rsv1 = rsv123 >> 2` somewhere around that part and use just `rsv1` later? Possibly adding a comment with a link to [Section 5.2 of RFC6455](https://tools.ietf.org/html/rfc6455#section-5.2) or copying the first 8 bits of the diagram in that section.

temoto:
I will squash no matter what. Use separate commits or not - as you like. I don't know whether it will corrupt comments. Useful comments should stay in code. Github comments related to patch will be forgotten next day, so don't worry about that.

eventlet/websocket.py
temoto: Thanks for explain.

eventlet/websocket.py
temoto: Yeah, I checked Travis PEP-8 build and was surprised also.

Still, assume glued keyword doesn't pass a check that is not automated for some weird reason.

eventlet/websocket.py
temoto: Please add link to RFC at first encounter of `rsv...` name.

tests/websocket_new_test.py
temoto: I see, leave it then. I will rewrite it some boring winter day.

Faboor:
I addressed the changes requested and added a few new tests to make sure the `server_... if self.client` is correct and stays correct in future revisions :)

tests/websocket_new_test.py
```diff
@@ -317,17 +316,17 @@ def test_accept_deflate_ext_window_max_bits_13(self):
...
             offered_parts = [part.strip().lower() for part in extension_string.split(';')]
             offered_parts_names = [part.split('=')[0].strip() for part in offered_parts]
             offered_parts_dict = dict(zip(offered_parts_names[1:], vals))
 
             accepted_ext_parts = match.groups()[0].decode().split('; ')
-            self.assertEqual(accepted_ext_parts[0], 'permessage-deflate')
+            assert accepted_ext_parts[0] == 'permessage-deflate'
             for param, val in (part.split('=') for part in accepted_ext_parts[1:]):
-                self.assertLessThanEqual(int(val), offered_parts_dict[param])
+                assert int(val) == offered_parts_dict[param]
```
temoto: `assertLessThanEqual -> assert <=`
are you sure strict equal is proper here?

Faboor: The specification says, it has to be less or equal, but our implementation is always using the offered value, thus avoiding possible conflicts. While I initially thought to use `<=`, I decided it provides no/very little additional value to allow the code to return a smaller value than specified in the negotiation offer and future revisions should attempt to use the same value as well.

temoto: Thanks.

temoto:
@costasgambit @Faboor thank you very much. Have good fun.

costasgambit:
Excellent, thanks everyone.

PR #421, opened 2017-07-02
Use tox's `TOXENV` environment variable

asottile:


temoto:
How is this useful?

asottile:
instead of

```
export TOX_ENV=foo
tox -e $TOX_ENV
```

`tox` has an environment variable designed for this:

```
export TOXENV=foo
tox
```

This is the pattern [used in many other repositories](https://github.com/pypa/pip/blob/d2574efdcf37a3f9f105c32746ca14f02daa7cd9/.travis.yml) and seems to be more "standard"

temoto:
Thank you for patch and education. Have a great time.

PR #423, opened 2017-07-10
Open the hosts file with 'r'

### LINKED ISSUE #422: DeprecationWarning: 'U' mode is deprecated (since Python 3.4)
https://github.com/eventlet/eventlet/blob/master/eventlet/support/greendns.py#L183
```with open(self.fname, 'rU') as fp:```

https://docs.python.org/3/library/functions.html#open

### PR DESCRIPTION
**Pull request by frexvahi**
Fixes #422

Also corrects the handling of comments which do not start at the first character of the line

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -182,7 +182,7 @@ class HostsResolver(object):
         if sys.version_info < (3, 4):
             mode = 'rU'
         else:
-            mode = 'r'
+            mode = 'rt'
         try:
             with open(self.fname, mode=mode) as fp:
                 for line in fp:
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -180,10 +180,11 @@ class HostsResolver(object):
         """
         lines = []
         try:
-            with open(self.fname, 'rU') as fp:
+            with open(self.fname, 'r') as fp:
                 for line in fp:
+                    line = line.split('#')[0]
                     line = line.strip()
-                    if line and line[0] != '#':
+                    if line:
                         lines.append(line)
         except (IOError, OSError):
             pass
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -157,7 +157,7 @@ class HostsResolver(object):
     def __init__(self, fname=None, interval=HOSTS_TTL):
         self._v4 = {}           # name -> ipv4
         self._v6 = {}           # name -> ipv6
-        self._aliases = {}      # name -> cannonical_name
+        self._aliases = {}      # name -> canonical_name
         self.interval = interval
         self.fname = fname
         if fname is None:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: eventlet/support/greendns.py [hunk 1]
```diff
@@ -180,10 +180,11 @@ def _readlines(self):
         """
         lines = []
         try:
-            with open(self.fname, 'rU') as fp:
+            with open(self.fname, 'r') as fp:
                 for line in fp:
+                    line = line.split('#')[0]
```
temoto: `.split('#', 1)[0]`

temoto:
Thanks for test.

eventlet/support/greendns.py
```diff
@@ -180,10 +180,11 @@ def _readlines(self):
         """
         lines = []
         try:
-            with open(self.fname, 'rU') as fp:
+            with open(self.fname, 'r') as fp:
```
temoto: `open(..., 'rb')`

always work explicitly with either bytes or strings. Files are obviously bytes.

temoto: That's not only dictatory rule, but also a practical reason for Python2 tests fail, see Travis build link below.

eventlet/support/greendns.py
```diff
@@ -179,11 +179,16 @@ def _readlines(self):
         Note that this performs disk I/O so can be blocking.
         """
         lines = []
+        if sys.version_info < (3, 4):
+            mode = 'rU'
+        else:
+            mode = 'r'
```
temoto: Again. `open(..., 'rb')` Always work explicitly with either bytes or strings. Files are obviously bytes.

It seems possible to read a text file without checking Python version.

frexvahi: It's a text file, why should I be using 'rb'?

frexvahi: The reason for the version check is, earlier versions have 'universal newlines' off by default, while newer versions have it on by default and give a `DeprecationWarning` if the 'U' flag is passed anyway.

frexvahi: in fact, 'rb' is exactly the opposite of what I want, as it will (among other things) turn off the 'universal newlines' feature.

frexvahi: I changed it to 'rt' instead

temoto: Why use `rb`:

https://docs.python.org/2/library/functions.html#open

> ... The default is to use text mode, which may convert '\n' characters to a platform-specific representation on writing and back on reading. Thus, when opening a binary file, you should append 'b' to the mode value to open the file in binary mode, which will improve portability. (Appending 'b' is useful even on systems that don’t treat binary and text files differently, where it serves as documentation.)

AFAIK, the only platform with explicit support for "text files" nonsense is Windows.

I think it would be more efficient for computing resource usage and for code clarity to just read whole file in one big read and iterate lines broken by any newline-like character. Which is what universal lines do, but without unnecessary substitution of line ending characters.

frexvahi: 'rb' means that everything that was being done 'magically' (and in C, i.e. faster) by `open` has to be done explicitly. However, if that's what you want, you can use #426 instead.

PR #426, opened 2017-07-13
Read /etc/hosts in binary mode

frexvahi:
This is a rewrite of #423 based on comments from @temoto

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -32,6 +32,7 @@
 # THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import re
 import struct
 import sys
 
@@ -154,6 +155,14 @@ class HostsResolver(object):
     :interval: The time between checking for hosts file modification
     """
 
+    LINES_RE = re.compile(r"""
+        \s*  # Leading space
+        ([^\r\n#]+?)  # The actual match, non-greedy so as not to include trailing space
+        \s*  # Trailing space
+        (?:[#][^\r\n]+)?  # Comments
+        (?:$|[\r\n]+)  # EOF or newline
+    """, re.VERBOSE)
+
     def __init__(self, fname=None, interval=HOSTS_TTL):
         self._v4 = {}           # name -> ipv4
         self._v6 = {}           # name -> ipv6
@@ -184,19 +193,10 @@ class HostsResolver(object):
         except (IOError, OSError):
             return []
 
-        lines = []
-        # Split by both \r and \n
-        # This introduces bogus empty lines for '\r\n' line breaks, but
-        # that doesn't matter because empty lines are ignored
-        for nline in fdata.split(b'\n'):
-            for line in nline.split(b'\r'):
-                if not isinstance(line, str):
-                    line = line.decode()
-                line = line.split('#', 1)[0]  # remove comments
-                line = line.strip()
-                if line:
-                    lines.append(line)
-        return lines
+        if not isinstance(fdata, str):  # Python 3
+            fdata = fdata.decode()
+
+        return self.LINES_RE.findall(fdata)
 
     def _load(self):
         """Load hosts file
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -157,7 +157,7 @@ class HostsResolver(object):
     def __init__(self, fname=None, interval=HOSTS_TTL):
         self._v4 = {}           # name -> ipv4
         self._v6 = {}           # name -> ipv6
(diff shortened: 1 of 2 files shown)


temoto:
Fully agree on C speed, we should use it when possible.
Nested loop on `\r` and `\n` is absolutely not acceptable, I'm sorry for being not entirely clear if that's what you read from my comment.

How do you like this version? I bet it's faster than universal newlines version for efficient removal of whitespace and comments.

```
import re
RE_HOSTS_LINES = re.compile(br'\s*([^\r\n#]+?)(?:\s*#[^\r\n]+)?(?:$|[\r\n]+)')

    def _readlines(self):
        """Read the contents of the hosts file

        Return list of lines, comment lines and empty lines are
        excluded.

        Note that this performs disk I/O so can be blocking.
        """
        try:
            with open(self.fname, 'rb') as fp:
                content = fp.read()
        except (IOError, OSError):
            return []
        return RE_HOSTS_LINES.findall(content)
```

frexvahi:
Your solution returns bytes on Python 3

frexvahi:
I decided it was better to decode before parsing with the regular expression, just in case some idiotic windows system writes its hosts file in UCS2

eventlet/support/greendns.py
```diff
@@ -178,16 +187,16 @@ def _readlines(self):
...
-                    if line and line[0] != '#':
-                        lines.append(line)
+            with open(self.fname, 'rb') as fp:
+                fdata = fp.read()
         except (IOError, OSError):
-            pass
-        return lines
+            return []
+
+        if not isinstance(fdata, str):  # Python 3
```
temoto: Again, it's very possible to read a text file without checking Python version. We asked for bytes, so decode must be applicable unconditionally.

You may want to add `ignore` flag to decode.

temoto:
Good catch with bytes, thanks.

eventlet/support/greendns.py
frexvahi: Yes, it is possible to read a text file without checking python version, by reading it in text mode. However, you said I had to read it in bytes mode.

frexvahi: The `_readlines` function returns a list of `str`. On Python 2, `str` is the same as `bytes`. On Python 3 it is not. That is why I have to check the Python version.

temoto: 2/3 type names and aliases get you confused. Think conceptually binary or text. You open file in binary mode regardless platform. Function must return text always. It means you have to `.decode()` always.

```
% python2.7                                                        :(
Python 2.7.10 (default, Jul 14 2015, 19:46:27)
>>> s = "text"
>>> type(s)
<type 'str'>
>>> s.decode('utf-8', 'ignore')
u'text'
% python3.6
Python 3.6.0 (default, Dec 24 2016, 08:02:28)
>>> s = b"text"
>>> type(s)
<class 'bytes'>
>>> s.decode('utf-8', 'ignore')
'text'
```

frexvahi: Yes, and on python 2, `isinstance(u'text', str)` is `False`

frexvahi: There should really be some function (e.g. `bytes_to_str`) in the `six` package to deal with this problem, but I couldn't find anything suitable. So there are two solutions - check the version directly, or read in text mode.

frexvahi: If you don't believe me, try running `decode` always and see if the tests pass.

temoto: > Yes, and on python 2, isinstance(u'text', str) is False

How come this check is relevant? Please explain chain of thought.

_readlines does not have to return exact `str` type. It must return conceptually text, which is `unicode` in 2 and `str` in 3.

> There should really be some function (e.g. bytes_to_str)

`.decode()` is that function. It works for both Python versions types representing binary.

frexvahi: On Python 2, the return value of `bytes.decode()` is not a `str`, therefore it is not usable as `bytes_to_str`

frexvahi: OK, so you want me to change the interface so that it always now expects the DNS query to be a `unicode` even on Python 2?

frexvahi: Ah OK, I didn't realise that, even though `isinstance(u'text', str)` is `False`, `'text' == u'text'` is `True` and they are also compatible for dictionary lookups. Python 2 is really messed up.

temoto: Python is new PHP. It never promised logical design, just that everything is going to work somehow. :-)

frexvahi:
One of the CI jobs failed with `TestIsTakingTooLong` on an unrelated test, could you restart it please?

eventlet/support/greendns.py
jstasiak: I'd say just focus on the code being correct on Python 3 ;)

frexvahi: It now works, barring the unrelated failure. Can you rerun the failed job and then merge it?

temoto:
Resquashing outside Github changed commit id if that matters for future reader.

temoto:
@frexvahi thank you very much. Have a great time.

PR #431, opened 2017-08-11
socket: context manager support

### LINKED ISSUE #430: socket.create_connection cannot be used as a context manager
Consider the following script:

```python
import socket
with socket.create_connection(('google.com', 80)) as s:
    pass
```

With the regular standard library socket (PY3 only), this works fine. If the library is monkey patched, though, I get the following:

```
(py36) $ python test.py
  File "test.py", line 6, in <module>
    with socket.create_connection(('google.com', 80)) as s:
AttributeError: __enter__
```

The green version of `socket` seems to be missing the context manager `__enter__` and `__exit__` methods.

miguelgrinberg:
Fixes #430 

This PR adds the `__enter__` and `__exit__` context manager methods to the green socket class.

### DIFF
--- a/eventlet/greenio/base.py
+++ b/eventlet/greenio/base.py
@@ -431,6 +431,12 @@ class GreenSocket(object):
     def gettimeout(self):
         return self._timeout
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
+
     if "__pypy__" in sys.builtin_module_names:
         def _reuse(self):
             getattr(self.fd, '_sock', self.fd)._reuse()


eventlet/greenio/base.py
```diff
@@ -431,6 +431,13 @@ def settimeout(self, howlong):
     def gettimeout(self):
         return self._timeout
 
+    if six.PY3:
```
temoto: Please don't deny Python2 users of this feature.

miguelgrinberg: You don't mind that the std library in Python 2 doesn't have this? Happy to make it available to Python 2 as well, I just thought it better to try to be compatible with the standard library.

temoto: A+B is compatible with A.

miguelgrinberg: Okay, I removed the PY3 check.

temoto:
Thank you very much.

Have a great life.

PR #438, opened 2017-10-04
Drop OpenSSL.rand support

hguemar:
PyOpenSSL deprecated OpenSSL.rand in 17.2.0 and removed it in 17.3.0.
To comply, update min version in tests to 17.3.0 and removes it.
https://pyopenssl.org/en/stable/changelog.html


temoto:
Thank you.

PR #444, opened 2017-11-02
wsgi: improve error message when returning without start_response

AGSPhoenix:
Today I discovered that if you return from your WSGI function without calling start_response() first, you get a rather unhelpful traceback:

```
  File "your-python/site-packages/eventlet/wsgi.py", line 515, in handle_one_response
    'Content-Length' not in [h for h, _v in headers_set[1]]:
IndexError: list index out of range
```
That you have to call start_response() before returning is WSGI 101, but I probably should have been asleep a few hours ago, so it took me a while to work backwards from this traceback to the error in my code. This patch should be a relatively simple change that makes it clear what the issue is.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -537,6 +537,9 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
                     self.close_connection = 1
                     return
 
+                if not headers_set:
+                    raise AssertionError("return before start_response()")
+
                 # Set content-length if possible
                 if not headers_sent and hasattr(result, '__len__') and \
                         'Content-Length' not in [h for h, _v in headers_set[1]]:


AGSPhoenix:
Just took a look at the CI results, and this definitely breaks some things. Given the number of failed tests, I looked over PEP-333  to check my assumption about start_response being required and found this:

>(Note: the application **must** invoke the `start_response()` callable before the iterable yields its first body string, so that the server can send the headers before any body content. However, this invocation **may** be performed by the iterable's first iteration, so servers **must not** assume that `start_response()` has been called before they begin iterating over the iterable.)

...which is exactly what my PR does. I'd still like for this to be changed, but it's late and I don't have any clever ideas for how to go about it. Moving this to an issue.

AGSPhoenix:
Would it make sense to move the Content-Length check and warning about start_response() after the first read of the returned data, but before any data is sent? That should satisfy the spec, although simply moving it into the for loop seems clunky.

temoto:
Ah, now I think AssertionError is what has broken tests. It's an exception raised with `assert` statement. Try plain `Exception` or create custom subclass, though I don't believe it could be helpful.

Also, you're breaking all generator applications, because their call returns instantly and they execute only while iterating `result`. So, yes, moving start_response check into loop is not only sensible, it's required to support generator WSGI apps. Which every performance obsessed application must be.

Moving content-length check is orthogonal to this matter. It would be better for sure. But please make it a separate patch/commit.

Issue #446, opened 2017-11-03
spurious "Connection reset by peer" errors in wsgi+ssl


stefannica:
The following behavior has been observed in an openstack HA environment configured with SSL enabled and having an HAProxy instance running in front of the API services. All openstack API services which rely on eventlet are flooding /var/log/messages with errors of the form:

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/eventlet/greenpool.py", line 82, in _spawn_n_impl
    func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/eventlet/wsgi.py", line 734, in process_request
    proto.__init__(sock, address, self)
  File "/usr/lib64/python2.7/SocketServer.py", line 652, in __init__
    self.handle()
  File "/usr/lib64/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/site-packages/eventlet/wsgi.py", line 339, in handle_one_request
    self.raw_requestline = self.rfile.readline(self.server.url_length_limit)
  File "/usr/lib64/python2.7/socket.py", line 480, in readline
    data = self._sock.recv(self._rbufsize)
  File "/usr/lib/python2.7/site-packages/eventlet/green/ssl.py", line 194, in recv
    return self._base_recv(buflen, flags, into=False)
  File "/usr/lib/python2.7/site-packages/eventlet/green/ssl.py", line 227, in _base_recv
    read = self.read(nbytes)
  File "/usr/lib/python2.7/site-packages/eventlet/green/ssl.py", line 139, in read
    super(GreenSSLSocket, self).read, *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/eventlet/green/ssl.py", line 113, in _call_trampolining
    return func(*a, **kw)
  File "/usr/lib64/python2.7/ssl.py", line 673, in read
    v = self._sslobj.read(len)
error: [Errno 104] Connection reset by peer

```
This is caused by the HAproxy health checker, which resets the TCP connection with a RST/ACK instead of going through the graceful connection close handshake.

The errors don't actually cause the API services to fail, they seem to be harmless in that respect. However, they do pollute the /var/log/messages file an awful lot, which makes them a nuisance.

vuntz:
@stefannica can be closed, I guess?

PR #447, opened 2017-11-03
wsgi: handle remote connection resets

stefannica:
While processing a WSGI request, if the remote client forcefully
resets the connection while the request data is being read from the
socket (e.g. HAProxy health check), the ECONNRESET socket error is
not properly handled and may end up being logged in strerr or in
a log file.

https://github.com/eventlet/eventlet/issues/446


temoto:
Do not call get_errno twice, performance police is sad.

Why do you think this condition doesn't deserve a line in log?

stefannica:
Thanks @temoto, I updated the PR.

> Do not call get_errno twice, performance police is sad.

Good point, thanks.

> 
> Why do you think this condition doesn't deserve a line in log?

This condition is ignored in the non-SSL case (see for instance [here](https://github.com/eventlet/eventlet/blob/master/eventlet/green/ssl.py#L247)), which is why this is only seen when SSL is enabled. Nevertheless, I added a log line for it, but only if debug is enabled. The point of this change is to keep the logs from being flooded.

eventlet/wsgi.py
```diff
@@ -381,7 +381,12 @@ def _read_request_line(self):
         except greenio.SSL.ZeroReturnError:
             pass
         except socket.error as e:
-            if support.get_errno(e) not in BAD_SOCK:
+            erno = support.get_errno(e)
```
temoto: Why does this line introduce error on line 389?

stefannica: Yeah, that would be my typo. Thank you for spotting that.

temoto: Nah, typo is not the point. There will be more typos, it's fine. Problem is that tiny typo coerces to another name.

This line introduces error because name `erno` is only 1 letter distance away from `errno`. Defensive coding says make things sufficiently distinct, to avoid errors in writing and, perhaps more importantly, in reading.

If it was called `bQ8kk` or `current_errno` or something else distinctive, then it would be hard to typo into existing name even in the middle of hangover night.

Now latest patch shadows `errno` module name. It works, but name shadowing is road to trouble in general.

If you are sick from picky review, please let me know, I will change code and keep your authorship to commit history.

Thanks for your work.

stefannica: I completely agree. It's just so easy to lose track of these common sense rules when you're going for a quick patch, so thank you for taking the time to offer your insight and for reminding me.

stefannica:
@temoto, could you give this another look ? Thanks.

PR #452, opened 2017-12-01
Avoid dependency on enum-compat

javacruft:
Directly depend on enum34, scoping requirement to < Python 3.4.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setuptools.setup(
     url='http://eventlet.net',
     packages=setuptools.find_packages(exclude=['benchmarks', 'tests', 'tests.*']),
     install_requires=(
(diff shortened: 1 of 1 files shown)


onovy:
+1

enum-compat is not compatible with Debian/Ubuntu packaging and we must to patch this in package.

Thanks.

temoto:
LGTM
@jstasiak what do you say?

jstasiak:
On December 1, 2017 7:26:59 PM GMT+01:00, Sergey Shepelev <notifications@github.com> wrote:
>LGTM
>@jstasiak what do you say?

Looks good
-- 
Jakub Stasiak

temoto:
@javacruft thank you. Have fun.

PR #459, opened 2018-01-12
greendns: Treat /etc/hosts entries case-insensitive

rhafer:
Hostname in /etc/hosts are not case-sensitive, this fixes
HostsResolver() accordingly.

eventlet#458

Co-Authored-By: Thomas Bechtold <tbechtold@suse.com>


temoto:
Thank you. I'll merge after adding test.

PR #472, opened 2018-03-08
tpool: support dynamic thread increase

Akrog:
Tpool's number of threads defaults to 20, or the value defined in
environmental variable "EVENTLET_THREADPOOL_SIZE", and this value can be
changed programmatically using method "set_num_threads", but the call to
"set_num_threads" will only have any effect if the "execute" method has
not been called before (since it automatically calls the "setup"
method).

This is usually not a problem for most projects, but when you are
dealing with a big project that relies on many libraries it may be
harder to control if any of those libraries have already called execute
before you try to set the initial its value.

For example in OpenStack most projects rely on a series of common
libraries that handle Services, Messaging, Configurations, DB access,
etc.  And if we want to allow this value to be set in a configuration
file we may encounter some difficulties, because in some cases by the
time we have read the configuration file we are already running a
service, connected to a DB, connected to RabbitMQ, etc.

This patch proposes a new mechanism where we can dynamically increase
the number of threads in the library calling the "set_num_threads"
method at any point in time, which will not only resolve the situation
mentioned before, but it will also provide a nice mechanism to
dynamically increase the value at runtime if necessary.

Decreasing the number of threads is not supported in order to prevent
gaps in the thread numbering.

### DIFF
--- a/eventlet/tpool.py
+++ b/eventlet/tpool.py
@@ -18,6 +18,7 @@ import imp
 import os
 import sys
 import traceback
+import warnings
 
 import eventlet
 from eventlet import event, greenio, greenthread, patcher, timeout
@@ -261,7 +262,6 @@ def setup():
 
     assert _nthreads >= 0, "Can't specify negative number of threads"
     if _nthreads == 0:
-        import warnings
         warnings.warn("Zero threads in tpool.  All tpool.execute calls will\
             execute in main thread.  Check the value of the environment \
             variable EVENTLET_THREADPOOL_SIZE.", RuntimeWarning)
@@ -282,16 +282,21 @@ def setup():
     _rsock = greenio.GreenSocket(csock)
     _rsock.settimeout(None)
 
-    for i in six.moves.range(_nthreads):
-        t = threading.Thread(target=tworker,
-                             name="tpool_thread_%s" % i)
-        t.setDaemon(True)
-        t.start()
-        _threads.append(t)
+    # Create initial `_nthreads` threads
+    _spawn_threads()
 
     _coro = greenthread.spawn_n(tpool_trampoline)
     # This yield fixes subtle error with GreenSocket.__del__
     eventlet.sleep(0)
+
+
+def _spawn_threads():
+    for i in six.moves.range(len(_threads), _nthreads):
+        t = threading.Thread(target=tworker,
+                             name="tpool_thread_{0}".format(i + 1))
+        t.setDaemon(True)
+        t.start()
+        _threads.append(t)
 
 
 # Avoid ResourceWarning unclosed socket on Python3.2+
@@ -333,4 +338,11 @@ def killall():
 
 def set_num_threads(nthreads):
     global _nthreads
-    _nthreads = nthreads
+    old_nthreads, _nthreads = _nthreads, nthreads
+
+    if _setup_already:
+        if old_nthreads > nthreads:
+            warnings.warn("Dynamically reducing the number of threads is not "
+                          "supported and will be ignored.", RuntimeWarning)
+        else:
+            _spawn_threads()
--- a/doc/threading.rst
+++ b/doc/threading.rst
@@ -26,5 +26,12 @@ The simplest thing to do with :mod:`~eventlet.tpool` is to :func:`~eventlet.tpoo
 
 By default there are 20 threads in the pool, but you can configure this by setting the environment variable ``EVENTLET_THREADPOOL_SIZE`` to the desired pool size before importing tpool.
 
+Alternatively you can also set the number of threads in the pool to any value
+after importing tpool using the ``set_num_threads`` method as long as you
+haven't called the ``execute`` method.  If the ``execute`` method has already
+been called, only increases to the thread pool size will take immediate effect,
+lower values will not take effect until we kill all the existing threads with a
+call to ``killall``.
+
 .. automodule:: eventlet.tpool
 	:members:
--- a/tests/tpool_test.py
+++ b/tests/tpool_test.py
@@ -305,9 +305,43 @@ class TestTpool(tests.LimitedTestCase):
         self.assertRaises(eventlet.Timeout, tpool.execute, raise_timeout)
 
     @tests.skip_with_pyevent
-    def test_tpool_set_num_threads(self):
+    @tests.mock.patch('eventlet.tpool._spawn_threads')
+    def test_tpool_set_num_threads(self, spawn_mock):
         tpool.set_num_threads(5)
         self.assertEqual(5, tpool._nthreads)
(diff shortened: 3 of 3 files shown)


eventlet/tpool.py
```diff
@@ -333,4 +338,12 @@ def killall():
 
 def set_num_threads(nthreads):
     global _nthreads
-    _nthreads = nthreads
+    old_nthreads, _nthreads = _nthreads, nthreads
+
+    if _setup_already:
+        if old_nthreads > nthreads:
+            import warnings
```
temoto: import inside function is a test for luck, please never do it again.

eventlet/tpool.py
```diff
@@ -333,4 +338,12 @@ def killall():
...
+            warnings.warn("Dynamically reducing the number of threads is not "
```
temoto: Please remove this warning.

eventlet/tpool.py
```diff
@@ -282,18 +282,23 @@ def setup():
...
 
     _coro = greenthread.spawn_n(tpool_trampoline)
     # This yield fixes subtle error with GreenSocket.__del__
     eventlet.sleep(0)
 
 
+def _spawn_threads():
+    for i in six.moves.range(len(_threads), _nthreads):
+        t = threading.Thread(target=tworker,
+                             name="tpool_thread_%s" % (i + 1))
```
temoto: Please make it `t = threading.Thread(target=tworker, name='tpool_thread_{0}'.format(i + 1))`
I know it is not related to work, but since you touched this line, might as well remove the bad and ugly formatting.

eventlet/tpool.py
Akrog: I'll more than gladly change this, as I only did it like this in the first place because I always try to follow the coding style and patterns used in the project I'm contributing to, even when they don't match my own, and in this case we have this same pattern of importing "warnings" in the "setup" method.

eventlet/tpool.py
Akrog: That doesn't sound quite right, if caller makes a request thinking it will dynamically changed and we are not going to fulfill it, shouldn't we issue a warning?  (I see that I forgot to pass the "RuntimeWarning" parameter).

eventlet/tpool.py
Akrog: Done

eventlet/tpool.py
temoto: Yes, breaking expected behavior is not right, I agree. Here's why I think there should be no warning. If you have little time, just skip to solutions. :-)

First, and most important, warning still doesn't deliver expected behavior. Second, in this particular case, I believe, warning is unlikely to find a human that cares or will change something. Alternatives down below.

There will be two kinds of human readers of warning:
- user of eventlet, writer of openstack or other software. It is target audience for warning indeed. Though blunt and plain `ValueError('must increase only')` would be far more effective here.
- final user of software. It may not be noticed or mildly annoy user who can't fix "unsupported reducing of what is it even talking about" (constructive criticism below). He might look for a way to disable warning and find a stupid ugly recipe to disable all warnings. This is bad and unfortunately, happens more often than I'd like to think. Python is next PHP I hope it conveys my point without hurting feelings. Coders (as opposed to software engineers) are the workhorse of IT industry and caring for them means making libraries that just work in the best possible way, without much distraction.

### Solution options

Basically, either fulfill expectation or maintain warning silence for really important things. :-)
- (best) actually implement pool scaling down to requested size. I can take this task.
- (acceptable) just remove warning and assume grow-only is acceptable (for a while) until scaling down is not implemented. Really, the way **global** tpool is used, grow-only can hardly make harm.
- (worst) rename `set_num_threads` to `increase_pool_size` with deprecation warning on former ("301 redirect" if HTTP is your second language). This removes wrong expectation.

FYI, don't go with this in this case, but good stuff in general. Separate warning class to ease disabling. You have unlimited room for description text. Don't assume reader knows what is thread pool. Explain how serious is the problem and what could be done. Link to case history. Example: https://github.com/eventlet/eventlet/blob/f1b63abd6db186c978077499f9670600da599d1a/eventlet/convenience.py#L53-L76

Akrog:
Thanks for the thorough comments.

I had looked into supporting the scaling down of the pool, but for some reasons all the implementations I could think of didn't look quite right, maybe I'm missing the trivial and elegant solution.

Below I have briefly explained the solutions I had for scaling down, what are your thoughts?

Solution 1:
--------------

Do `_reqq.put(None)` for as many threads as we have to remove and make `tworker` remove itself from the `_threads` list, since the alternative is to keep polling in `set_num_threads` all threads for their status to see which one finishes.

This has the problem that you can't control which threads are going to stop, so you may end up with spaces between the thread names, for example going from 3 threads to 2 you could end up with only tpool_thread_1 and tpool_thread_3 if tpool_thread_2 is the one that stops.

This is problematic if you later on scale up the number of threads since you'll need to name the new threads based on the highest numbered one and you'll leave spaces in between, or you'll have to look into the names of existing threads and fill in the gaps with new threads.

This adds way too much logic to the code to do something that should be simple.

Solution 2:
--------------

Stop all threads and then start as many as you need.  This sounds kind of "unclean" to me.

Solution 3:
--------------

Make tworker check `_nthreads` after each execution and see if they should stop executing based on its own position on the `_threads` list, and make the `set_num_threads` do a `.join()` on the threads that will stop.

This is the one I like best, but I'm not too happy about it because we are adding more logic to the `tworker` method that will be executed on each work request just for the rare occasion where we scale down.

temoto:
> Make tworker check _nthreads after each execution and see if they should
It's an elegant idea. You'd have to rewrite all tpool to make this "tworker actor" model consistent. You'd also need to put some bogus value into request queue to wake tworkers.

- (the most important) addresses polling and join: scaling is not an operation, it's a long process. This fact must be exposed in interface. I think `set_num_threads` must be non-blocking, because name doesn't suggest that. Then we can add another method, say `scale(block=True)` so caller is aware that he is waiting for sufficient number of thread workers to finish, which could take seconds or years.
- confusing names are solved easily. `Thread(..., name='eventlet_tpool_{pool_name}')`, currently the only `pool_name` is 'global'. There is no interface to supply work to particular thread, so different names aren't useful as well. There is `.ident` attribute to obtain really unique identifier.
- tworker did not add itself to _threads, and so it should not remove.
- `_reqq.put(None)` is existing and tested way to finish random thread. **If** synchronization on thread-stop needed, it must be added as separate, useful by itself feature, and then whoever wanted to wait for scale-down to finish would use it.

itamarst:
Unfortunately we are no longer accepting new features to eventlet, as we are switching to a maintenance focus. The hope is that OpenStack will be able to migrate off of eventlet.

Issue #480, opened 2018-04-18
eventlet import_patched doesn't honor PEP 302 import hooks


nat-goodspeed:
This is part of (e.g.) https://github.com/miguelgrinberg/python-socketio/issues/35#issuecomment-373960551 : [PyInstaller](https://www.pyinstaller.org/) has trouble with applications importing eventlet.

I know of two different problems.

One is that, in one-file mode, PyInstaller's boot loader uses [PEP 302](https://www.python.org/dev/peps/pep-0302/) import hooks to intercept dynamic imports. Its hook extracts the desired module from the packed file into a temp directory indicated in `sys._MEIPASS` and then loads it from there. But since the `__import__()` calls in `patcher.py` bypass import hooks, the boot loader doesn't extract modules (attempted to be) imported by `import_patched()`. They're not present in the temp directory.

The other, as indicated in the python-socketio issue comment linked above, is that in the PyInstaller one-file frozen environment, `os.path.dirname(__file__)` indicates a different directory than `sys._MEIPASS`.

In https://github.com/eventlet/eventlet/issues/7, [flub](https://github.com/eventlet/eventlet/issues/7#issuecomment-19206475) and [graingert](https://github.com/eventlet/eventlet/issues/7#issuecomment-19213338) suggest the idea of using `importlib.import_module()` instead of plain `__import__()`, potentially engaging import hooks. I will try to fix both problems noted above. I will submit a pull request referencing this github issue, but I wanted to create the issue first so I can reference it in the pull request.

Naturally, tinkering with the `import_patched()` machinery is likely to require some tweaks before all tests pass. Please bear with me during this process.

temoto:
Thanks for taking on this hard problem.

> but I wanted to create the issue first so I can reference it in the pull request.

Pull request concept is an issue enough over simple patch exchange. Feel free to skip any intermediate bullshit next time.

nat-goodspeed:
Okay, thanks. Of course I want to Do It Right.

I note that one useful side effect of creating an issue is that github cross-referenced that issue in the two that I mentioned above.

nat-goodspeed:
Evidently my premise was mistaken: __import__() does honor PEP 302 import hooks. This issue is a non-issue.

nat-goodspeed:
class DumbFounder(object):
        def find_module(self, fullname, path=None):
            print("find_module({}, {})".format(fullname, path))
            return None
    sys.meta_path.append(DumbFounder())
    __import__('eventlet')

[...]

PR #481, opened 2018-04-18
#480: make eventlet import_patched() compatible with PEP 302 import hooks

nat-goodspeed:


eventlet/patcher.py
```diff
@@ -1,5 +1,6 @@
 import imp
 import sys
+from importlib import import_module
```
temoto: Only module imports are fine. `from module import function` is a recipe for trouble and is prohibited unless absolutely necessary.

temoto:
Do you have an idea how to test this?

eventlet/patcher.py
nat-goodspeed: Sorry, will fix that importlib import and usage.

I'm okay with a more sweeping change to use importlib. My thought was to minimize risk. However, I do realize that unless all `__import__()` calls are changed, one application may work in a frozen environment while another, that uses more of eventlet, may not.

nat-goodspeed:
> Do you have an idea how to test this?

I have a specific use case -- an application using eventlet that I need to freeze to PyInstaller one-file mode -- that will serve to test whether import hooks are in fact engaged by eventlet-mediated imports. As packaged by PyInstaller, it does not currently work. I hope to make it work.

I suspect you mean: how should we enhance the eventlet test suite? I envision a test script that provides import hooks with observable side effects, invokes `monkey_patch()` and `import_patched()` and checks for the expected side effects.

I trust in the existing eventlet test suite to verify previous behavior.

temoto:
Thanks.

nat-goodspeed:
Hmm. I see this CI test failure:

    FAIL: zmq eats CPU after PUB send or DEALER recv.
      File "/home/travis/build/eventlet/eventlet/tests/__init__.py", line 96, in wrapped
        return func(*a, **kw)
      File "/home/travis/build/eventlet/eventlet/tests/zmq_test.py", line 464, in test_cpu_usage_after_pub_send_or_dealer_recv
        tests.check_idle_cpu_usage(0.2, 0.1)
      File "/home/travis/build/eventlet/eventlet/tests/__init__.py", line 241, in check_idle_cpu_usage
        allowed_part * 100)
    AssertionError: CPU usage over limit: user 12% sys 0% allowed 12%

but since there isn't an import function in the traceback, it's not obvious to me that the fail is due to my PR. Pushing forward.

temoto:
It's not related.

nat-goodspeed:
> if it's using importlib anyway, maybe there is no need to change it?

The original code uses `__import__()` and steps through the components of the dotted name, calling `getattr()` to drill down to the innermost module (as returned by `importlib.import_module()`) instead of the outermost (as returned by `__import__()`).

My change https://github.com/eventlet/eventlet/pull/481/commits/6ea095a9cb65e9b917cf973af3df927246e9d036 replaced `__import__()` within the helper function with `importlib.import_module()`, without recognizing the role of the helper function. This failed: `importlib.import_module()` returned the innermost module, and the local `import_module()` function then tried to call `getattr()` on that innermost module to dereference the second-level name.

> * submit patch to rthalley/dnspython then update version in `bin/pull-dnspython`
> * store this change as patch in eventlet tree and update `bin/pull-dnspython` to apply patch (it has example)

[That change](https://github.com/eventlet/eventlet/pull/481/commits/6ea095a9cb65e9b917cf973af3df927246e9d036#diff-87b6708a2a84cf5bb3d64a89e4320f58) is the only place I touched `eventlet/support/dns`. I think I would like to pursue the eventlet patch approach and concurrently try a pull request to `dnspython`. Never having engaged with that community, I would like to avoid adding `dnspython` review and acceptance to my critical path. Is it reasonable to set up the eventlet patch in such a way that should the `dnspython` pull request be accepted, failure of the eventlet patch won't fail the whole process?

The other point worth noting is that I also [touched eventlet/support/six.py](https://github.com/eventlet/eventlet/pull/481/commits/6ea095a9cb65e9b917cf973af3df927246e9d036#diff-96c143dfd9796d7ba81c5e0777c93ce5) -- must that change be recast in a similar way?

temoto:
> Is it reasonable to set up the eventlet patch in such a way that should the dnspython pull request be accepted, failure of the eventlet patch won't fail the whole process?

`bin/pull-*` are only called manually so any failure is isolated to attempt to update vendored packages. So there is no patch failure really. If/when patch is accepted upstream, we should just remove it here and run `bin/pull-dnspython`.

> The other point worth noting is that I also touched eventlet/support/six.py -- must that change be recast in a similar way?

Six is also vendored. I'm sorry, didn't look at your `__import__` commit. And I'm sorry again for double standards, but I believe six is kinda special and better left as upstream. I have more faith in six doing right thing because it's used in virtually every python package these days. Maybe six already switched to importlib? I only updated vendored packages when people find problem with current version.

nat-goodspeed:
> Maybe six already switched to importlib?

It seems not: https://github.com/benjaminp/six/blob/master/six.py#L80

So okay: https://github.com/benjaminp/six/pull/239

nat-goodspeed:
With the CI tests on benjaminp/six#239 running green, should I back out my change to eventlet/support/six.py?

temoto:
@nat-goodspeed please keep support/six same as upstream.

nat-goodspeed:
Commit c62d490 reconciles the code I touched in `eventlet/support/six.py` with benjaminp/six#239.

But did you mean I should also update `eventlet/support/six.py` to the current state of `benjaminp/six/six.py`, including changes unrelated to the `import_modules()` issue?

nat-goodspeed:
See also https://github.com/rthalley/dnspython/pull/304

nat-goodspeed:
Happy to note that rthalley/dnspython#304 CI tests pass.

temoto:
@nat-goodspeed I'm sorry for confusing. I should've automated it better. Here's another way to put it: treat `eventlet/support/{dns,monotonic,six}` as auto-generated files, ad-hoc changes in eventlet tree will be lost. To preserve changes: either point pull- script to new upstream version or create patch file.

For six, it's only one file so I copy it from upstream without pull- script. And no diverging via patches.

Your options regarding six:
- wait for benjaminp/six PR merge and then update vendored six
- don't wait and restore vendored six to 1.10

No matter what, `eventlet/support/six.py` must match some version of `benjaminp/six/six.py` in their master branch. I will probably assert this in travis build in near future.

nat-goodspeed:
Sorry for misunderstanding!

To verify that I now understand correctly: any changes I would like to make to `eventlet/support/six.py` *must* be made via pull request to benjaminp/six. Only when benjaminp/six's master branch includes those changes can its `six.py` be updated into `eventlet/support/six.py`. Yes?

As noted earlier in the conversation, I have open pull requests against both six and dnspython. I will work up an eventlet patch for dnspython; my PyInstaller-frozen application definitely fails to load `dns` modules from `eventlet.support`.

I have not (yet) observed such a failure in connection with `six`. My edits there are in response to your remark that it seems right to switch to importlib completely. I will commit a change reverting my PR's `eventlet/support/six.py` to match the master branch.

nat-goodspeed:
Shouldn't [this line](https://github.com/eventlet/eventlet/blob/master/bin/pull-dnspython#L4):

    version=${1-22e9de1d7957e558ea8f89f24e402cbbc8d50646}

be:

    version=${1:-22e9de1d7957e558ea8f89f24e402cbbc8d50646}

temoto:
> Only when benjaminp/six's master branch includes those changes can its six.py be updated into eventlet/support/six.py. Yes?

Yes.

temoto:
On `version=${1-...}`: bash manual indeed says it must be `:-` but it also works as just minus, maybe one of thousand shell compatibility things. Thank you for education.

nat-goodspeed:
Good, all the `import_module()` replacements seem to be working now!

nat-goodspeed:
Sorry, I still haven't added the tests we discussed. I explicitly state when I think this is ready to pull.

Thank you!

nat-goodspeed:
I am embarrassed to admit that my premise was mistaken: `__import__()` does honor PEP 302 import hooks. Therefore this PR is unnecessary. Very sorry for noise!!

That being the case, I'm still trying to diagnose why my PyInstaller-packaged application that imports eventlet fails to load. I still hope to resolve that; the solution may involve another pull request. But this PR was unfortunately based on a bad diagnosis.

nat-goodspeed:
https://github.com/eventlet/eventlet/pull/482 is much of the solution to PyInstaller + eventlet.

PR #482, opened 2018-04-25
Use relative imports for nested `dns` package. 

nat-goodspeed:
In a frozen environment, the runtime bootstrap loader may not properly handle
absolute imports of (e.g.) `dns.exception` when there is no top-level `dns`
package. Relative imports work, e.g.:

    from . import exception

But you can't say:

    import ..dns.exception

You can't even say:

    from .. import dns

because that doesn't work when `dns` is found via `sys.path`.

Since the referencing code consistently uses fully qualified names such as
`dns.exception.SyntaxError`, we could write:

    from . import exception as dns_exception

and change every reference to (e.g.) `dns_exception.SyntaxError`.

It was deemed less intrusive to introduce a dummy class as a namespace, e.g.:

    class dns(object):
        from . import exception

which allows references of the form `dns.exception.SyntaxError` to work as is.

Using relative imports means that `eventlet.support.greendns` need not tinker
with `sys.path`.

Also encapsulate some of the redundancy in `greendns.py`'s `import_patched()`
stanza in a transient `patch_imports()` helper function. Use `try`/`finally`
to ensure `patch_imports()` and the local `import_patched()` are transient.

Also attempt to circumvent #413 by catching (e.g.)
`eventlet.support.dns.exception.SyntaxError`. It's not clear that this
addresses the problem, however.

That being the case, I've included [the hosts-file parsing fix](https://github.com/eventlet/eventlet/commit/2798007c93c957601d280de16908f98dc97ebd5b) to eliminate a `SyntaxError` exception that was a showstopper in my runtime environment. I could break that out as a separate pull request if you want, though.


nat-goodspeed:
The other part of the PyInstaller puzzle is to inform PyInstaller of the "hidden" imports performed by `eventlet.support.greendns`. That can be done with this PyInstaller hook file, which must be named `hook-eventlet.support.greendns.py`:

    # PyInstaller hook for the dnspython snapshot embedded in eventlet

    from PyInstaller.utils.hooks import collect_submodules
    hiddenimports = collect_submodules('eventlet.support.dns')

Until that becomes part of the PyInstaller library of predefined hooks, I pass PyInstaller the `--additional-hooks-dir=` switch pointing to the directory containing `hook-eventlet.support.greendns.py`.

nat-goodspeed:
Okay, more work to do.

nat-goodspeed:
Ha -- looks like success to me, since the only failure is

    FAIL: zmq eats CPU after PUB socket .bind()

But let's leave this PR pending until we decide about the external-dependencies notion.

temoto:
> Without yet having done that, I believe the hosts-file parsing fix, along with its associated commits, is worth keeping. https://github.com/eventlet/eventlet/pull/482/commits/2798007c93c957601d280de16908f98dc97ebd5b

If you have time, please write a test. If not, I'll merge it when I write a test. Thank you very much.

temoto:
External dependencies have their downsides but look better overall.

PR #487, opened 2018-05-01
Fix parsing of Windows hosts file. 

nat-goodspeed:
The default Windows hosts file starts something like this:

    # Copyright (c) 1993-2009 Microsoft Corp.
    #
    # This is a sample HOSTS file used by Microsoft TCP/IP for Windows.
    ...

But `greendns.HostsResolver.LINES_RE` doesn't admit the possibility of a comment starting at the beginning of a line. It produces a sequence consisting of data from the comment lines as well as (eventually) the real content of the hosts file.

This has been tolerated since `is_ipv4_addr()` and `is_ipv6_addr()` catch `dns.exception.SyntaxError` and return `False`. But if a syntactically-valid line is commented out, `HostsResolver` will fail to ignore it.

Changing '+' to '*' allows `HostsResolver._readlines()` to recognize and skip comment lines. However, that produces empty strings in the sequence. Filter out the empty strings by passing the sequence through `six.moves.filter(None)`.

Change tests to compare `list(_readlines(etc.))`.


nat-goodspeed:
This captures the commits that are still useful from #482.

[I've added a test](https://github.com/eventlet/eventlet/pull/487/commits/2537ba23b8981548ae0136a67d09ed26a13d7453) that demonstrates the problem.

nat-goodspeed:
I believe this one is ready.

temoto:
Thank you. I'm without computer hopefully only for today. Please wait.

nat-goodspeed:
No problem, you've been very patient with me!

tests/greendns_test.py
```diff
@@ -10,6 +10,7 @@
 from eventlet.support.greendns import dns
 import tests
 import tests.mock
+from nose.tools import assert_equal
```
temoto: No nose tools, please. `-d` flag makes up for any convenience built into `assert_equal`.

I can remove it if you're busy. Sorry for delay.

nat-goodspeed: > `-d` flag makes up for any convenience built into `assert_equal`

I don't understand this remark, but will introduce a local definition of `assert_equal()` instead of changing all those lines back to `assert A == B`. I do find it valuable for debugging that when they are *not* equal, `assert_equal()` displays the values.

It would be helpful if I could add a comment explaining why we define `assert_equal()` locally rather than using the one from `nose.tools`.

temoto:
```
nosetests --help
...
  -d, --detailed-errors, --failure-detail
                        Add detail to error output by attempting to evaluate
                        failed asserts [NOSE_DETAILED_ERRORS]
```

No nose tools external or internal please.

nat-goodspeed: Okay: I now believe you're saying that you *prefer* `assert A == B` because the nose `-d` switch provides the extra information we could have gotten from `assert_equal()`. If so, I completely misread your remark earlier. Will revert source changes accordingly.

temoto: Yeah and with pytest (some time in future) plain assert is even more helpful pinpointing outliers in data structures.

temoto:
Thanks, LGTM, merging after tests.

temoto:
Was it really specific to Windows?

nat-goodspeed:
No. It was on Windows that it caused me problems, but the bug itself was not platform-specific. Sorry for misleading title line.

Issue #479, opened 2018-03-28
eventlet import failure


abergou:
Hello,
eventlet seems to be causing various issues on our system in its interaction with dnspython. In particular there is bad interaction in that if dnspython is already imported then eventlet will try to patch the imported dnspython causing various imports to fail. Here is a short reproducer:

```
>>> import dns.rdtypes
>>> import eventlet
---------------------------------------------------------------------------
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "eventlet/__init__.py", line 10, in <module>
    from eventlet import convenience
  File "eventlet/convenience.py", line 7, in <module>
    from eventlet.green import socket
  File "eventlet/green/socket.py", line 21, in <module>
    from eventlet.support import greendns
  File "eventlet/support/greendns.py", line 73, in <module>
    setattr(dns.rdtypes.ANY, pkg, import_patched('dns.rdtypes.ANY.' + pkg))
  File "eventlet/support/greendns.py", line 61, in import_patched
    return patcher.import_patched(module_name, **modules)
  File "eventlet/patcher.py", line 120, in import_patched
    *additional_modules + tuple(kw_additional_modules.items()))
  File "eventlet/patcher.py", line 94, in inject
    module = __import__(module_name, {}, {}, module_name.split('.')[:-1])
  File "/Users/temoto/dev/github.com/eventlet/eventlet/venv-27/lib/python2.7/site-packages/dns/rdtypes/ANY/CDNSKEY.py", line 23, in <module>
    class CDNSKEY(dns.rdtypes.dnskeybase.DNSKEYBase):
AttributeError: 'module' object has no attribute 'dnskeybase'
```

The only workaround I have found so far is to import eventlet before anything else. This, however, is not ideal since we don't always have control over the import order of modules. Are there any fixes available?

Thanks!

abergou:
It seems like this may be related to both [issue#381](https://github.com/eventlet/eventlet/issues/381) and [issue#454](https://github.com/eventlet/eventlet/issues/454).

temoto:
Not cutting error information is a start to fixes.

abergou:
```python
>>> import dns.rdtypes
>>> import eventlet
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/eventlet/__init__.py", line 10, in <module>
    from eventlet import convenience
[...]

PR #490, opened 2018-05-13
Patch all modules in dns.rdtypes

bjmb:
Fixes https://github.com/eventlet/eventlet/issues/479


temoto:
Please, try again with latest code.

`pip install https://github.com/eventlet/eventlet/archive/53b63c66f0d9a7d6ba9182a207529ebd949e95f3.zip`

it uses external dnspython package and maybe will work without your patch.

bjmb:
Nice test, I tried but failed writing one that failed before my commit/passed after

PR #495, opened 2018-06-04
Fix blocking put to empty pool

smerritt:
If you have a pool with no free items, one greenthread blocked in
pool.get(), and then you call pool.put(item), sometimes the put will
block.

This happens when the greenthread blocked in pool.get() has a pending
timeout. The timeout's timer has fired, the call to throw() has been
scheduled, but throw() has not actually run yet. In pool.put(), we see
a waiting getter, so we do a blocking self.channel.put()... but when
the getter runs, it unwinds its stack and does not take the item,
leaving the caller of pool.put() blocked despite there being enough
free space.

This commit fixes that by (a) making LightQueue.put() and .get() work
with 0-length queues, even with timeouts, and (b) checking for
queue.Full in Pool.put() and handling it correctly.

### DIFF
--- a/eventlet/pools.py
+++ b/eventlet/pools.py
@@ -121,12 +121,16 @@ class Pool(object):
             return
 
         if self.waiting():
-            self.channel.put(item)
+            try:
+                self.channel.put(item, block=False)
+                return
+            except queue.Full:
+                pass
+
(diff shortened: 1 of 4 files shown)


temoto:
Thank you.

Issue #468, opened 2018-03-06
WSGI environ string should be latin-1 encoding


StoneMoe:
PEP-0333 encoding part: https://www.python.org/dev/peps/pep-0333/#unicode-issues

Patch pull request: https://github.com/eventlet/eventlet/pull/467

Known issue:
not compatible with werkzeug
```
Traceback (most recent call last):
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/eventlet/wsgi.py", line 539, in handle_one_response
    result = self.application(self.environ, start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/flask_socketio/__init__.py", line 43, in __call__
    start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/engineio/middleware.py", line 49, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/debug/__init__.py", line 468, in __call__
    request.path == self.console_path:
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/utils.py", line 73, in __get__
    value = self.func(obj)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/wrappers.py", line 596, in path
    self.charset, self.encoding_errors)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/_compat.py", line 176, in wsgi_decoding_dance
    return s.encode('latin1').decode(charset, errors)
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 11-12: ordinal not in range(256)

127.0.0.1 - - [06/Mar/2018 13:31:44] "GET /test/%E4%BD%A0%E5%A5%BD HTTP/1.0" 500 1603 0.013243
```

temoto:
Thank you.

StoneMoe:
CC @Xiami2012

temoto:
Fix is merged 4e576cbca07cd2c50da4baffd033a67c2ab10d1a

Thank you @StoneMoe

PR #467, opened 2018-03-06
wsgi string should be latin-1 encoding (here only PATH_INFO is fixed)

StoneMoe:
According to: https://www.python.org/dev/peps/pep-0333/#unicode-issues

This will fix a possible encoding issue in wsgi app, when percent encoding appeared in URL

Example:
```
Traceback (most recent call last):
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/eventlet/wsgi.py", line 539, in handle_one_response
    result = self.application(self.environ, start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/flask/app.py", line 1997, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/flask_socketio/__init__.py", line 43, in __call__
    start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/engineio/middleware.py", line 49, in __call__
    return self.wsgi_app(environ, start_response)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/debug/__init__.py", line 468, in __call__
    request.path == self.console_path:
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/utils.py", line 73, in __get__
    value = self.func(obj)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/wrappers.py", line 596, in path
    self.charset, self.encoding_errors)
  File "/home/stonemoe/Desktop/Code/test/venv/lib/python3.6/site-packages/werkzeug/_compat.py", line 176, in wsgi_decoding_dance
    return s.encode('latin1').decode(charset, errors)
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 11-12: ordinal not in range(256)

127.0.0.1 - - [06/Mar/2018 13:31:44] "GET /test/%E4%BD%A0%E5%A5%BD HTTP/1.0" 500 1603 0.013243
```

PR #497, opened 2018-06-07
wsgi: Use byte strings on py2 and unicode strings on py3

### LINKED ISSUE #488: #468 fix not pep 0333 compliant under cpython2
The fix for #468 made the wsgi environment string pep 0333 compliant under cpython3 by coercing those unicode strings to be latin1. In doing so they are now unicode strings under cpython2 as well. Reading pep 0333 only python implementations where "the str or StringType type is in fact Unicode-based" may use a unicode string type with this subset restriction. Otherwise "All encoding/decoding must be handled by the application; all strings passed to or from the server must be standard Python byte strings, not Unicode objects."

This likely wouldn't be an issue except that python2 urllib path quoting methods seem to only operate on byte strings and not unicode strings (the lookup tables don't have keys for unicode strings, only bytestrings). The end result is that the fix for #468 breaks python2 urllib methods on these environment values. (Example of this happening at https://gist.github.com/notmyname/43413494ba62b22ba3d704a495563b74)

Considering this original change was made to be pep 0333 compliant, encode_dance() should be updated to return bytes on cpython2.

tipabu:
Closes #488; for some additional context, see #468 and #467.

PEP-0333 says [1]

> all strings referred to in this specification **must** be of type ``str``

but muddied that message earlier with a bunch of talk about

> all strings passed to or from the server must be standard Python byte
> strings, not Unicode objects

PEP-3333 sought to clarify [2] with

> WSGI therefore defines two kinds of "string":
>
> * "Native" strings (which are always implemented using the type
>   named ``str``) that are used for request/response headers and
>   metadata
>
> * "Bytestrings" (which are implemented using the ``bytes`` type
>   in Python 3, and ``str`` elsewhere), that are used for the bodies
>   of requests and responses (e.g. POST/PUT input data and HTML page
>   outputs).
>
> Do not be confused however: even if Python's ``str`` type is actually
> Unicode "under the hood", the *content* of native strings must
> still be translatable to bytes via the Latin-1 encoding!

And later, in "Unicode Issues" [3], it adds

> For values referred to in this specification as "bytestrings"
> (i.e., values read from ``wsgi.input``, passed to ``write()``
> or yielded by the application), the value **must** be of type
> ``bytes`` under Python 3, and ``str`` in earlier versions of
> Python.

So the upshot seems to be

  - All request and response bodies must be bytes, regardless of Python
    version.
  - All headers and other WSGI environment keys need to be native
    strings; i.e. bytes strings on Python 2 and unicode strings on
    Python 3.

[1] https://www.python.org/dev/peps/pep-0333/#unicode-issues
[2] https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types
[3] https://www.python.org/dev/peps/pep-3333/#unicode-issues

### DIFF
--- a/tests/wsgi_test.py
+++ b/tests/wsgi_test.py
@@ -1423,8 +1423,13 @@ class TestHttpd(_TestBase):
         # in all WSGI environment strings application must observe either bytes in latin-1 (ISO-8859-1)
         # or unicode code points \u0000..\u00ff
         # wsgi_decoding_dance from Werkzeug to emulate concerned application
-        decoded = g[0].encode('latin1').decode('utf-8', 'replace')
-        assert decoded == u'/你好'
+        msg = 'Expected PATH_INFO to be a native string, not {0}'.format(type(g[0]))
+        assert isinstance(g[0], str), msg
+        if six.PY2:
+            assert g[0] == u'/你好'.encode('utf-8')
+        else:
+            decoded = g[0].encode('latin1').decode('utf-8', 'replace')
+            assert decoded == u'/你好'
 
     @tests.skip_if_no_ipv6
     def test_ipv6(self):
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -62,6 +62,8 @@ def addr_to_host_port(addr):
 def encode_dance(s):
     if not isinstance(s, bytes):
         s = s.encode('utf-8', 'replace')
+    if six.PY2:
+        return s
     return s.decode('latin1')
 
 


tipabu:
An alternative approach that would get rid of `encode_dance` entirely:

``` diff
diff --git a/eventlet/wsgi.py b/eventlet/wsgi.py
index 23a44fb..4040f69 100644
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -59,14 +59,6 @@ def addr_to_host_port(addr):
     return (host, port)
 
 
-def encode_dance(s):
-    if not isinstance(s, bytes):
-        s = s.encode('utf-8', 'replace')
-    if six.PY2:
-        return s
-    return s.decode('latin1')
-
-
 # Collections of error codes to compare against.  Not all attributes are set
 # on errno module on all platforms, so some are literals :(
 BAD_SOCK = set((errno.EBADF, 10053))
@@ -639,7 +631,10 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
 
         pq = self.path.split('?', 1)
         env['RAW_PATH_INFO'] = pq[0]
-        env['PATH_INFO'] = encode_dance(urllib.parse.unquote(pq[0]))
+        if six.PY2:
+            env['PATH_INFO'] = urllib.parse.unquote(pq[0])
+        else:
+            env['PATH_INFO'] = urllib.parse.unquote(pq[0], encoding='latin1')
         if len(pq) > 1:
             env['QUERY_STRING'] = pq[1]
 
```

Not sure which you'd prefer.

tests/wsgi_test.py
```diff
@@ -1423,8 +1423,13 @@ def wsgi_app(environ, start_response):
         # in all WSGI environment strings application must observe either bytes in latin-1 (ISO-8859-1)
         # or unicode code points \u0000..\u00ff
         # wsgi_decoding_dance from Werkzeug to emulate concerned application
-        decoded = g[0].encode('latin1').decode('utf-8', 'replace')
-        assert decoded == u'/你好'
+        assert isinstance(g[0], str), \
+            'Expected PATH_INFO to be a native string, not %r' % type(g[0])
```
temoto: `.format` please.

tipabu:
Anything else I can do to help move this forward?

temoto:
@tipabu sorry my time management sucks again. Merged.

tipabu:
No worries! Thanks for landing this -- makes it much easier to port a project to py3 when I can use a singular version of eventlet for both py2 and py3 🙂

PR #498, opened 2018-06-08
Always remove the right listener from the hub

smerritt:
When in hubs.trampoline(fd, ...), a greenthread registers itself as a
listener for fd, switches to the hub, and then calls
hub.remove(listener) to deregister itself. hub.remove(listener)
removes the primary listener. If the greenthread awoke because its fd
became ready, then it is the primary listener, and everything is
fine. However, if the greenthread was a secondary listener and awoke
because a Timeout fired then it would remove the primary and promote a
random secondary to primary.

This commit makes hub.remove(listener) check to make sure listener is
the primary, and if it's not, remove the listener from the
secondaries.

I found this when I saw a process running at 100% CPU and strace showed it calling poll() in a tight loop. There was a readable file descriptor, but the process never read it.

### DIFF
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -225,15 +225,23 @@ class BaseHub(object):
 
         fileno = listener.fileno
         evtype = listener.evtype
-        self.listeners[evtype].pop(fileno, None)
-        # migrate a secondary listener to be the primary listener
-        if fileno in self.secondaries[evtype]:
-            sec = self.secondaries[evtype].get(fileno, None)
-            if not sec:
-                return
-            self.listeners[evtype][fileno] = sec.pop(0)
-            if not sec:
-                del self.secondaries[evtype][fileno]
+        if listener is self.listeners[evtype].get(fileno):
+            self.listeners[evtype].pop(fileno, None)
+            # migrate a secondary listener to be the primary listener
+            if fileno in self.secondaries[evtype]:
+                sec = self.secondaries[evtype].get(fileno, None)
+                if not sec:
+                    return
+                self.listeners[evtype][fileno] = sec.pop(0)
+                if not sec:
+                    del self.secondaries[evtype][fileno]
+        else:
+            sec = [l for l in self.secondaries[evtype].get(fileno, [])
+                   if l is not listener]
+            if sec:
+                self.secondaries[evtype][fileno] = sec
+            else:
+                self.secondaries[evtype].pop(fileno, None)
 
     def mark_as_reopened(self, fileno):
         """ If a file descriptor is returned by the OS as the result of some
--- a/tests/hub_test.py
+++ b/tests/hub_test.py
@@ -1,11 +1,14 @@
 from __future__ import with_statement
+import errno
+import fcntl
+import os
 import sys
 import time
 
 import tests
 from tests import skip_with_pyevent, skip_if_no_itimer, skip_unless
 import eventlet
-from eventlet import hubs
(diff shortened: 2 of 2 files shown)


temoto:
Sorry for delay, I'm in road.

clayg:
I opened #645 if we wanted to move in the direction @charmster suggested - I'm confident this patch fixes a gaping hole for folks using multiple readers as written.  IMHO this should merge, and any additional improvements should be follow-ons.

eventlet/hubs/hub.py
```diff
@@ -224,15 +224,23 @@ def remove(self, listener):
...
-        # migrate a secondary listener to be the primary listener
-        if fileno in self.secondaries[evtype]:
-            sec = self.secondaries[evtype].get(fileno, None)
-            if not sec:
-                return
-            self.listeners[evtype][fileno] = sec.pop(0)
-            if not sec:
-                del self.secondaries[evtype][fileno]
+        if listener is self.listeners[evtype].get(fileno):
+            self.listeners[evtype].pop(fileno, None)
```
tipabu, line 228: Should be safe to just
```
del self.listeners[evtype][fileno]
```
but either way; functionally equivalent.

eventlet/hubs/hub.py
```diff
@@ -224,15 +224,23 @@ def remove(self, listener):
...
+            # migrate a secondary listener to be the primary listener
+            if fileno in self.secondaries[evtype]:
+                sec = self.secondaries[evtype].get(fileno, None)
+                if not sec:
+                    return
```
tipabu, line 233: This was pre-existing, but I wonder if we should
```
del self.secondaries[evtype][fileno]
```
before returning...

temoto:
@smerritt now that core of this patch is merged, please decide whether you want to rebase or close this.

eventlet/hubs/hub.py
```diff
@@ -224,15 +224,23 @@ def remove(self, listener):
...
+                self.listeners[evtype][fileno] = sec.pop(0)
+                if not sec:
+                    del self.secondaries[evtype][fileno]
+        else:
+            sec = [l for l in self.secondaries[evtype].get(fileno, [])
+                   if l is not listener]
+            if sec:
+                self.secondaries[evtype][fileno] = sec
+            else:
+                self.secondaries[evtype].pop(fileno, None)
```
tipabu, line 243: This stuff is the key fix -- if we're not the primary listener, *don't touch `self.listeners`*! Love it!

tipabu (approved):

smerritt:
#645 looks good to me, thanks.

eventlet/hubs/hub.py
temoto: @tipabu hey Tim do you think #645 that was merged is in good shape regarding this comment? Thanks.

tipabu: Yeah, absolutely -- it's even cleaner, down to just
```
            self.secondaries[evtype][fileno].remove(listener)
            if not self.secondaries[evtype][fileno]:
                del self.secondaries[evtype][fileno]
```
Clay even addressed my other thoughts while he was making that `remove(listener)` still pass tests. Kudos!

Sorry for the extra noise after #645 was approved; I'd already started looking at this and didn't notice everything was already taken care of before I left the review.

temoto: It's not noise, but a signal (yes) that there are multiple people so well familiar with hub code. That is awesome.

Issue #466, opened 2018-02-27
DNS exception: AttributeError: module 'eventlet.hubs' has no attribute 'epolls'


me21:
I get the following exception in my application:

```
  File "C:/Users/z/Documents/Python Projects/waves\engine.py", line 190, in <lambda>
    for orderBook in executor.map(lambda assetPair: [assetPair, orderBooks.get(assetPair) or WavesDexEngine.__loadOrderBook(assetPair)],
  File "C:/Users/z/Documents/Python Projects/waves\utils.py", line 23, in inner
    with eventlet.Timeout(DATA_LOAD_TIMEOUT_SECONDS):
  File "C:\Users\z\Documents\Python Projects\waves\venv\lib\site-packages\eventlet\timeout.py", line 55, in __init__
    self.start()
  File "C:\Users\z\Documents\Python Projects\waves\venv\lib\site-packages\eventlet\timeout.py", line 66, in start
    self.timer = get_hub().schedule_call_global(
  File "C:\Users\z\Documents\Python Projects\waves\venv\lib\site-packages\eventlet\hubs\__init__.py", line 116, in get_hub
    use_hub()
  File "C:\Users\z\Documents\Python Projects\waves\venv\lib\site-packages\eventlet\hubs\__init__.py", line 70, in use_hub
    mod = get_default_hub()
  File "C:\Users\z\Documents\Python Projects\waves\venv\lib\site-packages\eventlet\hubs\__init__.py", line 39, in get_default_hub
    return eventlet.hubs.epolls
AttributeError: module 'eventlet.hubs' has no attribute 'epolls'
```

I saw in debugger that `get_default_hub` is called several times, succeeding first and selecting `eventlet.hubs.selects` as default hub, then on the 3rd or 4th invocation it throws that exception.
Setting EVENTLET_HUB environment variable to `eventlet.hubs.selects` fixes this.

temoto:
@me21 thank you.

Please post output of this patch if you can
```diff
    select = patcher.original('select')
    try:
        import eventlet.hubs.epolls
+       print('!!!!!', vars(eventlet.hubs))
        return eventlet.hubs.epolls
```

me21:
https://gist.github.com/me21/1cf99f77abee9ed8bb9107f5cb16474a

This is on Windows 10, Python 3.6.3 x64. I must say this happens during sending many requests with `concurrent.futures.ThreadPoolExecutor`. So there are multiple imports in different threads.
[...]

PR #500, opened 2018-06-25
hubs: use threading.Lock in get_default_hub() (issue #466)

ivknv:
Fixes issue #466

### DIFF
--- a/eventlet/hubs/__init__.py
+++ b/eventlet/hubs/__init__.py
@@ -10,6 +10,8 @@ __all__ = ["use_hub", "get_hub", "get_default_hub", "trampoline"]
 threading = patcher.original('threading')
 _threadlocal = threading.local()
 
+_get_default_hub_lock = threading.Lock()
+
 
 def get_default_hub():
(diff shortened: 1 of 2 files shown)


temoto:
@ivknv I assume you can reproduce original issue, please try recent master, supposedly it fixes the issue in another way. Please don't kill me for rival patch. :-)

`pip install https://github.com/eventlet/eventlet/archive/77bccbe48d4d9a46982b2e0503e76784e76b066a.zip`

ivknv:
That patch appears to solve the problem as well, tested it the same way:

```python
import threading
import eventlet

def test_attribute_error():
    # See https://github.com/eventlet/eventlet/issues/466
    threads = [threading.Thread(target=eventlet.hubs.get_default_hub) for i in range(10)]

    for thread in threads:
        thread.start()

    for thread in threads:
        thread.join()
```

Ran it a couple of times on Windows &#8212; no errors.

temoto:
Thank you.

Issue #499, opened 2018-06-22
import eventlet breaks dns lookups using tcp


rgbyrnes:
Here's a simple reproducer.  It works if we don't import eventlet, or if we omit tcp=True.

```
Python 2.7.15 (default, May  9 2018, 14:07:52)
[GCC 7.3.1 20180303 (Red Hat 7.3.1-5)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import dns.version
>>> dns.version.version
'1.15.0'
>>> from dns.resolver import query
>>> import eventlet
>>> eventlet.__version__
'0.23.0'
>>> query('github.com', 'a', tcp=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/dns/resolver.py", line 1132, in query
    raise_on_no_answer, source_port)
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/dns/resolver.py", line 957, in query
    source_port=source_port)
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/eventlet/support/greendns.py", line 803, in tcp
    r = dns.message.from_wire(wire, keyring=q.keyring, request_mac=q.mac)
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/dns/message.py", line 806, in from_wire
    ignore_trailing)
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/dns/message.py", line 599, in __init__
    self.wire = dns.wiredata.maybe_wrap(wire)
  File "/usr/local/python/python-2.7/std/lib/python2.7/site-packages/dns/wiredata.py", line 103, in maybe_wrap
    raise ValueError("unhandled type %s" % type(wire))
ValueError: unhandled type <type 'bytearray'>
```

temoto:
@rgbyrnes please, try latest master with patch from @beltran 

`pip install https://github.com/eventlet/eventlet/archive/master.zip`

PR #501, opened 2018-06-26
wire should be bytes or string

beltran:
Fixes https://github.com/eventlet/eventlet/issues/499

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -26,7 +26,7 @@ import eventlet
 from eventlet import tpool
 import six
 import socket
-from threading import Thread, Event
+from threading import Thread
 import struct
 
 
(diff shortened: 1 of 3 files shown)


tests/greendns_test.py
```diff
@@ -854,3 +854,15 @@ def test_hosts_priority():
 def test_import_rdtypes_then_eventlet():
     # https://github.com/eventlet/eventlet/issues/479
     tests.run_isolated('greendns_import_rdtypes_then_eventlet.py')
+
+
+def test_nonameserver_is_risen():
```
temoto: These apply to all code:

- test is named `nonameserver_is_risen` but it doesn't check for any exception
- imports inside function is a recipe for disaster, use it only as last resort
- `reset_default_resolver()` modifies global state, so it renders all other tests results unreliable

This is eventlet project specific, some don't mind: don't rely on external services. Maybe you can't access 8.8.8.8, or this address is not what you expect or it doesn't know about `github.com`.

Solutions:
- for naming - name should be consistent with code
- for imports and global state - isolated tests, see code above
- for external dependencies - mock network IO or run small DNS server that's programmed to respond what you want

temoto:
Thank you for valuable contribution. You can fix tests or do nothing, I will write a test.

beltran:
I will fix the test, thank you for the detailed review and sorry for the clumsy test

tests/greendns_test.py
```diff
@@ -854,3 +857,28 @@ def test_hosts_priority():
 def test_import_rdtypes_then_eventlet():
     # https://github.com/eventlet/eventlet/issues/479
     tests.run_isolated('greendns_import_rdtypes_then_eventlet.py')
+
+
+def test_raise_dns_tcp():
+    # https://github.com/eventlet/eventlet/issues/499
+    try:
+        dns.resolver.query('host.example.com', 'a', tcp=True)
+        assert False, "Should have arisen NXDOMAIN or NoNameservers"
```
temoto: Test does not ensure that `host.example.com` query to unknown default resolver results in NXDOMAIN.

tests/__init__.py
```diff
@@ -406,3 +409,88 @@ def test_run_python_pythonpath_extend():
...
+                self.data = client_sock.recv(1024)
+                if bytearray(self.data) == b"DEATH_PILL":
+                    client_sock.close()
+                    break
+
+                self.init_domain()
+                client_sock.send(self.answer())
+                client_sock.close()
+        except socket.error:
+            pass
```
temoto: Ignoring errors is always a bad idea.

tests/greendns_test.py
```diff
@@ -854,3 +857,28 @@ def test_hosts_priority():
...
+    try:
+        tiny_server = tests.TinyDNSServerTCP("192.168.1.1")
+        tiny_server.start()
+        resolver.nameserver_ports["127.0.0.1"] = tiny_server.get_port()
+        response = resolver.query('host.example.com', 'a', tcp=True)
+
+        assert isinstance(response, Answer)
+    finally:
+        tiny_server.stop()
+        pass
```
temoto: (minor) redundant `pass`

tests/greendns_test.py
```diff
@@ -854,3 +857,28 @@ def test_hosts_priority():
...
+def test_raise_dns_tcp():
+    # https://github.com/eventlet/eventlet/issues/499
+    try:
+        dns.resolver.query('host.example.com', 'a', tcp=True)
+        assert False, "Should have arisen NXDOMAIN or NoNameservers"
+    except (dns.resolver.NXDOMAIN, dns.resolver.NoNameservers):
+        pass
+
+
+def test_noraise_dns_tcp():
```
temoto: - `TinyDNSServerTCP` listens `localhost` (and there is no way to change it and that's fine)
- test goes to `127.0.0.1` which is not the same as localhost
- `127...` string duplicate
- `192.168.1.1` is not used
- response is not checked against `192...`

temoto:
On DNS server design. I understand it works, feel free to leave it as is.

Here's a simpler version:
```
@contextlib.contextmanager
def dns_tcp_server(request_count=1):
  response = random_ip()
  state = [0]  # request count storage writable by thread

  def check(x):
    if isinstance(x, dns.resolver.Answer):
      assert response in x
    else:
      assert x == response

  def thread_fun():  # thread target
    while True:
      accept, recv, sendall
      # note socket.send is not enough https://docs.python.org/2/library/socket.html#socket.socket.send
      state[0] += 1
      if state[0] >= request_count:
        return

  sock = create, bind, listen
  server_addr = sock.getsocketname()
  thread = create, start
  yield server_addr, check
  thread.join()  # forever, all tests run with some form of timeout control

def test_dns_tcp():
    # https://github.com/eventlet/eventlet/issues/499
    with dns_tcp_server() as s:
        dnsaddr, check = s
        resolver = Resolver()
        resolver.nameservers = [dnsaddr[0]]
        resolver.nameserver_ports[dnsaddr[0]] = dnsaddr[1]
        response = resolver.query('host.example.com', 'a', tcp=True)
        check(response)
```

How is this better?
- no classes, instant win :-)
- no Event synchronisation, there was no reason to delay thread start and socket listen, so port is available at creation
- you may forget to do `server.stop()` and it still works
- check value of randomized response
- address of DNS server is not defined in test

beltran:
Thank you very much for the review @temoto, I think I've addressed all your comments.

temoto:
The purpose of request_count was to replace death_pill "end cleaner" with strict contract: test specifies how many requests it intends to send, receiving more or less is an error in eventlet greendns middleware or in test code.

But I'm not going to test your patience. :-) Thank you very much for your work.

beltran:
Ohh I see, I thought it was throttling the max number of connections. I could've put up with that :muscle:

Issue #502, opened 2018-07-04
Python 3.7 test failures


hroncok:
After https://github.com/eventlet/eventlet/issues/475 is fixed on greenlet side, there are still test failures on python 3.7:

```
======================================================================
ERROR: test_connect_ssl (tests.api_test.TestApi)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/eventlet-0.23.0/tests/api_test.py", line 67, in test_connect_ssl
    server_side=True
  File "/builddir/build/BUILD/eventlet-0.23.0/eventlet/convenience.py", line 158, in wrap_ssl
    return wrap_ssl_impl(sock, *a, **kw)
  File "/builddir/build/BUILD/eventlet-0.23.0/eventlet/green/ssl.py", line 379, in wrap_socket
    return GreenSSLSocket(sock, *a, **kw)
  File "/builddir/build/BUILD/eventlet-0.23.0/eventlet/green/ssl.py", line 68, in __init__
    ca_certs, do_handshake_on_connect and six.PY2, *args, **kw)
  File "/usr/lib64/python3.7/ssl.py", line 788, in __init__
    f"{self.__class__.__name__} does not have a public "
TypeError: GreenSSLSocket does not have a public constructor. Instances are returned by SSLContext.wrap_socket().

======================================================================
ERROR: test_wrap_ssl (tests.convenience_test.TestServe)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/eventlet-0.23.0/tests/convenience_test.py", line 121, in test_wrap_ssl
    server_side=True)
  File "/builddir/build/BUILD/eventlet-0.23.0/eventlet/convenience.py", line 158, in wrap_ssl
    return wrap_ssl_impl(sock, *a, **kw)
  File "/builddir/build/BUILD/eventlet-0.23.0/eventlet/green/ssl.py", line 379, in wrap_socket
    return GreenSSLSocket(sock, *a, **kw)
  File "/builddir/build/BUILD/eventlet-0.23.0/eventlet/green/ssl.py", line 68, in __init__
    ca_certs, do_handshake_on_connect and six.PY2, *args, **kw)
  File "/usr/lib64/python3.7/ssl.py", line 788, in __init__
    f"{self.__class__.__name__} does not have a public "
TypeError: GreenSSLSocket does not have a public constructor. Instances are returned by SSLContext.wrap_socket().

======================================================================
[...]

PR #503, opened 2018-07-04
[WIP] A naive attempt to fix Python 3.7 test failures

hroncok:
SSLSocket no longer has a public constructor.
Instances are returned by SSLContext.wrap_socket().

See https://github.com/eventlet/eventlet/issues/502

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -366,7 +366,7 @@ SSLSocket = GreenSSLSocket
 
 
 def wrap_socket(sock, *a, **kw):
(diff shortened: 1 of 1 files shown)


hroncok:
Travis is testing this on older Pythons, I'm running 3.7 build in Fedora infrastructure to see if it helps.

hroncok:
This breaks with:

    AttributeError: type object 'GreenSSLSocket' has no attribute 'wrap_socket'

hroncok:
OK, I've been trying to get this fixed but it's failure after failure. I give up (at least for now).

PR #504, opened 2018-07-05
wsgi: Don't strip all Unicode whitespace from headers on py3

tipabu:
An application that wants to accept UTF-8 header values on Python 2 can do so fairly easily: since str is bytes, it can decode from UTF-8 directly and return appropriate errors if that decoding fails.

On Python 3, it gets a little more interesting. Since str is unicode, the bytes-on-the-wire are decoded as Latin-1 before being put in the WSGI environment, so the application must encode as Latin-1 then decode as UTF-8. For the most part, this Just Works as the transformation should be lossless.

Sometimes, however, headers get truncated on py3. This is the result of a difference in how `.strip()` behaves for bytes vs unicode — in particular, there are some unicode characters that are considered whitespace whose Latin-1 encoded byte string is not:

    >>> [x for i in range(256) for x in (chr(i),)
    ...  if x.isspace() and not x.encode('latin1').isspace()]
    ['\x1c', '\x1d', '\x1e', '\x1f', '\x85', '\xa0']

Looking at RFC 7230, it defines header fields as

>   a case-insensitive field name followed by a colon (":"), optional leading whitespace, the field value, and optional trailing whitespace

where "whitespace" is limited to SP or HTAB. So, let's just strip *those* from header values.

Note that (some versions of?) py2's `mimetools.Message` includes the trailing CRLF, so strip that, too.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -670,7 +670,7 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
         else:
             headers = [h.split(':', 1) for h in headers]
 
-        env['headers_raw'] = headers_raw = tuple((k, v.strip()) for k, v in headers)
+        env['headers_raw'] = headers_raw = tuple((k, v.strip(' \t\n\r')) for k, v in headers)
         for k, v in headers_raw:
             k = k.replace('-', '_').upper()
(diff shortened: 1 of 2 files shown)


temoto:
Makes sense, thank you.

PR #506, opened 2018-07-11
Fix for Python 3.7

Dormouse759:
This pull request attempts to solve incompatibilities with Python 3.7.
Backwards compatibility shouldn't be broken.


Traceur759:
The coverage problem is due to tests not being run against 3.7.

eventlet/green/ssl.py
```diff
@@ -57,11 +58,41 @@ def __init__(self, sock, keyfile=None, certfile=None,
...
+
+                self = _original_wrap_socket(
+                    sock=sock.fd,
+                    keyfile=keyfile,
+                    certfile=certfile,
+                    server_side=server_side,
+                    cert_reqs=cert_reqs,
+                    ssl_version=ssl_version,
+                    ca_certs=ca_certs,
+                    do_handshake_on_connect=do_handshake_on_connect and six.PY2,
```
hroncok: the py2 part here is redundant given that the entire section is in if > 3.7 block

eventlet/green/ssl.py
```diff
@@ -57,11 +58,40 @@ def __init__(self, sock, keyfile=None, certfile=None,
...
-            ca_certs, do_handshake_on_connect and six.PY2, *args, **kw)
+        if sys.version_info >= (3, 7):
+            # Monkey-patch the sslsocket so our modified self gets
+            # injected into its _create method.
+            def fake_new(self, cls, *args, **kwargs):
+                return self
+
+            orig_new = _original_sslsocket.__new__
+            try:
+                _original_sslsocket.__new__ = fake_new.__get__(self, GreenSSLSocket)
```
jstasiak: Do I understand correctly that this patches/unpatches the original SSLSocket class? If that's the case there's gonna be a race condition - if another thread uses SSLSocket after we patch it but before we unpatch it'll break the application if there's no global monkey patching. Unfortunately I have no competing solution ready right now.

Traceur759: Yes, you understand it correctly.
That was also a problem that crossed my mind, but I couldn't think of better way since __init__ is no longer a thing for SSLSocket and the _create class method calls __new__ with it's own class as first argument. Because of this, there is no way of overriding the behavior of _create method.
TL;DR: There is no way of making _create use subclass' instance.

temoto:
@Traceur759 tests are run against 3.7 and it produces recursion in traceback. Could you please look at it?

https://travis-ci.org/eventlet/eventlet/jobs/405772034
https://travis-ci.org/eventlet/eventlet/jobs/405772035

Traceur759:
The problem with 3.7 is that the buildbot is using outdated version of 3.7.
In that version, the wrap_socket calls SSlSocket, which calls __init__ and vice versa.
In the release version, it calls SSLContext instead, and that doesn't cause this recursion.

temoto:
@Traceur759 can you fix recursion with some condition check or code refactor and enable 3.7 in travis? Maybe they already have `- python: 3.7` without -dev suffix.

hroncok:
See for 3.7 final on Travis: https://github.com/pyvec/elsa/commit/e5a832a1afbb460eb5ea7fc6d05a3f680669bfcd

Traceur759:
@temoto Fixing that would mean supporting a dev version, I don't see a point in that.
Using released 3.7 is itself a fix for the recursion.

.travis.yml
```diff
@@ -22,9 +22,9 @@ matrix:
     - {python: 3.6, env: TOXENV=py36-poll}
     - {python: 3.6, env: TOXENV=py36-selects}
 
-    - {python: 3.7-dev, env: TOXENV=py37-epolls}
-    - {python: 3.7-dev, env: TOXENV=py37-poll}
-    - {python: 3.7-dev, env: TOXENV=py37-selects}
+    - {python: 3.7, env: TOXENV=py37-epolls}
+    - {python: 3.7, env: TOXENV=py37-poll}
+    - {python: 3.7, env: TOXENV=py37-selects}
```
hroncok: Note that `sudo: required` and `dist: xenial` was required last time I've tried.

temoto:
This is what was failing, basically, deprecation warning from third-party package.

```
venv-37/lib/python3.7/site-packages/dns/namedict.py:33: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  class NameDict(collections.MutableMapping):
```

temoto:
@jstasiak @hroncok do you have more to say or approve or reject?

hroncok:
I don't like the race condition, but this is still better than current state.

jstasiak:
I'd rather have a code that obviously doesn't work rather than something that can result in subtle issues *but* I think I have an idea how to solve this so we don't have to choose - it'll take few days though as I'm a bit busy at the moment.

onovy:
hi, any progress wit Python 3.7 please? This is blocker for Debian Buster, where we want to have Python 3.7 as default and Eventlet is really important lib for for example OpenStack. Thanks.

Traceur759:
@onovy 
I think everyone is currently waiting on @jstasiak to come up with the solution or to at least provide some info.
In the meantime, it should be okay to backport this patch as a temporary solution.

nat-goodspeed:
@jstasiak, is there any way I could be helpful to your better solution?

eventlet/green/ssl.py
kashirin-alex: if to stay with a fake_new, may be it is to assign it much earlier at the initial patch import,
https://github.com/eventlet/eventlet/blob/master/eventlet/patcher.py#L416

jstasiak:
@nat-goodspeed sorry, I went down a bit of a rabbit hole testing different approaches to this, I'll have something ready today, thanks for the offer though.

jstasiak:
Sorry, I need to give up, it turned out to be much trickier than I expected. I have no better solution at the moment and I won't have time to work on one for a week or so.

nat-goodspeed:
What would be the least invasive upstream change that would permit a robust solution on the eventlet side?

eventlet/green/ssl.py
Traceur759: Wouldn't that just mean higher probability of breaking other apps trying to use SSLsocket?
The less instructions between assignment and reassignment, the lower chance for the race-condition going wrong.

temoto:
@jstasiak @nat-goodspeed ping

nat-goodspeed:
What I was trying to get at with my previous question is: What could we submit as a pull request against the official Python library that would permit a robust eventlet-side solution to this problem? Presumably the smaller/simpler/more-obvious such an upstream change is, the more likely it is to be accepted.
But I haven't dug into this problem as much as others have, which is why I pose it as a question rather than suggesting such an upstream pull request myself.

temoto:
IMHO, as Python3.7 is spreading, it is important to merge non-ideal solution and work on robust later.

Use case 1, old project, say OpenStack: customer says we must support Python3.7, - here's patch, few things don't work, something something and eventlet, - not again! are you kidding... the pure Python lib. OK postpone 3.7 later. - This is angry user.
Use case 2, new project: hey we could use this lib for network with code that doesn't support asyncio yet - just monkey patch and c10k is OK. Oh it doesn't support 3.7? - This is lost user.

nat-goodspeed:
Okay. Discussion so far has not been a resounding, "Go for it!!"

If we believe that this non-ideal solution will work "well enough" for now, and better than the status quo, then let's get it done. @Dormouse759, will you please merge tests/hub_test.py as indicated in the status panel?

Dormouse759:
Here you go.

PR #512, opened 2018-08-02
greendns: don't contact nameservers if one entry is returned from hosts file

danalsan:
getaddrinfo() behaves differently from the standard implementation as
it will try to contact nameservers if only one (IPv4 or IPv6) entry
is returned from /etc/hosts file.

This patch avoids getaddrinfo() querying nameservers if at least one
entry is fetched through the hosts file to match the behavior of
the original socket.getaddrinfo() implementation.

Closes #515

Signed-off-by: Daniel Alvarez <dalvarez@redhat.com>

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -318,12 +318,13 @@ class ResolverProxy(object):
 
     def query(self, qname, rdtype=dns.rdatatype.A, rdclass=dns.rdataclass.IN,
               tcp=False, source=None, raise_on_no_answer=True,
-              _hosts_rdtypes=(dns.rdatatype.A, dns.rdatatype.AAAA)):
+              _hosts_rdtypes=(dns.rdatatype.A, dns.rdatatype.AAAA),
+              use_network=True):
         """Query the resolver, using /etc/hosts if enabled.
 
         Behavior:
         1. if hosts is enabled and contains answer, return it now
-        2. query nameservers for qname
+        2. query nameservers for qname if use_network is True
         3. if qname did not contain dots, pretend it was top-level domain,
            query "foobar." and append to previous result
         """
@@ -360,7 +361,7 @@ class ResolverProxy(object):
 
         if (self._hosts and (rdclass == dns.rdataclass.IN) and (rdtype in _hosts_rdtypes)):
             if step(self._hosts.query, qname, rdtype, raise_on_no_answer=False):
-                if (result[0] is not None) or (result[1] is not None):
+                if (result[0] is not None) or (result[1] is not None) or (not use_network):
                     return end()
 
         # Main query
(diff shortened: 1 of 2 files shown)


eventlet/support/greendns.py
```diff
@@ -398,10 +399,12 @@ def getaliases(self, hostname):
 resolver = ResolverProxy(hosts_resolver=HostsResolver())
 
 
-def resolve(name, family=socket.AF_INET, raises=True, _proxy=None):
+def resolve(name, family=socket.AF_INET, raises=True, _proxy=None,
+            use_dns=True):
```
temoto: Unless this name is commonly known in other APIs, could you please rename it to `use_network`?

eventlet/support/greendns.py
```diff
@@ -469,16 +473,19 @@ def _getaddrinfo_lookup(host, family, flags):
...
+                        raise
+                    err = e
+                else:
+                    if answer.rrset:
+                        addrs.extend(rr.address for rr in answer.rrset)
+            if addrs:
+                break
         if err is not None and not addrs:
             raise err
     elif family == socket.AF_INET6 and flags & socket.AI_V4MAPPED:
```
temoto: When `resolve()` family is specified, it works as before, going to network. Is that how `getaddrinfo` works also?

temoto: There's another `else` block below too.

temoto:
Could you write a test for new behavior?

eventlet/support/greendns.py
danalsan: ack

eventlet/support/greendns.py
danalsan: When resolve() family is specified it works as before, ie. returning the entry on hosts file and not using network if it's there.
My change basically applies to getaddrinfo() where if a IPv4 or IPv6 entry was fetched from /etc/hosts it won't attempt to get it from the network. Before this patch, if for example only an IPv4 entry was found, it'll query the network for the IPv6 resolution. Makes sense?

danalsan:
I sent a new version with the test

eventlet/support/greendns.py
temoto: Yes, sorry, I meant `_getaddrinfo_lookup()` with family not AF_UNSPEC.

tests/greendns_test.py
```diff
@@ -918,6 +923,57 @@ def test_hosts_priority():
...
+    addr_from_ns = '1.0.2.0'
+    addr6_from_ns = 'dead:beef::1'
+
+    hr = _make_host_resolver()
+    rp = greendns.ResolverProxy(hosts_resolver=hr, filename=None)
+    base = _make_mock_base_resolver()
+    base.rr.address = addr_from_ns
+    base.rr6.address = addr6_from_ns
+    rp._resolver = base()
+    greendns.resolver = rp
```
temoto: This modifies global state for all other tests. Please use `@/with mock.patch` or isolated test or `try/finally`.

eventlet/support/greendns.py
danalsan: Yes if family not AF_UNSPEC then my patch is not changing the behavior

temoto: It should, to match system `getaddrinfo`, right?

danalsan: @temoto before my patch, individual calls to `resolve()` (ie. family not being AF_UNSPEC) where already checking hosts file first. Problem was that calls with AF_UNSPEC will call twice to resolve (one with AF_INET6 and one with AF_INET) and if only one (4 or 6) entry is present in hosts file, the other call will contact the network.

temoto:
@jstasiak review please

mangelajo:
Looks good to me! :)

temoto:
@danalsan thank you, merged in 10ff67f84a1e2f53c8eb254944ebf81762041b45

PR #514, opened 2018-08-08
tests: Add ipv4 udp tests for greendns

lhh:
Add a few tests when using ipv4 to udp() function, just to cover all our bases.

Signed-off-by: Lon Hohberger <lon@metamorphism.com>


temoto:
Test only contribution, what an awesome day! Thank you very much Lon!

PR #516, opened 2018-08-22
Regenerate test crt

onovy:
Newer OpenSSL requires RSA key at least 2048 bits


nat-goodspeed:
@beltran, comments?

temoto:
@onovy do you want to make patch that will generate key/cert instead?

onovy:
hi,

this is command:

> openssl req -x509 -sha256 -nodes -days 3650 -newkey rsa:2048 -keyout test_server.key -out test_server.crt

Because validity of certificate is not checked I don't see any benefit in autogenerating certificate during test. Generating RSA2048 key is time consuming and needs lot of random numbers. When OpenSSL changes minimum size and/or cipher setup, you need to change that command / regenerate certificate again.

temoto:
Right, CPU intensive, I didn't think of that.

temoto:
@onovy thank you.

PR #517, opened 2018-08-23
a Result of an upgrade from 0.19 to the latest 0.24

kashirin-alex:
Initially I could not find why the latest 0.24 version gone 2x slower than 0.19

basic test in my case is:
 ./h2load -t 8 -n 5000000 -c 250 -m 1 --h1 http://www.thither.direct/health-check/
 - with 0.19 was around the 50K+ req/s
 - with 0.24 release got to be around the 20K+ req/s

Tried to find all the reasons for the negative effect and the follow rise-up
while, the main fact is the Monotonic is much slower than the Time module by a factor of 135.2

eventlet.hubs.selects, I think can go a change, considering cases when:
  - the listers count get really high, a selector in parts is preferred
  - an error of one fd should not be a point of failure for the rest

Thanks,
Kashirin Alex

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -15,12 +15,11 @@ from eventlet.greenio import (
 from eventlet.hubs import trampoline, IOClosed
 from eventlet.support import get_errno, PY33
 import six
-import monotonic
+from eventlet.green import time as green_time
 
 orig_socket = __import__('socket')
 socket = orig_socket.socket
 timeout_exc = SSLError
-ts_now = monotonic.time.time
 
 __patched__ = [
     'SSLSocket', 'SSLContext', 'wrap_socket', 'sslwrap_simple',
@@ -282,7 +281,7 @@ class GreenSSLSocket(_original_sslsocket):
                         else:
                             raise
             else:
-                end = ts_now() + self.gettimeout()
+                end = green_time.monotonic() + self.gettimeout()
                 while True:
                     try:
                         real_connect(self, addr)
@@ -290,12 +289,12 @@ class GreenSSLSocket(_original_sslsocket):
                         if get_errno(exc) in CONNECT_ERR:
                             trampoline(
                                 self, write=True,
-                                timeout=end - ts_now(), timeout_exc=timeout_exc('timed out'))
+                                timeout=end - green_time.monotonic(), timeout_exc=timeout_exc('timed out'))
                         elif get_errno(exc) in CONNECT_SUCCESS:
                             return
                         else:
                             raise
-                    if ts_now() >= end:
+                    if green_time.monotonic() >= end:
                         raise timeout_exc('timed out')
 
     def connect(self, addr):
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -444,16 +444,17 @@ class BaseHub(object):
         when = self.clock()
 
         while t:
-            if when < t[0][0]:
+            exp, tmr = t[0]
+            if when < exp:
                 break
+            t.pop(0)
             try:
-                t[0][1]()
+                tmr()
             except self.SYSTEM_EXCEPTIONS:
                 raise
             except:
-                self.squelch_timer_exception(t[0], sys.exc_info())
+                self.squelch_timer_exception(tmr, sys.exc_info())
                 clear_sys_exc_info()
-            t.pop(0)
 
     # for debugging:
 
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -412,7 +412,7 @@ class BaseHub(object):
         nxt_timers = self.next_timers
         while nxt_timers:
             scheduled_time, tmr = nxt_timers.pop(-1)
-            if id(tmr) in self.canceled_timers:  # timer got cancelled before assigned
+            if tmr.called:  # timer got cancelled before assigned
                 continue
 
             added = False
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -394,17 +394,19 @@ class BaseHub(object):
             clear_sys_exc_info()
 
     def add_timer(self, tmr):
-        self.timers_count += 1
         scheduled_time = self.clock() + tmr.seconds
-        if not self.timers:
-            self.timers.append((scheduled_time, tmr))
-            return scheduled_time
 
+        added = False
         for i in range(0, self.timers_count):
             if self.timers[i][0] < scheduled_time:
                 continue
+            added = True
             self.timers.insert(i, (scheduled_time, tmr))
             break
+        self.timers_count += 1
+
+        if not added:
+            self.timers.append((scheduled_time, tmr))
         return scheduled_time
 
     def timer_canceled(self, tmr):
@@ -412,8 +414,8 @@ class BaseHub(object):
         for i in range(0, self.timers_count):
             if tmr_id != id(self.timers[i]):  # tmr.scheduled_time != self.timers[i][0]:
                 continue
-            self.timers.pop(i)
             self.timers_count -= 1
+            self.timers.pop(i)
             break
 
     def prepare_timers(self):
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -352,7 +352,7 @@ class BaseHub(object):
                     self.block_detect_post()
 
                 sleep_time = self.timers[0][0] - self.clock() if self.timers else 60.0
-                self.wait(sleep_time if sleep_time > 0 else None)
+                self.wait(sleep_time if sleep_time > 0 else 0)
 
             else:
                 del self.timers[:]
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -467,7 +467,6 @@ class BaseHub(object):
         return t
 
     def fire_timers(self, when):
-        when += 0.000000001
         t = self.timers
         while t:
             exp, tmr = t[0]
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -428,7 +428,6 @@ class BaseHub(object):
                         self.timers_canceled -= 1
                         if not self.timers:
                             break
-                        i -= 1
                         continue
                     if exp < scheduled_time:
                         continue
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -344,12 +344,12 @@ class BaseHub(object):
                     self.close_one()
                 if self.debug_blocking:
                     self.block_detect_pre()
-                self.fire_timers(self.clock())
+                self.fire_timers()
                 if self.debug_blocking:
                     self.block_detect_post()
 
                 sleep_time = self.timers[0][0] - self.clock() if self.timers else 5.0
-                self.wait(sleep_time if sleep_time > 0 else 0)
+                self.wait(sleep_time if 5.0 > sleep_time > 0 else 5.0)
 
             else:
                 del self.timers[:]
@@ -439,8 +439,9 @@ class BaseHub(object):
         self.add_timer(t)
         return t
 
-    def fire_timers(self, when):
+    def fire_timers(self):
         t = self.timers
+        when = self.clock()
 
         while t:
             if when < t[0][0]:
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -352,7 +352,7 @@ class BaseHub(object):
                     self.block_detect_post()
 
                 sleep_time = self.timers[0][0] - self.clock() if self.timers else 60.0
-                self.wait(sleep_time if sleep_time > 0 else 0)
+                self.wait(sleep_time if sleep_time > 0 else None)
 
             else:
                 del self.timers[:]
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -429,9 +429,10 @@ class BaseHub(object):
                         if not self.timers:
                             break
                         continue
+                    i += 1
+                    
                     if exp < scheduled_time:
                         continue
-                    i += 1
                     self.timers.insert(i, (scheduled_time, tmr))
                     added = True
                     break
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -409,10 +409,9 @@ class BaseHub(object):
         self.timers_canceled += 1
 
     def prepare_timers(self):
-        nxt_timers = self.next_timers
         timers = self.timers
-        while nxt_timers:
-            scheduled_time, tmr = nxt_timers.pop(-1)
+
+        for scheduled_time, tmr in self.next_timers:
             if tmr.called:  # timer got cancelled before assigned
                 self.timers_canceled -= 1
                 continue
@@ -422,14 +421,14 @@ class BaseHub(object):
                 i = 0
                 while self.timers_count > i:  # one loop to clean and assign next
                     exp, t = timers[i]
-                    if t.called:   # clear called
-                        timers.pop(i)
+                    if t.called:                # clear called
+                        del timers[i]
                         self.timers_count -= 1
                         self.timers_canceled -= 1
                         if not timers:
                             break
                         continue
-                    if exp > scheduled_time:
+                    if exp > scheduled_time:    # assign next
                         timers.insert(i, (scheduled_time, tmr))
                         added = True
                         break
@@ -437,6 +436,8 @@ class BaseHub(object):
             if not added:
                 timers.append((scheduled_time, tmr))
             self.timers_count += 1
+
+        del self.next_timers[:]
 
     def schedule_call_local(self, seconds, cb, *args, **kw):
         """Schedule a callable to be called after 'seconds' seconds have
@@ -470,7 +471,7 @@ class BaseHub(object):
             exp, tmr = t[0]
             if when < exp:
                 break
-            t.pop(0)
+            del t[0]
             self.timers_count -= 1
             try:
                 tmr()
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -356,15 +356,10 @@ class BaseHub(object):
                 if self.debug_blocking:
                     self.block_detect_post()
                 self.prepare_timers()
-                wakeup_when = self.sleep_until()
-                if wakeup_when is None:
-                    sleep_time = self.default_sleep()
-                else:
-                    sleep_time = wakeup_when - self.clock()
-                if sleep_time > 0:
-                    self.wait(sleep_time)
-                else:
-                    self.wait(0)
+
+                sleep_time = self.timers[0][0] - self.clock() if self.timers else 60.0
+                self.wait(sleep_time if sleep_time > 0 else 0)
+
             else:
                 self.timers_canceled = 0
                 del self.timers[:]
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -420,8 +420,8 @@ class BaseHub(object):
             if self.timers:
                 i = 0
                 while self.timers_count > i:  # one loop to clean and assign next
-                    i += 1
                     exp, t = self.timers[i]
+                    i += 1
 
                     if t.called:   # clear called
                         self.timers.pop(i)
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -418,9 +418,9 @@ class BaseHub(object):
 
             added = False
             if self.timers:
-                skip = 0
-                for i in range(0, self.timers_count):  # one range to clean and assign next
-                    i -= skip
+                i = 0
+                while self.timers_count > i:  # one loop to clean and assign next
+                    i += 1
                     exp, t = self.timers[i]
 
                     if t.called:   # clear called
@@ -429,7 +429,7 @@ class BaseHub(object):
                         self.timers_canceled -= 1
                         if not self.timers:
                             break
-                        skip += 1
+                        i -= 1
                         continue
                     if exp < scheduled_time:
                         continue
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -397,16 +397,17 @@ class BaseHub(object):
         scheduled_time = self.clock() + tmr.seconds
 
         added = False
-        for i in range(0, self.timers_count):
-            if self.timers[i][0] < scheduled_time:
-                continue
-            added = True
-            self.timers.insert(i, (scheduled_time, tmr))
-            break
-        self.timers_count += 1
+        if self.timers:
+            for i in range(0, self.timers_count):
+                if self.timers[i][0] < scheduled_time:
+                    continue
+                added = True
+                self.timers.insert(i, (scheduled_time, tmr))
+                break
 
         if not added:
             self.timers.append((scheduled_time, tmr))
+        self.timers_count += 1
         return scheduled_time
 
     def timer_canceled(self, tmr):
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -419,18 +419,19 @@ class BaseHub(object):
             if self.timers:
                 skip = 0
                 for i in range(0, self.timers_count):  # one range to clean and assign next
+                    i -= skip
                     if self.timers[i][1].called:     # GC
                         self.timers_count -= 1
                         self.timers.pop(i)
-                        skip -= 1
                         self.timers_canceled -= 1
+                        skip -= 1
                         continue
 
-                    if self.timers[i - skip][0] < scheduled_time:
+                    if self.timers[i][0] < scheduled_time:
                         continue
 
                     added = True
-                    self.timers.insert(i - skip, (scheduled_time, tmr))
+                    self.timers.insert(i, (scheduled_time, tmr))
                     break
 
             if not added:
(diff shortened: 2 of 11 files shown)


kashirin-alex:
Hmm... since when Travis Checks py code on the commit comments?

jstasiak:
Hey, thanks for the patches but I'm afraid this is unlikely to be accepted.

Here are the reasons:
* Your change removes `monotonic` usage which was put there for a reason (https://github.com/eventlet/eventlet/pull/303) - you can override eventlet.hubs.default_clock if you want to use another clock implementation
* There's also a bunch of commit messages that are not self-contained, the commit messages are repeated (which suggests the commits should be squashed) and the commit messages do not explain why the changes are being made. There needs to be an explanation for those changes. If it's about optimisation there need to be some benchmarks performed to show, that the changes are in fact beneficial (and even then they may not be worth it).

jstasiak:
Let ma clarify - the monotonic part is the most problematic, the rest is not strictly out of the window but the changes need to live in logically self-contained (but not too big also) commits with commit descriptions explaining why the changes are being done (+ benchmarks if it's about optimising things).

kashirin-alex:
Ye, squashed should do, while equivalent these would be the same changes/commits
(not much to explain on attribs access reduction)

regarding monotonic, if it would not harm performance it would not have be noticed, unless there is other explanation on this (with py2/py3/pypy), ignoring in this case TimeZones offset (it is used for timer, ), whereas nano-secs are useful.

kashirin-alex:
Was it suppose to be time.monotonic (py3) and not with other implementations?
while py2, in that case could have retained with time.time, 

-- I'll try to squash the commits  right, while the commit should be considered, as why to use eventlet.Event when it is an import of Event from eventlet.event to work with. (performance test would be only showing up only on big/huge number of iteration) 

-- I'm happy with the 0.24 release, while, the clock has a big effect and I'll be using my fork for now.

Thanks

temoto:
Don't worry with squash, it's here in Github interface.

eventlet/hubs/hub.py
```diff
@@ -119,7 +122,7 @@ def __init__(self, clock=None):
         self.closed = []
 
         if clock is None:
-            clock = monotonic.monotonic
+            clock = monotonic.time.time
```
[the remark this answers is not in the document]
kashirin-alex: system time adjustments apply to any system that actual in sync.
monotonic.time.time has the same performance results as time.time while it is the system clock without accounting for screw offsets.
Will be good to confirm if that is not correct, https://stackoverflow.com/a/48491746

eventlet/event.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import print_function
 
-from eventlet import hubs
-from eventlet.support import greenlets as greenlet
+from eventlet.hubs import get_hub
```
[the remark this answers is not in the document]
kashirin-alex: if that is the case for get_hub,
there are other cases it is imported, https://github.com/eventlet/eventlet/blob/master/eventlet/timeout.py#L28

actually runtime, it is PyPy (else-wise not intending to move from py on existing projects)

kashirin-alex:
relating to importing methods,
what can be the correct to:
from eventlet.support.greenlets import getcurrent
It should be a really direct access methods, while if considering patches to getcurrent, 
than the greenlets will need to be imported first in the chain, setting it's method.
-- let me know, I'll align this one.

kashirin-alex:
related to importing methods, what will be the correct about:
`from eventlet.support.greenlets import getcurrent`
It should be a direct accessed method, while if to consider to patch it, than the greenlets should be imported first in chain.

-- let me know and I'll align it.

kashirin-alex:
There some more updates to come and under-go evaluations.
If commits to my branch(to this pull request) bother in any way , you can close the request. and when time comes I'll re-open.

Thanks, 
there are somethings, of wonder, (eg, Listeners of type write get a noop of type read, on case when cb executed anyway for a give event type on poll select)

eventlet/hubs/selects.py
```diff
@@ -18,43 +18,62 @@ def _remove_bad_fds(self):
...
+                if not self.listeners_r and not self.listeners_w:
+                    return
+                seconds = 0
             else:
-                raise
+                return
+
+        try:
+            r, w, er = select.select(list(self.listeners_r), list(self.listeners_w),
+                                     list(self.listeners_r)+list(self.listeners_w),
```
temoto: `list(self.listeners_[rw])` is computed twice. This is a perfect example of low hanging fruit optimization.

eventlet/hubs/selects.py
```diff
@@ -18,43 +18,62 @@ def _remove_bad_fds(self):
...
+                        self.listeners_w.get(fd, noop).cb(fd)
                 except self.SYSTEM_EXCEPTIONS:
-                    raise
+                    continue
                 except:
-                    self.squelch_exception(fileno, sys.exc_info())
+                    self.squelch_exception(fd, sys.exc_info())
                     clear_sys_exc_info()
+
+        except select.error as e:
```
temoto: Avoid deep nested code.

```
try:
  r, w, er = select.select()
except select.error as e:
  # handle select error

# handle r, w, er
```

eventlet/hubs/selects.py
```diff
@@ -18,43 +18,62 @@ def _remove_bad_fds(self):
...
+                if get_errno(e) == errno.EINTR:
```
temoto: Indentation problem.

eventlet/hubs/timer.py
```diff
@@ -1,15 +1,17 @@
...
 import six
-from eventlet.hubs import get_hub
 
 """ If true, captures a stack trace for each timer when constructed.  This is
 useful for debugging leaking timers, to find out where the timer was set up. """
 _g_debug = False
 
 
 class Timer(object):
+    __slots__ = ['scheduled_time', 'seconds', 'tpl', 'called', 'traceback']
```
temoto: As with FdListener, namedtuple/list may be even better than slots.

eventlet/timeout.py
```diff
@@ -175,5 +174,4 @@ def fun(*args, **kwargs):
 
 
 def is_timeout(obj):
-    py3err = getattr(__builtins__, 'TimeoutError', Timeout)
-    return bool(getattr(obj, 'is_timeout', False)) or isinstance(obj, py3err)
+    return bool(getattr(obj, 'is_timeout', False)) or isinstance(obj, getattr(__builtins__, 'TimeoutError', Timeout))
```
temoto: This needs to be supported by numbers.

eventlet/greenpool.py
```diff
@@ -1,18 +1,23 @@
...
 __all__ = ['GreenPool', 'GreenPile']
 
 DEBUG = True
+getcurrent = greenlet.getcurrent
 
 
 class GreenPool(object):
     """The GreenPool class is a pool of green threads.
     """
+    __slots__ = ['size', 'coroutines_running', 'sem', 'no_coros_running']
```
temoto: This will definitely break a lot of code in other projects. Needs very strong support by numbers.

eventlet/greenpool.py
```diff
@@ -1,18 +1,23 @@
...
+from eventlet.queue import LightQueue
+from eventlet.semaphore import Semaphore
+from eventlet.event import Event
 from eventlet.support import greenlets as greenlet
 import six
 
 __all__ = ['GreenPool', 'GreenPile']
 
 DEBUG = True
+getcurrent = greenlet.getcurrent
```
temoto: This is only confusing, please don't play.

eventlet/greenpool.py
```diff
@@ -1,18 +1,23 @@
 import traceback
 
 import eventlet
-from eventlet import queue
+from eventlet.greenthread import GreenThread
```
temoto: Again, no `from module import name`.

eventlet/event.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import print_function
 
 from eventlet import hubs
-from eventlet.support import greenlets as greenlet
+from eventlet.support.greenlets import getcurrent
```
temoto: Again, no `from module import name`.

eventlet/greenthread.py
```diff
@@ -206,7 +206,7 @@ def unlink(self, func, *curried_args, **curried_kwargs):
 
         Remove successfully return True, otherwise False
         """
-        if not getattr(self, '_exit_funcs', None):
+        if self._exit_funcs is None or not self._exit_funcs:
```
temoto: Just `if not self._exit_funcs` will work for `None` value also.

eventlet/hubs/poll.py
```diff
@@ -19,43 +18,43 @@ def __init__(self, clock=None):
         self.poll = select.poll()
 
     def add(self, evtype, fileno, cb, tb, mac):
-        listener = super(Hub, self).add(evtype, fileno, cb, tb, mac)
+        listener = BaseHub.add(self, evtype, fileno, cb, tb, mac)
```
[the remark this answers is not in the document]
kashirin-alex: a case usage on the epoll:
listener = BaseHub.add(self, evtype, fileno, cb, tb, mac)
of the reason, it is inherits from poll.Hub


no problem, that came in of a check-ups on primitive or object(w.slots)

eventlet/hubs/hub.py
```diff
@@ -38,8 +43,9 @@ def closed_callback(fileno):
 
 
 class FdListener(object):
+    __slots__ = ['evtype', 'fileno', 'cb', 'tb', 'mark_as_closed', 'spent', 'greenlet', 'where_called']
```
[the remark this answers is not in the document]
kashirin-alex: may be lightest, 
in corresponding order in a list/tuple, with helping methods of get_"named" ,(without confusedness on idx-k)
-- might not better for access ,  obj_attrib+key_seek , now it is only obj_attrib

good to know, it is a place to lighten things.

eventlet/hubs/hub.py
```diff
@@ -100,26 +100,33 @@ def __repr__(self):
...
 def alarm_handler(signum, frame):
-    import inspect
     raise RuntimeError("Blocking detector ALARMED at" + str(inspect.getframeinfo(frame)))
 
 
 class BaseHub(object):
     """ Base hub class for easing the implementation of subclasses that are
     specific to a particular underlying event architecture. """
-
+    __slots__ = ['listeners_r', 'listeners_w', 'secondaries_r', 'secondaries_w', 'listeners', 'secondaries',
```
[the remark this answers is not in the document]
kashirin-alex: OK! , so as corresponding/hence changes will be rolled-back(adjusted back)

eventlet/hubs/hub.py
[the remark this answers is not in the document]
kashirin-alex: It is a damn with a surprise (monotonic.time.time is time.time) !
to pass a clock, it is all well. While, monotonic is of no exception of not to use, NTP is an  
inseparable part on my systems and to skip timers of expiry is unacceptable (delay of a less weight).

eventlet/green/ssl.py
```diff
@@ -137,6 +140,7 @@ def read(self, *args, **kwargs):
             return b''
 
     def send(self, data, flags=0):
+        # *NOTE: overflow on big data
```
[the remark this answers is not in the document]
kashirin-alex: I need to check further,
Socket dropping , might be the server side, a case happen with Hypertable ThriftClient using send(data) method from TTransport.TFramedTransport(TSocket.TSocket(host, port))
 overflow can be the main issue, tried for 1~GB while 100~MB ok

temoto:
I think we can put correct changes into three categories:

1) Obviously safe and useful changes such as:
- `GreenThread` `_exit_funcs`
- `Hub` `register(fileno, new=new)`
- `BaseHub` `_old_signal_handler`
- `exp, tmr = t[0]`
- `switch = hubs.get_hub().switch`
2) Risky changes that need benchmark support, such as `__slots__`, `listeners_[rw]`.
3) Weird shit like getattr optimizations. Do benchmarks. I am happy to be proven wrong, but most likely it's a waste of time.

eventlet/timeout.py
kashirin-alex: this one is pretty simple, to agree upon.
lighter condition should come first on OR operator
the py3err  happens to be getattr+isinstance

kashirin-alex:
OK, i'll keep you posted, if I miss a category.
probably 4th is missing, 
for global var used in local scopes, "timer" has most of occurrences , changed to "tmr" in this cases.

eventlet/timeout.py
temoto: This is how support by numbers works:
- you take a real-like program and make it as determinant as possible: remove network, disk, human operations
- run program few times and measure CPU time, numbers from different runs must be close, otherwise we don't have a stable base to compare against
- change code
- run program few times and measure CPU time
- compare CPU times before and after change

temoto:
Changes like `timer -> tmr` are safe and useful for clarity when changing code. But they don't touch performance, so irrelevant in this talk? If you're trying to fix all problems in Eventlet within one pull request, that will be slow and frustrating for all involved parties.

kashirin-alex:
Something like, timer -> tmr, I can't ignore such mistakes(path to an fatal error) it is imported as timer and used in local scopes as timer,
one case:
does, from eventlet.hubs import timer
followed, https://github.com/eventlet/eventlet/blob/master/eventlet/greenthread.py#L34

eventlet/timeout.py
kashirin-alex: It is OK, to decide over two different methods and so on,
while, if the same operation is faster against additional operation to the same, there is nothing to check about.

temoto:
On default clock via environment.

- First, thank you for working on this. We've discussed this idea, but just using `monotonic` was acceptable for everyone.
- This code should be in hub, because it's not related to greening. Green modules are stdlib modules with blocking methods replaced by cooperative version. Which means green module requires hub to work, not the other way around.
- Method is callable in class instance. Callable at the top level of module is called "function". Having two environment variables to configure clock is too complicated, one is very enough: `EVENTLET_CLOCK=time.time` `rsplit('.', 1)` is your friend.

kashirin-alex:
yes, the split of the full method/func of clock path will do.

considering, the green.ssl need timestamp , (and it cannot use the time built)
Why is there the time in green, complicates the import time (for within green module)

temoto:
You can run tests with simple `tox -e py27-selects` command on your machine.

temoto:
I will unsubscribe from this spam of commits, write me when you are finished.

kashirin-alex:
OK, that will be great, 
soon I'll put a summary to these (hub evaluation).
got a 10%+  (https://github.com/eventlet/eventlet/pull/517/commits/719dab55f894dfa87ac7a6cc6645f994e60dbe7b)
and with less to compute got to more sleeping periods

temoto:
How to run your benchmark?

kashirin-alex:
I need to make one,  overall is based on my platform results,  the upper
test "h2load"  (it includes more than 1k timers considering h2 around 20k)

As so,  what would be a good test for GC of timers test,  Should an
expected count be enough?

kashirin-alex:
for fun facts,
- "timers" is by the hub.get_timers_count
- "readers" is by len(hub.listeners['read'])

eventlet-hubs-monitoring

temoto:
I don't understand exactly this question about GC of timers.

But here's general rule that applies everywhere: if you fix issue, first you write test that fails with current code, then change code and all tests must pass.

If you think there is excess GC, write a benchmark that measures them in master, see if new code makes less.

kashirin-alex:
GC slight miss defines the meaning to "clean" of called timers (while it is
garbage collection logic.
Yes,  of course corresponding test and benchmark must be supporting updates
beyond of py preferences for a process  .

temoto:
What other languages you speak?

kashirin-alex:
Hebrew and Russian

temoto:
Пожалуйста, объясни про gc, я переведу.

kashirin-alex:
Regarding the languages,  English is the easier(most used) on my side,
while at no point actually matter until I get used to back it, which is a
positive of not to forget and a good reminder.

GC,  the prepare_timers do a clean and the type of clean it does is to
remove the junk(the called timers)  that are no longer needed to remain.
Это всего об этом. Конещно GC о статусе "ref" и удаление это другая
процедура.

temoto:
OK. Do you see a problem with timers garbage collection?

kashirin-alex:
The current related tests don't match to expectations that threshold occurs
on a 1000~ base to clear the called timers.

temoto:
What threshold do you think is adequate?

kashirin-alex:
ERROR: test_cancel_proportion (tests.hub_test.TestTimerCleanup)
to https://travis-ci.org/eventlet/eventlet/builds/420679142?utm_source=github_status&utm_medium=notification

the Takes To Long, it has entered the 60 second timer OR the default (follow a sleep step applied)
-- Is it correct evaluation ?

temoto:
You didn't respond on new threshold question.

If cleanup happened without threshold as you say, test would pass without modification. You changed code to no cleanup until entering hub `eventlet.sleep()`.

Think like this, there is code that tests exactly this behavior: start and cancel a lot of timers and assert their number remains under control. Why somebody would check for it? Must be they was in situation where Eventlet timer management is source of problem.

Changing test condition (sleep before assert) is not acceptable.

kashirin-alex:
Regarding a commit (https://github.com/eventlet/eventlet/pull/517/commits/67753c767048b36b7bc40306ba8dbd76b2e52298) prior "hub evaluation" commits
main remaining questions:

- _Should a noop be called for a none existing listeners?_ https://github.com/eventlet/eventlet/blob/master/eventlet/hubs/poll.py#L97 https://github.com/eventlet/eventlet/blob/master/eventlet/hubs/hub.py#L250

some tests(benchmarks) results for the https://github.com/eventlet/eventlet/pull/517/commits/67753c767048b36b7bc40306ba8dbd76b2e52298 commit , eventlet/examples/regression.py
configurations by default
    _num_workers = 10
    num_range_timers = 1000
    ms_delay_range = [1, 5001, 500]_

**Python 2.7.15**

_commit https://github.com/kashirin-alex/eventlet/archive/67753c767048b36b7bc40306ba8dbd76b2e52298.zip_

-------------------------------------------------------------------------------
Results:
****************************************
Timers Expected:        100000
Timers Initiated:       100000
Timers Executed:        100000
Time Took:              4.63258099556
Delay Differential Sum: 20.143487215
Delay Differential Avg: 0.00020143487215
****************************************
-------------------------------------------------------------------------------

_eventlet master version:_

-------------------------------------------------------------------------------
Results:
****************************************
Timers Expected:        100000
Timers Initiated:       100000
Timers Executed:        100000
Time Took:              4.7053539753
Delay Differential Sum: 21.0580363274
Delay Differential Avg: 0.000210580363274
****************************************
-------------------------------------------------------------------------------


**Python 3.7.0b4**

_commit https://github.com/kashirin-alex/eventlet/archive/67753c767048b36b7bc40306ba8dbd76b2e52298.zip_

-------------------------------------------------------------------------------
Results:
****************************************
Timers Expected:        100000
Timers Initiated:       100000
Timers Executed:        100000
Time Took:              4.959421634674072
Delay Differential Sum: 45.71040272716551
Delay Differential Avg: 0.0004571040272716551
****************************************
-------------------------------------------------------------------------------

_eventlet master version:_

------------------------------------------------------------------------------
Results:
****************************************
Timers Expected:        100000
Timers Initiated:       100000
Timers Executed:        100000
Time Took:              5.029201984405518
Delay Differential Sum: 12.336063385023511
Delay Differential Avg: 0.00012336063385023511
****************************************
-------------------------------------------------------------------------------


**Python 3.7.0b4**

_commit https://github.com/kashirin-alex/eventlet/archive/67753c767048b36b7bc40306ba8dbd76b2e52298.zip_

-------------------------------------------------------------------------------
Results:
****************************************
Timers Expected:        100000
Timers Initiated:       100000
Timers Executed:        100000
Time Took:              5.06739807129
Delay Differential Sum: 61.8325881958
Delay Differential Avg: 0.000618325881958
****************************************
-------------------------------------------------------------------------------

_eventlet master version:_

-------------------------------------------------------------------------------
Results:
****************************************
Timers Expected:        100000
Timers Initiated:       100000
Timers Executed:        100000
Time Took:              5.07813811302
Delay Differential Sum: 56.9594521523
Delay Differential Avg: 0.000569594521523
****************************************
-------------------------------------------------------------------------------

temoto:
I think this pull request is out of control.

- Thank you for safe and obviously correct changes, such as `_exit_funcs`, listed above. I am opening another review for those.
- If you want to have `EVENTLET_CLOCK` merged, please make a separate patch which does not change anything else + test for new code. Environment parsing should go into hub code. ssl can use `get_hub().clock`
- If you want to speedup timer management, please make a separate patch which does not change anything else, start with benchmark that anybody can run with a single command. `eventlet/examples` is not place for tests.
- Pull request is not place for active development. Make it when your work is done. AFAIK, Travis will happily build every branch in your fork without distracting other people.

kashirin-alex:
the EVENTLET_CLOCK  pull request in the making

kashirin-alex:
Some many Commits on my master branch,
https://github.com/kashirin-alex/eventlet
Main Major is the Hub's optimal loop self.run that do it all, https://github.com/kashirin-alex/eventlet/blob/master/eventlet/hubs/hub.py#L346
(so far so good no issues seen)
-- Not making a Pull Request from my master branch on all current commit at once.
if really the Optimal Loop do better than time will come.

temoto:
@kashirin-alex understood. If you see the new benchmarks runner 90c398387442654dc1f9392ba0e4a2296875334b it may help you see which changes are more important than others.

kashirin-alex:
@temoto , OK thanks, Surely will do.

temoto:
@kashirin-alex From first look it seems `Event` will not work with socket functions like `select`. Try and see. If you write a new hub in a more efficient way and everything remains correct - that's your answer.

The most important part: `select` is a backup plan for Windows.
If you really run performance critical application on Python/Windows, then switch from Eventlet to something else.
There are two important hubs actually used in production servers: epoll/Linux and kqueue/FreeBSD. If you run `select` on your server, that is your performance problem, not monotonic, not timers.

If you want general chat with people with experience in handling event loops, look at nginx, gevent, tornado, twisted, redis, go, haskell, erlang beam, java nio, guides for using nonblocking sockets.

kashirin-alex:
the select hub works OK/Equal comparing to epolls or poll, just it's sleep is time.sleep (for the no FDs to select)
and improvement can go in-place of that sleep with an interruptible one 

question,
ns/op is it time took for an operation , or else?


Python 3.7.0b4 | my master |   |   | eventlet master |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | -- | --
Benchmark_hub_timers | 2187 | 2821492 |   | 2187 | 5470501 |   | 0.5157648267 | ns/op
Benchmark_pool_spawn | 1000000 | 6105 |   | 1000000 | 6124 |   | 0.9968974526 | ns/op
Benchmark_pool_spawn_n | 100000 | 4105 |   | 100000 | 4427 |   | 0.9272645132 | ns/op
Benchmark_sleep | 1000000 | 4359 |   | 1000000 | 6152 |   | 0.708550065 | ns/op
Benchmark_pool_spawn | 1000000 | 6055 |   | 1000000 | 6089 |   | 0.9944161603 | ns/op
Benchmark_spawn | 1000000 | 12292 |   | 200000 | 15014 |   | 0.8187025443 | ns/op
Benchmark_spawn_link1 | 1000000 | 13535 |   | 200000 | 16751 |   | 0.808011462 | ns/op
Benchmark_spawn_link5 | 200000 | 16996 |   | 200000 | 19889 |   | 0.8545427121 | ns/op
Benchmark_spawn_link5_unlink3 | 200000 | 17247 |   | 200000 | 21011 |   | 0.8208557422 | ns/op
Benchmark_pool_spawn_n | 100000 | 4137 |   | 100000 | 4415 |   | 0.9370328426 | ns/op
Benchmark_spawn_n | 100000 | 2599 |   | 100000 | 2703 |   | 0.9615242323 | ns/op
Benchmark_spawn_n_kw | 100000 | 2714 |   | 100000 | 2977 |   | 0.9116560296 | ns/op
Benchmark_spawn_nowait | 100000 | 3392 |   | 100000 | 3591 |   | 0.9445836814 | ns/op

temoto:
Run `bin/bench-compare`, do not invent tables.

temoto:
ns/op is nanoseconds per operation.

temoto:
Ah no git, sorry. Main point: new benchmarks runner was specifically written to allow machine compare numbers automatically.

```
cd eventlet-upstream
/path/to/virtualenv/bin/python -m benchmarks >bench-base
cd eventlet-kashirin
/path/to/virtualenv/bin/python -m benchmarks >bench-new
go get golang.org/x/tools/cmd/benchcmp  # this is like pip install, need once but fast to skip
benchcmp bench-{base,new}
```

temoto:
^ this is to save your time in future.

I see that hub_timers is much faster, congratulations.

kashirin-alex:
Thanks, I miss "go" for today,

Still need to under-go checks/runtime , occurrences noticed it is sleep_till happens more occasionally.
hub timers under-go big reduction in the mem size (float less to all and tuple)

kashirin-alex:
I'll have a full benchmark run tomorrow.
(and actual do have GIT on my servers but not using for pushing, and not funny it should not have been but might got with some of the app inner dependencies)

kashirin-alex:
sorry, I could for today run the GO comparison and to commit it not yet proper. There are something that need to be concluded (or yet conclude)
Attached are the original benchmarks outputs by the py version.




[benchmarks.zip](https://github.com/eventlet/eventlet/files/2329614/benchmarks.zip)

PR #518, opened 2018-08-26
configure hub clock source via EVENTLET_CLOCK

kashirin-alex:
Hub's default clock is conditioned on the following order:

1.  a clock method is passed to hub's init
2.  a passed method on environ EVENTLET_CLOCK=module.module.method
3.  a monotonic.monotonic is used

  - at run time, the clock is available at hubs.get_hub().clock
  - green.ssl changed to use hub's clock

Related:
#517 
#388
#303
#270
#132

if at some point a clock need to be changed for the initiated hub at run time, that can be done by: 
`hubs.get_hub().clock = new_clock_method`
(that cannot be recommend for clocks with big difference)

### DIFF
--- a/eventlet/__init__.py
+++ b/eventlet/__init__.py
@@ -21,8 +21,11 @@ if os.environ.get('EVENTLET_IMPORT_VERSION_ONLY') != '1':
     # Helpful when CPython < 3.5 on Linux blocked in `os.waitpid(-1)` before first use of hub.
     # Example: gunicorn
     # https://github.com/eventlet/eventlet/issues/401#issuecomment-327500352
-    import monotonic
-    del monotonic
+    import time
+    if not hasattr(time, 'monotonic'):
+        import monotonic
+        del monotonic
+    del time
 
     connect = convenience.connect
     listen = convenience.listen
--- a/eventlet/__init__.py
+++ b/eventlet/__init__.py
@@ -21,11 +21,8 @@ if os.environ.get('EVENTLET_IMPORT_VERSION_ONLY') != '1':
     # Helpful when CPython < 3.5 on Linux blocked in `os.waitpid(-1)` before first use of hub.
     # Example: gunicorn
     # https://github.com/eventlet/eventlet/issues/401#issuecomment-327500352
-    import time
-    if not hasattr(time, 'monotonic'):
-        import monotonic
-        del monotonic
-    del time
+    import monotonic
+    del monotonic
 
     connect = convenience.connect
     listen = convenience.listen
--- a/eventlet/greenpool.py
+++ b/eventlet/greenpool.py
@@ -37,8 +37,7 @@ class GreenPool(object):
         their tasks to drop the overall quantity below *new_size*.  Until
         then, the return value of free() will be negative.
         """
-        size_delta = new_size - self.size
-        self.sem.counter += size_delta
+        self.sem.counter += new_size - self.size
         self.size = new_size
 
     def running(self):
@@ -94,9 +93,7 @@ class GreenPool(object):
         finally:
             if coro is None:
                 return
-            else:
-                coro = eventlet.getcurrent()
-                self._spawn_done(coro)
+            self._spawn_done(eventlet.getcurrent())
 
     def spawn_n(self, function, *args, **kwargs):
         """Create a greenthread to run the *function*, the same as
@@ -105,8 +102,7 @@ class GreenPool(object):
         """
         # if reentering an empty pool, don't try to wait on a coroutine freeing
         # itself -- instead, just execute in the current coroutine
-        current = eventlet.getcurrent()
-        if self.sem.locked() and current in self.coroutines_running:
+        if self.sem.locked() and eventlet.getcurrent() in self.coroutines_running:
             self._spawn_n_impl(function, args, kwargs, None)
         else:
             self.sem.acquire()
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -27,10 +27,7 @@ import six
 
 mod = os.environ.get('EVENTLET_CLOCK')
 if not mod:
-    time = patcher.original('time')
-    eventlet_clock = getattr(time, 'monotonic') if hasattr(time, 'monotonic')\
-        else getattr(patcher.original('monotonic'), 'monotonic')
-    del time
+    from monotonic import monotonic as eventlet_clock
 else:
     mod = mod.rsplit('.', 1)
     eventlet_clock = getattr(patcher.original(mod[0]), mod[1])
--- a/eventlet/hubs/hub.py
+++ b/eventlet/hubs/hub.py
@@ -2,6 +2,7 @@ import errno
 import heapq
 import math
 import signal
+import os
 import sys
 import traceback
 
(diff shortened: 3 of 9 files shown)


eventlet/__init__.py
```diff
@@ -21,8 +21,11 @@
     # Helpful when CPython < 3.5 on Linux blocked in `os.waitpid(-1)` before first use of hub.
     # Example: gunicorn
     # https://github.com/eventlet/eventlet/issues/401#issuecomment-327500352
-    import monotonic
-    del monotonic
+    import time
+    if not hasattr(time, 'monotonic'):
```
temoto: This is redundant. monotonic already uses `time.monotonic` if present.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,20 @@ def alarm_signal(seconds):
 
 from eventlet.hubs import timer, IOClosed
 from eventlet.support import greenlets as greenlet, clear_sys_exc_info
-import monotonic
+from eventlet import patcher
 import six
 
+mod = os.environ.get('EVENTLET_CLOCK')
+if not mod:
+    time = patcher.original('time')
```
temoto: There is no need in original time.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,20 @@ def alarm_signal(seconds):
...
+    eventlet_clock = getattr(time, 'monotonic') if hasattr(time, 'monotonic')\
```
temoto: This is redundant. monotonic already uses `time.monotonic` if present.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,20 @@ def alarm_signal(seconds):
...
+        else getattr(patcher.original('monotonic'), 'monotonic')
+    del time
+else:
+    mod = mod.rsplit('.', 1)
+    eventlet_clock = getattr(patcher.original(mod[0]), mod[1])
```
temoto: There is no need in original here either.

setup.py
```diff
@@ -1,11 +1,21 @@
...
 import eventlet
 
+install_requires = [
+        'dnspython >= 1.15.0',
+        'enum34;python_version<"3.4"',
+        'greenlet >= 0.3',
+        'six >= 1.10.0',
+    ]
+if not hasattr(time, 'monotonic'):
+    install_requires.append('monotonic >= 1.4')
```
temoto: Arguably, constant list is better.

I can imagine a bug when eventlet package/wheel is built on machine with `time.monotonic` and requirement is not added for everyone later.

temoto: monotonic already uses builtin time when possible.

Every line of code is something that can break and something we will support later. Only benefit of optional include of monotonic is to make virtualenv few KiB smaller and install few seconds faster *only* if other packages don't depend on monotonic.

kashirin-alex: OK, i'll set monotonic pkg for the decision .

eventlet/hubs/hub.py
kashirin-alex: Should it than be import_patched?
or do you mean something else?
 it is about importing user's original

setup.py
jstasiak: I've done something like this before and it won't work - if a wheel package is generated on a platform with `time.monotonic` available it won't be able to be used on platforms that lack it.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,17 @@ def alarm_signal(seconds):
 
 from eventlet.hubs import timer, IOClosed
 from eventlet.support import greenlets as greenlet, clear_sys_exc_info
-import monotonic
+from eventlet import patcher
 import six
 
+mod = os.environ.get('EVENTLET_CLOCK')
+if not mod:
+    from monotonic import monotonic as eventlet_clock
```
temoto: Again, no `from module import name` please.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,17 @@ def alarm_signal(seconds):
...
+else:
+    mod = mod.rsplit('.', 1)
+    eventlet_clock = getattr(patcher.original(mod[0]), mod[1])
```
temoto: There is no need in patcher.original here.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,17 @@ def alarm_signal(seconds):
...
+del mod
```
temoto: Putting all above in a function `def default_clock()` would fix the need for del.

temoto: But then import in function is a disaster waiting to happen, so keep it here, sorry for noise.

Please rename `eventlet_clock` to `default_clock`. Everything in this project is implicitly `eventlet_` so this prefix does not help.

eventlet/hubs/hub.py
temoto: I mean should be regular Python import.

kashirin-alex: that is good for a switch case for known options, 
on current, user's options with __import__ remain dynamic

eventlet/greenpool.py
```diff
@@ -37,8 +37,7 @@ def resize(self, new_size):
         their tasks to drop the overall quantity below *new_size*.  Until
         then, the return value of free() will be negative.
         """
-        size_delta = new_size - self.size
-        self.sem.counter += size_delta
+        self.sem.counter += new_size - self.size
```
temoto: This is not related to current pull request about EVENTLET_CLOCK.

temoto:
New commits are not related to pull request topic. I think by this time you know what you are doing. I don't know how to handle this situation gracefully.

kashirin-alex:
got by mistake,
is there anyhow to make a pull and lock it from changes on it's associated
branch ?

*Kashirin Alex*(ABN 78 176 155 352)

*Thither Int LLC - CEO & Founder*

@:  kashirin.alex@thither.direct
*s:*    kashirin.alex
*m:*   (+972)-528-733778  *|*  (+7)-963-7856575* |*  (+61)-3-90133928
*ln:*   http://www.linkedin.com/in/kashirinalex
*w:*   Thither.Direct <http://thither.direct/>
*T:*   twitter.com/thitherint


<http://thither.direct/>

Disclaimer: This email and any files transmitted with it are confidential
and intended solely for the use of the individual or entity to whom they
are addressed. If you have received this email in error please notify the
system manager. This message contains confidential information and is
intended only for the individual named. If you are not the named addressee
you should not disseminate, distribute or copy this e-mail. Please notify
the sender immediately by e-mail if you have received this e-mail by
mistake and delete this e-mail from your system. If you are not the
intended recipient you are notified that disclosing, copying, distributing
or taking any action in reliance on the contents of this information is
strictly prohibited.

eventlet/hubs/hub.py
```diff
@@ -21,9 +22,16 @@ def alarm_signal(seconds):
...
-import monotonic
+from eventlet import patcher
 import six
 
+if os.environ.get('EVENTLET_CLOCK'):
+    mod = os.environ.get('EVENTLET_CLOCK').rsplit('.', 1)
+    default_clock = getattr(patcher.original(mod[0]), mod[1])
+    del mod
+else:
+    import monotonic.monotonic as default_clock
```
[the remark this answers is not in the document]
kashirin-alex: the last commit on the hub by the way is the https://github.com/eventlet/eventlet/pull/518/commits/3a910a228e43bc384d72498a29f61ce95a980269

PR #519, opened 2018-08-26
greenthread: optimize _exit_funcs getattr/del dance; Thanks to Alex Kashirin

temoto:
Origin: https://github.com/eventlet/eventlet/pull/517

@kashirin-alex do you want me to put you as git committer for credits?


kashirin-alex:
OK, thanks.

kashirin-alex:
there are couple more thing in the greenthread,

- global scope vars in assigned in local
from eventlet.hubs import timer as ev_timer
and following timer.Timer() to chg to ev_timer.Timer() , or it is to change the local scope var name

- unused vars
https://github.com/eventlet/eventlet/blob/master/eventlet/greenthread.py#L132
https://github.com/eventlet/eventlet/blob/master/eventlet/greenthread.py#L145
https://github.com/eventlet/eventlet/blob/master/eventlet/greenthread.py#L159

kashirin-alex:
not clear condition at https://github.com/eventlet/eventlet/blob/master/eventlet/greenthread.py#L274
if not, it should not have prior checked for dead attribute

temoto:
@kashirin-alex If you find unused names (Python does not have concept of variables), please let us know. What you pointed is not unused vars, they're used once. There is no problem with that. Also it's not related to _exit_funcs, see pull request title.

```
t = expr()
u = expr()  # this is unused name
return t  # t is used here
```

> global scope vars assigned in local

It's not a problem if there is local assignment before reading. It will be problem to use `timer` in that function. So far wasn't needed. Also it's not related to _exit_funcs, see pull request title.

kashirin-alex:
no problem, not to this pull
unused equivalent meaning var is redundant, function's value can be returned without a new var.

benchmarks/spawn.py
```diff
@@ -42,17 +49,42 @@ def run_spawn_n_kw():
...
+    run_spawn_n_kw,
+    run_spawn_link1_wait,
+    run_spawn_link5_wait,
+)
+print(format_result("eventlet.spawn", best[run_spawn], iters))
+print(format_result("eventlet.spawn_n", best[run_spawn_n], iters))
+print(format_result("eventlet.spawn_n(**kw)", best[run_spawn_n_kw], iters))
+print(format_result("eventlet.spawn.link[1]+wait", best[run_spawn_link1_wait], iters))
+print(format_result("eventlet.spawn.link[5]+wait", best[run_spawn_link5_wait], iters))
+print("spawn/spawn_n: {:0.1f} %".format((best[run_spawn] - best[run_spawn_n]) / best[run_spawn_n] * 100))
```
nat-goodspeed: Maybe comparison of results should itself become a helper function, for lines 87 and 118?

eventlet/greenthread.py
```diff
@@ -170,6 +170,7 @@ def __init__(self, parent):
         greenlet.greenlet.__init__(self, self.main, parent)
         self._exit_event = event.Event()
         self._resolving_links = False
+        self._exit_funcs = None
```
nat-goodspeed: Why not unconditionally initialize `self._exit_funcs` to `deque()`?

eventlet/greenthread.py
```diff
@@ -196,7 +197,8 @@ def func(gt, [curried args/kwargs]):
         functions by doing things like switching explicitly to another
         greenthread.
         """
-        self._exit_funcs = getattr(self, '_exit_funcs', deque())
+        if self._exit_funcs is None:
```
nat-goodspeed: Initializing `_exit_funcs` to `deque` would make this test vanish entirely.

eventlet/greenthread.py
```diff
@@ -206,7 +208,7 @@ def unlink(self, func, *curried_args, **curried_kwargs):
 
         Remove successfully return True, otherwise False
         """
-        if not getattr(self, '_exit_funcs', None):
+        if not self._exit_funcs:
```
nat-goodspeed: This early exit could vanish too. Once `_exit_funcs` is always a `deque`, removing this test would optimize valid `unlink()` calls rather than optimizing invalid calls.

eventlet/greenthread.py
```diff
@@ -229,11 +231,12 @@ def _resolve_links(self):
         # ca and ckw are the curried function arguments
         if self._resolving_links:
             return
+        if not self._exit_funcs:
+            return
```
nat-goodspeed: If `_exit_funcs` is always a `deque`, this early exit becomes optional, and its value depends on anticipated usage. But I could live with the assumption that more `GreenThread`s are run without `link()` calls than with `link()` calls.

benchmarks/spawn.py
```diff
@@ -9,21 +9,28 @@ def cleanup():
     eventlet.sleep(0.2)
 
 
-iters = 10000
+def format_result(prefix, time, n):
+    ops = int(n / time)
```
nat-goodspeed: Very minor notion: `ops` could reasonably be computed after the early exit.

benchmarks/spawn.py
nat-goodspeed: That would help address the pep8 style issues.

eventlet/greenthread.py
kashirin-alex: My only worry of wonder,
the problem with self._exit_funcs.remove((func, curried_args, curried_kwargs))
at any point the args or kwargs might be not the same order, and that might effect the match for value to remove

nat-goodspeed: Well, we'd be increasing resource usage for code clarity and fewer inline tests. But I take your point.
Another approach would be a breaking change: requiring the consuming application to explicitly designate that it intends to use `link()` and `unlink()`, and instantiating a `GreenThread` subclass containing `_exit_funcs` for such applications. But such a change is out of scope for this PR in any case, and may be completely untenable.

benchmarks/spawn.py
temoto: Fixed benchmarks in a different way.

PR #520, opened 2018-08-26
ssl: connect used non-monotonic time.time() for timeout; Thanks to Alex Kashirin

temoto:
@kashirin-alex do you want me to put you as git committer for credits?

Origin: https://github.com/eventlet/eventlet/pull/517

Related:
https://github.com/eventlet/eventlet/pull/388
https://github.com/eventlet/eventlet/pull/303
https://github.com/eventlet/eventlet/issues/270
https://github.com/eventlet/eventlet/issues/132


eventlet/green/ssl.py
```diff
@@ -260,6 +259,7 @@ def do_handshake(self):
             super(GreenSSLSocket, self).do_handshake)
 
     def _socket_connect(self, addr):
+        clock = hubs.get_hub().clock
```
kashirin-alex: get clock should be in the else of if self.gettimeout() is None:
it is not used in other cases

temoto:
@jstasiak I'm not sure how to write reason for change in imperative form. It fits our guidlines though. :-) https://github.com/eventlet/eventlet/blob/master/CONTRIBUTING.md

jstasiak:
I think the reason (commit message body) can be non-imperative just fine, it's about the first line. Anyway it looks good to me.

eventlet/green/ssl.py
temoto: Thanks for caring of performance so much, fixed.

nat-goodspeed:
`AssertionError: CPU usage over limit: user 12% sys 0% allowed 12%` in `py27-poll` is a known issue, correct?

PR #523, opened 2018-09-08
wsgi: make Expect 100-continue field-value case-insensitive.

jkasarherou:
The wsgi component is not handling correctly the Expect 100-continue field value.

I found this using the `aws-sdk-go`, that sends the value `100-Continue` (!= `100-continue`).

See https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.20

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -684,7 +684,7 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
             else:
                 env[envk] = v
 
-        if env.get('HTTP_EXPECT') == '100-continue':
+        if env.get('HTTP_EXPECT', '').lower() == '100-continue':
             wfile = self.wfile
             wfile_line = b'HTTP/1.1 100 Continue\r\n'
         else:
--- a/tests/wsgi_test.py
+++ b/tests/wsgi_test.py
@@ -748,27 +748,29 @@ class TestHttpd(_TestBase):
         result = read_http(sock)
         self.assertEqual(result.status, 'HTTP/1.1 417 Expectation Failed')
         self.assertEqual(result.body, b'failure')
-        fd.write(
-            b'PUT / HTTP/1.1\r\nHost: localhost\r\nContent-length: 7\r\n'
-            b'Expect: 100-continue\r\n\r\ntesting')
-        fd.flush()
-        header_lines = []
-        while True:
-            line = fd.readline()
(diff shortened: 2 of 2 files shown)


eventlet/wsgi.py
```diff
@@ -684,7 +684,7 @@ def get_environ(self):
             else:
                 env[envk] = v
 
-        if env.get('HTTP_EXPECT') == '100-continue':
+        if env.get('HTTP_EXPECT') and env['HTTP_EXPECT'].lower() == '100-continue':
```
temoto: `if env.get('HTTP_EXPECT', '').lower() ...`

tests/wsgi_test.py
```diff
@@ -748,27 +748,29 @@ def wsgi_app(environ, start_response):
...
-                break
-            else:
-                header_lines.append(line)
-        assert header_lines[0].startswith(b'HTTP/1.1 200 OK')
-        assert fd.read(7) == b'testing'
+
+        for expect_value in ('100-continue', '100-Continue'):
+            fd.write(
+                b'PUT / HTTP/1.1\r\nHost: localhost\r\nContent-length: 7\r\n'
+                b'Expect: %s\r\n\r\ntesting' % expect_value.encode())
```
temoto: Please, never use `%` templating.
It's bad in general (because tuple special case) and in this particular case it broke py3.4 tests.

`'bla bla bla: {}'.format().encode()`

temoto:
I see no problem merging this patch on basis "be strict in what you send, be liberal in what you accept".

But, strictly speaking, Eventlet behaves correctly here. HTTP spec says header names are case insensitive, but header values are case sensitive and per RFC2616#8.2.3 https://tools.ietf.org/html/rfc2616#section-8.2.3 the string must be `100-continue` lower case. So if you care, please also open cross issue to aws-sdk-go or golang/http/client, whichever defines this string.

jkasarherou:
Thanks @temoto for reviewing my PR, I made the necessary changes based on your comments.

HTTP spec says this:

> Comparison of expectation values is case-insensitive for unquoted tokens (including the 100-continue token), and is case-sensitive for quoted-string expectation-extensions.

https://tools.ietf.org/html/rfc2616#section-14.20

temoto:
Oh, I'm sorry, missed that, thanks for education.

vincent-legoll:
LGTM

Issue #408, opened 2017-05-01
Memory leak when GreenThread objects reference themselves


mattbennett:
GreenThreads and their locals do not appear to be garbage collected if one of the locals contains a reference to the thread and the thread results in an exception.

Example:

``` python
import eventlet
import eventlet.debug
eventlet.debug.hub_exceptions(False)

import objgraph
import gc

threads = {}

class Foo(object):
    def __init__(self, index):
        self.thread = threads[index]

def target(index):
    foo = Foo(index)
    if index % 2 == 0:
        raise Exception("boom")

for index in range(10):
    gt = eventlet.spawn(target, index)
    threads[index] = gt

eventlet.sleep()

del threads
while gc.collect():
    pass

print("Foo count: ", objgraph.count('Foo'))
print("GreenThread count: ", objgraph.count('eventlet.greenthread.GreenThread'))
``` 

I would expect there to be exactly 1 `GreenThread` (the main thread) and 0 `Foo` objects. Instead there are 6 and 5 -- one extra for each thread that threw an exception.

Calling `gc.get_referrers` from one of the leftover `Foo` objects led me to this line: https://github.com/eventlet/eventlet/blob/54be7593ab8a35c9c80683c6d1af0adbe7763fa4/eventlet/greenthread.py#L220

Is this a bug, or is it unreasonable for a GreenThread to hold a reference to itself?

There is another example at https://gist.github.com/mattbennett/50509277a33c43658fe3fffa445f4a0a#file-growing-py where the number of objects in memory keeps growing.

temoto:
This is complicated.

First, thanks for reference to objgraph, I didn't know about that library.
I appreciate hard work digging to the root of issue.

Now as you found, the exception (or a simple value returned by greenthread target function) is sent into `GreenThread._exit_event`, an `Event` object used to wait for greenthread to finish and retrieve result/error.

And this is where things get tricky, you are not supposed to keep references to "self" greenthread in its result/error, because that creates live ref cycle event-gt. But obviously, with every exception there would be such reference.

Maybe good solution here is to travel exception object and clean references to greenthread.

If somebody have better ideas, please share.

temoto:
Extracting `GreenThread.main` method as module function helps with leakage of Foo objects.
But `GreenThread` objects still leak.

kashirin-alex:
I have made a test,

- it is for python2.7 python3.7:
('Foo count: ', 5)
('GreenThread count: ', 6)

[...]

PR #524, opened 2018-09-12
Issue 408 fix, Memory leak was generated from Event

kashirin-alex:
#408, fix for MemLeak on event exceptions  …
changes event.py:
  - self._result and self._exc should not be class's static attributes (the Values were passing between the different Event instances)
  - self._exc  should not store exception trace, instead if exception occurred True (references kept them self, causing GC not to collect, Event has _exc and _exc has trace frames object in touched)
  - self._exc  do not need to store exception trace, the same exception is called with the schedule_call_global for self._waiters timers
  - no need to throw self._exc on wait, it is thrown on the switch

  changes greenthread.py: (adjustments)
   - use of super for init inherited class
   - main method, try send, except send_exception and raise, and finally _resolve_links

#408, Test for fix for MemLeak on event exceptions  …
 - added test_no_mem_leaks to event_test.py
 - added tox.ini dependencies objgraph==3.4.0

### DIFF
--- a/eventlet/event.py
+++ b/eventlet/event.py
@@ -38,10 +38,10 @@ class Event(object):
     >>> evt.wait()
     4
     """
-    _result = None
-    _exc = None
 
     def __init__(self):
+        self._result = None
+        self._exc = None
         self._waiters = set()
         self.reset()
 
@@ -128,8 +128,6 @@ class Event(object):
                 return result
             finally:
                 self._waiters.discard(current)
-        if self._exc is not None:
-            current.throw(*self._exc)
         return self._result
 
     def send(self, result=None, exc=None):
@@ -161,13 +159,14 @@ class Event(object):
         """
         assert self._result is NOT_USED, 'Trying to re-send() an already-triggered event.'
         self._result = result
-        if exc is not None and not isinstance(exc, tuple):
-            exc = (exc, )
-        self._exc = exc
-        hub = hubs.get_hub()
+        if exc is not None:
+            if not isinstance(exc, tuple):
+                exc = (exc, )
+            self._exc = True
+
+        schedule_call_global = hubs.get_hub().schedule_call_global
         for waiter in self._waiters:
-            hub.schedule_call_global(
-                0, self._do_send, self._result, self._exc, waiter)
+            schedule_call_global(0, self._do_send, self._result, exc, waiter)
 
     def _do_send(self, result, exc, waiter):
         if waiter in self._waiters:
--- a/eventlet/greenthread.py
+++ b/eventlet/greenthread.py
@@ -167,7 +167,7 @@ class GreenThread(greenlet.greenlet):
     """
 
     def __init__(self, parent):
-        greenlet.greenlet.__init__(self, self.main, parent)
+        super(GreenThread, self).__init__(self.main, parent)
         self._exit_event = event.Event()
         self._resolving_links = False
         self._exit_funcs = None
@@ -218,13 +218,11 @@ class GreenThread(greenlet.greenlet):
 
     def main(self, function, args, kwargs):
         try:
-            result = function(*args, **kwargs)
+            self._exit_event.send(function(*args, **kwargs))
         except:
             self._exit_event.send_exception(*sys.exc_info())
-            self._resolve_links()
             raise
-        else:
-            self._exit_event.send(result)
+        finally:
             self._resolve_links()
 
     def _resolve_links(self):
--- a/tests/event_test.py
+++ b/tests/event_test.py
@@ -103,3 +103,38 @@ def test_wait_timeout_exceed():
         td = eventlet.hubs.get_hub().clock() - t1
         assert not result
         assert td >= delay
+
+
+def test_no_mem_leaks():
+
+    import objgraph
+    import gc
+    import sys
+    eventlet.hubs.get_hub().g_prevent_multiple_readers = False
+
+    threads = {}
+
+    class Foo(object):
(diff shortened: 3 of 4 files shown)


kashirin-alex:
Sorry, Unable to find the Fix.
-- the actual del threads of the test's ref count need to be possible
big issues are seen without the throw for a wait() with a results been an exception.

eventlet/greenthread.py
```diff
@@ -167,7 +167,7 @@ class GreenThread(greenlet.greenlet):
     """
 
     def __init__(self, parent):
-        greenlet.greenlet.__init__(self, self.main, parent)
+        super(GreenThread, self).__init__(self.main, parent)
```
jstasiak: Also seems unrelated to the issue at hand

tests/event_test.py
```diff
@@ -103,3 +103,46 @@ def test_wait_timeout_exceed():
         td = eventlet.hubs.get_hub().clock() - t1
         assert not result
         assert td >= delay
+
+
+def test_no_mem_leaks():
+
+    import objgraph
+    import gc
+    import sys
```
jstasiak: Those can be imported at the module level just fine, why import here?

tests/event_test.py
```diff
@@ -103,3 +103,46 @@ def test_wait_timeout_exceed():
...
+    for index in range(10):
+        gt = eventlet.spawn(target, index)
+        threads[index] = gt
+
+    eventlet.sleep()
+    threads.clear()
+    while gc.collect():
+        pass
+
+    eventlet.hubs.get_hub().set_timer_exceptions(True)
```
jstasiak: We have to make sure this is executed even if this test fails earlier on.

tox.ini
```diff
@@ -67,6 +67,7 @@ deps =
     py{27,34}-{selects,poll,epolls}: pyopenssl==17.3.0
     setuptools==38.5.1
     {selects,poll,epolls}: pyzmq==17.0.0
+    objgraph==3.4.0
```
jstasiak: This is an unnecessarily heavy dependency for what's being done here, we can count the amount of objects of a specific type just fine without it.

tests/event_test.py
```diff
@@ -103,3 +103,46 @@ def test_wait_timeout_exceed():
         td = eventlet.hubs.get_hub().clock() - t1
         assert not result
         assert td >= delay
+
+
+def test_no_mem_leaks():
```
jstasiak: This is not a test that tests the `Event` class, so most likely the code needs to be moved to another file. `greenthread_test.py` seems appropriate after a quick glance if it's actually `GreenThread` that's gonna be changed.

kashirin-alex: it is seen from using greenthread,
while the issue is generated in the Event instance
instance of Event is keeping in it's self._exc the frames trace including a greenthead instance/s.
It could have been anything else using Event

eventlet/event.py
```diff
@@ -38,10 +38,10 @@ class Event(object):
     >>> evt.wait()
     4
     """
-    _result = None
-    _exc = None
 
     def __init__(self):
+        self._result = None
+        self._exc = None
```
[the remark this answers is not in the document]
kashirin-alex: At this point it is the only thing that can cause a one not garbage collected object.

tox.ini
kashirin-alex: Will be great to know which method can replace the objgraph.count('eventlet.greenthread.GreenThread')

kashirin-alex:
Yes, the follow-up commits came out of nowhere, as there was more to the issue and initial checks were not sufficient although no issues have been seen for a while now (different type of occurrences/usages)
At this point, I don't yet know how to fix this one:
 - Instance need to preserve state, while it would include instances that need to be deleted/collected. 
 - One thing would change the issue, it is `wait `to always return results (None for exception), and a call on `wait `after `send` been issued, it is equivalent to the result with exception that is `None`.

tests/event_test.py
jstasiak: That's fair, I misread the code.

tox.ini
jstasiak:
```python
>>> import gc
>>> import objgraph
>>> 
>>> class X:
  2     pass
>>> 
>>> objgraph.count('X')
0
>>> sum(1 for o in gc.get_objects() if isinstance(o, X))
0
>>> x = X()
>>> objgraph.count('X')
1
>>> sum(1 for o in gc.get_objects() if isinstance(o, X))
1
```

kashirin-alex: OK, thanks,
I'll keep it updated with a solution to overcome/change the use of self._exc (if found of-course)

kashirin-alex: it might should be a method of the tests/__init__.py, and it has already gc imported
```python 
def check_ref_count(Instance, expected=0):
    c = sum(1 for o in gc.get_objects() if isinstance(o, Instance))
    assert c == expected, \
        "%s ref expected: %d counted: %d " % (Instance, expected, c)
```

kashirin-alex: or it can be better with a `with statement`
```python 
def check_ref_count(Instance, expected=0):
    try:
        pre_c = sum(1 for o in gc.get_objects() if isinstance(o, Instance))    
    finally:
        c = sum(1 for o in gc.get_objects() if isinstance(o, Instance))
        assert c-pre_c == expected, \
            "%s ref pre-counted: %d expected: %d \ counted: %d "
             % (Instance, pre_c, expected, c)
```
and work with:
```python 
with check_ref_count(Instance):
    #code 
```

PR #527, opened 2018-10-05
Fix for case where db_pool connections can be lost.

erickj00001:
I've encountered a case where **eventlet.db_pool** can lose a connection even though it was returned using put(), and after that, all greenthreads will wait indefinitely for a connection (assuming max_size=1, or if it happens enough times).

The problem occurs when:
(1) A greenthread was waiting for a connection from the pool
(2) Another thread calls put() to return a connection to the pool, but it has expired, so the first thread will be allowed to create a new connection
(3) The create() method raises an exception that doesn't inherit from Exception, such as **eventlet.Timeout**

Note that **Pool.get** uses "except:" instead of "except: Exception" to make sure all exceptions from create() are accounted for, but **BaseConnectionPool.get** incorrectly uses "except: Exception" in the case where a returned connection was found to be unusable.

Here's a standalone test that demonstrates the problem, with comments that explain in more detail:
https://github.com/erickj00001/test-scripts/blob/c1f041b4428b01cf62c6fb68718d5d30f95224ed/db_pool_test.py#L1-L108

Output:
```
[ 0.000] [conn-1] Opening (delay 0)...
[ 0.000] [conn-1] Opened
[ 0.000] (main) Checked out connection conn-1
[ 0.000] (test1) Requesting connection
[ 6.007] (main) Returning connection conn-1
[ 6.007] [conn-1] Closed
[ 6.007] [conn-2] Opening (delay 10)...
[ 6.007] Waiting for test1...
[ 8.003] [conn-2] Connection aborted!
[ 8.003] (test1) Timed out
[ 8.003] test1 finished
[ 9.004] (test2) Requesting connection
[17.005] (test2) Timed out
[17.005] --- TEST FAILED ---
```
After changing the "except Exception:" to "except:" in **BaseConnectionPool.get**, the test passes:

```
[ 0.000] [conn-1] Opening (delay 0)...
[ 0.000] [conn-1] Opened
[ 0.000] (main) Checked out connection conn-1
[ 0.000] (test1) Requesting connection
[ 6.005] (main) Returning connection conn-1
[ 6.005] [conn-1] Closed
[ 6.006] [conn-2] Opening (delay 10)...
[ 6.006] Waiting for test1...
[ 8.003] [conn-2] Connection aborted!
[ 8.003] (test1) Timed out
[ 8.003] test1 finished
[ 9.004] (test2) Requesting connection
[ 9.004] [conn-3] Opening (delay 0)...
[ 9.004] [conn-3] Opened
[ 9.004] (test2) Checked out connection conn-3
[ 9.004] --- TEST PASSED ---
```


erickj00001:
It works fine for me in Python 3.

Anyway, it raises the exception again immediately (after decrementing a counter).  It's just important that the counter get decremented there.  The base class version of get() uses "except:", like I said.  Why do you suppose it's correct there, but not in the derived class version?

temoto:
I said nothing about `except:` in base class. Likely it is not appropriate everywhere except hub and greenthread code. Last time I looked at fixing it, proper syntax broke too many tests and I postponed it.

erickj00001:
Please help me understand here.  Are you saying that there isn't a problem with the code as is?  I believe my test script adequately demonstrates that there is a problem, but if you disagree then please let me know your reasoning.

Or, are you just saying that you don't like the syntax of the proposed solution?  We can change it to use finally: instead if you prefer, something like this:
```python
try:
    conn = self.create()
finally:
    if conn is None:
        # unconditionally increase the free pool because
        # even if there are waiters, doing a full put
        # would incur a greenlib switch and thus lose the
        # exception stack
        self.current_size -= 1
```

You mention that you want the test script checked in as part of the patch.  Do I understand that to mean you are not able to try it as presented, and you will only run it if I integrate it into the unit test framework?  Again, this is not to complain -- I just wish to understand the requirements.

temoto:
Clearly there are problems with current code, arguably the biggest one is `class Timeout(BaseException)`. Then `except:` catching things it shouldn't.

Sorry for ambiguity.

My points are:
- (important) since you have test, it should be part of patch
- because all tests pass is requirement before merge (it will guard against recurring problem in same code forever)
- Eventlet doesn't rely on humans running tests, it's done automatically with help of TravisCI
- if possible, don't use `except:`
  - if you target particular exception types, write them
  - if you target all exceptions, write `except Exception`, but it doesn't work with `eventlet.Timeout`, so if it works with everything except X - that's a problem with X, so it should be `except (Exception, Timeout)`
  - but if you don't need exception case at all and could do with `try/finally` - of course use it

erickj00001:
Ok, I've added my script as a unit test (as part of the existing db_pool_test.py).  I reduced the sleep delays to the minimum, so as not to slow down the testing unnecessarily.

I've also updated the proposed fix so it uses finally: instead of except:.

PR #530, opened 2018-11-30
Added green nanomsg patcher

wtfuzz:
This adds a green nanomsg monkeypatcher. I've tested this in a Flask application using eventlet, and Flask-SocketIO with a background thread.

Example of a subscriber.

```python
from eventlet.green.nanomsg import *

sock = Socket(SUB)
sock.connect("tcp://localhost:6001")
sock.set_string_option(SUB, SUB_SUBSCRIBE, '')

while True:
  msg = sock.recv()
  print(msg)
```

### DIFF
--- a/eventlet/green/nanomsg.py
+++ b/eventlet/green/nanomsg.py
@@ -6,11 +6,11 @@ found in :mod:`nanomsg-python <nanomsg>` to be non blocking.
 __nanomsg__ = __import__('nanomsg')
 from eventlet.patcher import slurp_properties
 from eventlet.hubs import trampoline
-from nanomsg import wrapper
+from errno import EAGAIN
 
 __nanomsg__.__all__ = ['wrapper', 'NanoMsgError', 'NanoMsgAPIError', 'Device', 'Socket']
 
-for name, value in wrapper.nn_symbols():
+for name, value in __nanomsg__.wrapper.nn_symbols():
     if name.startswith('NN_'):
         name = name[3:]
     __nanomsg__.__all__.append(name)
@@ -24,33 +24,32 @@ class Socket(__nanomsg__.Socket):
     def recv(self, buf=None, flags=0):
         """Recieve a message."""
 
-        # Wait for the receive file descriptor using eventlet
+        flags |= __nanomsg__.DONTWAIT
+
         while True:
-
-            # Call the eventlent hub trampoline function to wait for a notification on the
-            # receive file descriptor.
-            trampoline(self.recv_fd, read=True)
-
-            # Don't allow nn_recv() to block
-            flags = flags | __nanomsg__.DONTWAIT
-
             if buf is None:
-                rtn, out_buf = wrapper.nn_recv(self.fd, flags)
+                rtn, out_buf = __nanomsg__.wrapper.nn_recv(self.fd, flags)
             else:
-                rtn, out_buf = wrapper.nn_recv(self.fd, buf, flags)
+                rtn, out_buf = __nanomsg__.wrapper.nn_recv(self.fd, buf, flags)
 
-            # Return if we received a valid message
-            if(rtn > 0):
+            if rtn < 0:
+                if __nanomsg__.wrapper.nn_errno() == EAGAIN:
+                    trampoline(self.recv_fd, read=True)
+            elif rtn >= 0:
                 return bytes(memoryview(out_buf))[:rtn]
+            else:
+                raise __nanomsg__.NanoMsgAPIError()
 
     def send(self, msg, flags=0):
         """Send a message"""
 
         while True:
             trampoline(self.send_fd, write=True)
-            ret = wrapper.nn_send(self.fd, msg, flags | __nanomsg__.DONTWAIT)
-            if(ret > 0):
+            ret = __nanomsg__.wrapper.nn_send(self.fd, msg, flags | __nanomsg__.DONTWAIT)
+            if ret > 0:
                 break
+            else:
+                raise __nanomsg__.NanoMsgAPIError()
 
     def poll(in_sockets, out_sockets, timeout=-1):
         raise NotImplementedError("poll is not implemented in the nanomsg eventlet wrapper")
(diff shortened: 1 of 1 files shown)


eventlet/green/nanomsg.py
```diff
@@ -0,0 +1,56 @@
+# coding: utf-8
+"""The :mod:`nanomsg` module wraps the :class:`Socket`
+found in :mod:`nanomsg-python <nanomsg>` to be non blocking.
+"""
+
+__nanomsg__ = __import__('nanomsg')
+from eventlet.patcher import slurp_properties
+from eventlet.hubs import trampoline
+from nanomsg import wrapper
```
nat-goodspeed: Shouldn't this be based on `__nanomsg__`?

eventlet/green/nanomsg.py
```diff
@@ -0,0 +1,56 @@
...
+            # receive file descriptor.
+            trampoline(self.recv_fd, read=True)
+
+            # Don't allow nn_recv() to block
+            flags = flags | __nanomsg__.DONTWAIT
+
+            if buf is None:
+                rtn, out_buf = wrapper.nn_recv(self.fd, flags)
+            else:
+                rtn, out_buf = wrapper.nn_recv(self.fd, buf, flags)
```
nat-goodspeed: I'm sorry, I'm unfamiliar with nanomsg and its implementation, but it's atypical for a Python function to be "overloaded" in such a way as to change the meaning of a positional argument.

Is it really true that if you pass `nanomsg.wrapper.nn_recv()` two positional arguments, the second one is `flags`, but if you pass three, the second one is `buf`?

eventlet/green/nanomsg.py
```diff
@@ -0,0 +1,56 @@
...
+
+__nanomsg__.__all__ = ['wrapper', 'NanoMsgError', 'NanoMsgAPIError', 'Device', 'Socket']
+
+for name, value in wrapper.nn_symbols():
+    if name.startswith('NN_'):
+        name = name[3:]
+    __nanomsg__.__all__.append(name)
+
+__patched__ = ['Socket']
+slurp_properties(__nanomsg__, globals(), ignore=__patched__)
```
jstasiak: Do those two blocks above mutate global state on import?

temoto:
- `if(ret > 0):` is not Python, should be `if ret > 0:` PEP8 checker should've found this, don't know why.
- it's better to try optimistically recv/send first and trampoline on EAGAIN or similar state.

eventlet/green/nanomsg.py
wtfuzz: I agree this is very odd, but python-nanomsg does this internally:

```python
def nn_recv(socket, *args):
    "receive a message"
    if len(args) == 1:
        flags, = args
        pointer = ctypes.c_void_p()
        rtn = _nn_recv(socket, ctypes.byref(pointer), ctypes.c_size_t(-1),
                       flags)
        if rtn < 0:
            return rtn, None
        else:
            return rtn, _create_message(pointer.value, rtn)
    elif len(args) == 2:
        msg_buf, flags = args
        mv_buf = memoryview(msg_buf)
        if mv_buf.readonly:
            raise TypeError('Writable buffer is required')
        rtn = _nn_recv(socket, ctypes.addressof(msg_buf), len(mv_buf), flags)
        return rtn, msg_buf
```

eventlet/green/nanomsg.py
wtfuzz: They do. Unfortunately nanomsg doesn't provide an ``__all__``, which is required by ``slurp_properties()``, so the monkeypatcher builds one for it.

wtfuzz:
> * `if(ret > 0):` is not Python, should be `if ret > 0:` PEP8 checker should've found this, don't know why.

Strange. flake8 didn't catch it here locally.

> * it's better to try optimistically recv/send first and trampoline on EAGAIN or similar state.

I've pushed these changes. Great advice, thanks.

eventlet/green/nanomsg.py
wtfuzz: I've submitted a PR to tonysimpson/nanomsg-python#71 to add an ``__all__`` variable to the nanomsg module.

eventlet/green/nanomsg.py
nat-goodspeed: Wow! A reader of your `Socket.recv()` method would certainly benefit from an explanatory comment. You don't have to paste their code, just explain that they really do take different branches depending on the number of passed arguments.

eventlet/green/nanomsg.py
nat-goodspeed: It would be even better if your code defers to their `__all__` once it exists upstream.

eventlet/green/nanomsg.py
```diff
@@ -0,0 +1,55 @@
...
+"""The :mod:`nanomsg` module wraps the :class:`Socket`
+found in :mod:`nanomsg-python <nanomsg>` to be non blocking.
+"""
+
+__nanomsg__ = __import__('nanomsg')
+from eventlet.patcher import slurp_properties
+from eventlet.hubs import trampoline
+from errno import EAGAIN
+
+__nanomsg__.__all__ = ['wrapper', 'NanoMsgError', 'NanoMsgAPIError', 'Device', 'Socket']
```
temoto, line 11:
```suggestion
if not hasattr(__nanomsg__, "__all__"):
    __nanomsg__.__all__ = ['wrapper', 'NanoMsgError', 'NanoMsgAPIError', 'Device', 'Socket']
```

eventlet/green/nanomsg.py
```diff
@@ -0,0 +1,56 @@
...
+    def send(self, msg, flags=0):
+        """Send a message"""
+
+        while True:
+            trampoline(self.send_fd, write=True)
+            ret = __nanomsg__.wrapper.nn_send(self.fd, msg, flags | __nanomsg__.DONTWAIT)
+            if ret > 0:
+                break
+            else:
+                raise __nanomsg__.NanoMsgAPIError()
```
temoto, line 53:
```suggestion
        flags |= __nanomsg__.DONTWAIT

        while True:
            ret = __nanomsg__.wrapper.nn_send(self.fd, msg, flags)
            if ret >= 0:
                return ret
            if __nanomsg__.wrapper.nn_errno() == EAGAIN:
                trampoline(self.send_fd, write=True)
                continue
            raise __nanomsg__.NanoMsgAPIError()
```

itamarst:
We haven't quite nailed down the new policy (#835), but there's a very good chance no new features will be accepted going forward. If that happens, unfortunately this PR won't be accepted.

Issue #371, opened 2017-01-04
SSL.py recursion crash in (eventlet > 0.17.4) and python 3.6


Renmusxd:
Using python 3.6 with the following code causes an infinite recursion of python super calls leading to a crash

```
import eventlet
eventlet.monkey_patch()

import os
import socket
from flask import Flask, render_template, request, Response, send_file
from flask import make_response
from flask_socketio import SocketIO
from OpenSSL import SSL, crypto

'''
requirements.txt
cffi==1.9.1
click==6.6
cryptography==1.7.1
enum-compat==0.0.2
eventlet==0.20.1
Flask==0.12
Flask-SocketIO==2.8.2
greenlet==0.4.11
idna==2.2
itsdangerous==0.24
Jinja2==2.8.1
MarkupSafe==0.23
pyasn1==0.1.9
pycparser==2.17
pyOpenSSL==16.2.0
python-engineio==1.1.0
python-socketio==1.6.2
six==1.10.0
Werkzeug==0.11.15
'''

app = Flask(__name__)
app.config['SECRET_KEY'] = 'secret!'
async_mode = None
socketio = SocketIO(app, async_mode=async_mode)
def create_self_signed_cert(certfile, keyfile, certargs, cert_dir="."):
    C_F = os.path.join(cert_dir, certfile)
    K_F = os.path.join(cert_dir, keyfile)
    if not os.path.exists(C_F) or not os.path.exists(K_F):
        k = crypto.PKey()
        k.generate_key(crypto.TYPE_RSA, 1024)
        cert = crypto.X509()
        cert.get_subject().C = certargs["Country"]
        cert.get_subject().ST = certargs["State"]
        cert.get_subject().L = certargs["City"]
        cert.get_subject().O = certargs["Organization"]
        cert.get_subject().OU = certargs["Org. Unit"]
        cert.get_subject().CN = 'Example'
        cert.set_serial_number(1000)
        cert.gmtime_adj_notBefore(0)
        cert.gmtime_adj_notAfter(315360000)
        cert.set_issuer(cert.get_subject())
        cert.set_pubkey(k)
        cert.sign(k, 'sha1')
        open(C_F, "wb").write(crypto.dump_certificate(crypto.FILETYPE_PEM, cert))
        open(K_F, "wb").write(crypto.dump_privatekey(crypto.FILETYPE_PEM, k))

CERT_FILE = "cert.pem"
KEY_FILE = "key.pem"
[...]

PR #531, opened 2018-12-09
Fix compatibility with Python 3.7 ssl.SSLSocket 

luanjunyi:
## What

This PR fixes the issue in https://github.com/eventlet/eventlet/issues/526. There was [another PR](https://github.com/eventlet/eventlet/pull/506) that tried to fix the incompatibility with ssl module in Python 3.7, but the hack doesn't work with Monkey patch.

The reason of the break is in ssl of Python 3.7, SSLSocket can't be initialized from __init__, but have to be initialized from it's _create method. In _create code, it accesses **SSLSocket**'s parent class via **super**. If 'SSLSocket' is patched at the execution time of _create, that symbol is in fact **GreenSSLSocket** class, so **super** will return wrong class.

## More details and context

In python 3.7's ssl.py, there're SSLContext and SSLSocket symbols. These symbols are accessible globals in _create method of SSLSocket(the above method that is used to create SSLSocket) and wrap_socket method.

When monkey_patch is called, the module from sys.modules['ssl'] is patched with GreenSSLConext and GreenSSLSocket. The monkey patch is done by setattr builtin method. This will eventually modify  ssl.\_\_dict__. This is the same dict as _create.\_\_globals\_\_. I found this out by experimenting in iPython terminal and looking at cPython's implementation(start [here](https://github.com/python/cpython/blob/master/Python/bltinmodule.c#L1068)). This fact may imply fatal flaws of the patching method used in Eventlet. But it's clearly beyond the scope of this PR.

At run time, the only way to get a SSLSocket object is by calling _create. In _create it reference _create.__globals__['SSLContext']. This symbol must be pointing to the original SSLContext when _create is executed. Because the monkey patch changes _create.__globals__['SSLContext'] to point to GreenSSLContext, the _create code was not executed correctly, hence the [issue](https://github.com/eventlet/eventlet/issues/526).

The fix is hard however. The best solution I found so far is to swap back SSLContext and SSLSocket in the two __globals__ of _create and ssl.wrap_socket method during the construction and then put back the objects as before the swapping. This will make the code thread-unsafe. As during the swapping period, other threads may in theory reference 'wrong' object. For example, patched ssl.SSLContext.wrap_socket will return a non-green SSLSocket during the swapping. The root cause is the __globals__ referenced by _create is same dict used by the imported module object's getattr and setattr methods. This is CPython's implementation and I didn't find a way to modify this behavior.

Another way considered is to pop ssl from sys.modules and reimport it to get the original _create function. However, the result returned from the new _create will have different object layout and can't assign it's __class__ to GreenSSLSocket, resulting this error:

    TypeError: __class__ assignment: 'GreenSSLSocket' object layout differs from 'SSLSocket'

This can be fixed by change the SSLSocket attribute in the newly import ssl to _original_sslsocket, but then the patched ssl will be back to the game, making _create the patched one. So this approach doesn't work.

## How

The fix in https://github.com/eventlet/eventlet/pull/506 didn't work for three reasons:

* It used self = _create() in __init__ of GreenSSLSocket. This doesn't initialize the GreenSSLSocket object because this statement just bind self to the object returned from _create an the GreenSSLSocket is not affected from there. 

* SSLSocket and SSLContext is monkey patched before calling _create. As described in **what** part.

* The potential race condition pointed out in the PR.

The fix approach is 

* override __new__ of GreenSSLSocket and call SSLSocket._create to create the GreenSSLSocket object

* Before calling _create, swap back the original **SSLSocket** class in green/ssl.py

## Beyond this fix

As I mentioned above, functions use their __globals__ to locate the global variables in the module it's defined and this __globals__ is same as the module's __dict__ which get modified in Eventlet patch. This is intended most of the time, but not always, especially for the functions defined in the patched modules.

For instance, GreenSocket will delegate the accept method to socket.socket's accept method where 'socket.set_nonblocking' will be called. Here the 'socket' symbol is a global and it may be already patched to be GreenSocket thus setblocking will not work as intended. Indeed, if we monkey_patch on tests/api_test.py, the test will fail due to this issue. Fixing that is non trivial since other tests will be affected. I would propose to fix those in other PRs.

@Dormouse759 @jstasiak


luanjunyi:
Still figuring out how to add unit testing. But let me know if this fix makes sense.

luanjunyi:
socket initialization is not correct which breaks the test. Let me try fix that

eventlet/green/ssl.py
```diff
@@ -58,34 +57,48 @@ def __init__(self, sock, keyfile=None, certfile=None,
...
-            finally:
-                # Unpatch
-                _original_sslsocket.__new__ = orig_new
+            import imp
+            need_swap_ssl = 'ssl' in sys.modules
+            if need_swap_ssl:
+                # In case 'ssl' is greened, temporarily swap back original SSLSocket
+                # and SSLContext because Python 3.7 use this exact Name to locate the
+                # correct parent class in SSLSocket._create
+                original_sslcontext = _original_sslcontext
```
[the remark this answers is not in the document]
luanjunyi: Fixed

nat-goodspeed:
@luanjunyi, from your remark "We need to investigate" I assume you're still working on this; that is, it's not yet ready to merge.

luanjunyi:
@nat-goodspeed Yes, I don't think we should merge now. I didn't fully understand the imp locking mechanism and what's the equivalent in importlib. I think we need to get enough knowledge about it before merging. Now that all tests are passed, let me do the following:

* Address comments from @nat-goodspeed. Especially the imp's deprecation issue.
* Add unit tests for the code change
* Add unit tests for the issues I found during fixing this issue

After that, we can have a final review of this PR. Then we can merge it. Let me know if this is sound plan.

nat-goodspeed:
I approve of your plan.

eventlet/green/ssl.py
```diff
@@ -40,6 +39,63 @@ class GreenSSLSocket(_original_sslsocket):
     settimeout(), and to close/reopen the connection when a timeout
     occurs at an unexpected juncture in the code.
     """
+
```
nat-goodspeed: I respect your decision to use a context manager instead of `try` / `finally`. However, I have several things to say about `SwapBackOriginal`.

* Please don't nest it within `GreenSSLSocket`. Even though the distinction between `GreenSSLSocket` methods and `SwapBackOriginal` methods is clear to Python, it's not as clear to a human reader.
* I generally prefer to use a generator function decorated with [`contextlib.contextmanager`](https://docs.python.org/2/library/contextlib.html#contextlib.contextmanager) to express all but complex context managers -- I find that more readable.
* I would like your context manager even more if it were parameterized with the module name and (e.g.) a `**kwds` parameter with which to specify the symbols to save/restore and the values with which to (temporarily) replace them. One of the benefits of using a `contextlib.contextmanager` function is that, of course, function parameters are straightforward.

It's possible that a generalized "temporarily use these instead" context manager could be useful in more places within `eventlet`, though I'm content to leave it in this module for now.

eventlet/green/ssl.py
```diff
@@ -40,6 +39,63 @@ class GreenSSLSocket(_original_sslsocket):
...
+    class SwapBackOriginal:
+        SYMBOLS = ['SSLSocket', 'SSLContext']
+
+        def __init__(self):
+            self.need_swap = False
+            self.temp_objs = {}
```
nat-goodspeed: I do like that you generalized `SYMBOLS` and `self.temp_objs`.

eventlet/green/ssl.py
```diff
@@ -40,6 +39,63 @@ class GreenSSLSocket(_original_sslsocket):
...
+                )
+            ret.keyfile = keyfile
+            ret.certfile = certfile
+            ret.cert_reqs = cert_reqs
+            ret.ssl_version = ssl_version
+            ret.ca_certs = ca_certs
+            ret.__class__ = GreenSSLSocket
+            return ret
+        else:
+            return super(GreenSSLSocket, cls).__new__(cls)
```
nat-goodspeed: If one branch of an `if` / `else` is much shorter (fewer lines of code) than the other, I generally prefer to express the condition so as to put the short branch with `if` and the long branch with `else`. That way, as you read the top of each branch, the condition that brought you there is just a glance away.

eventlet/green/ssl.py
```diff
@@ -58,34 +57,48 @@ def __init__(self, sock, keyfile=None, certfile=None,
...
-                # Unpatch
-                _original_sslsocket.__new__ = orig_new
+            import imp
+            need_swap_ssl = 'ssl' in sys.modules
+            if need_swap_ssl:
+                # In case 'ssl' is greened, temporarily swap back original SSLSocket
+                # and SSLContext because Python 3.7 use this exact Name to locate the
+                # correct parent class in SSLSocket._create
+                original_sslcontext = _original_sslcontext
+                imp.acquire_lock()
```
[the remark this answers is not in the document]
nat-goodspeed: I confess that I know no more about the `imp` lock than what I quoted above. I agree that, absent a definite answer, it's safer to retain the locking than to remove it.

That said, is it necessary to hold the lock during the body of the `with` statement? Does `ssl.wrap_socket()` aka `_original_wrap_socket()` patch anything in `sys.modules`? Would it be just as safe to acquire the lock while patching, release it, execute the body of the `with` statement, reacquire the lock to restore the patched values and release it again?

I guess the problem case would be if some other thread imports something else that depends on `ssl` during that time: it would see the original `ssl` module rather than the patched one. Is that a fair statement? If so, it appears that holding the lock for the duration is still valuable. Yes?

eventlet/greenio/base.py
```diff
@@ -145,7 +145,7 @@ def __init__(self, family=socket.AF_INET, *args, **kwargs):
         except AttributeError:
             self._timeout = socket.getdefaulttimeout()
 
-        if should_set_nonblocking:
+        if should_set_nonblocking and fd.fileno() != -1:
```
jstasiak: A comment explaining the second part of the condition would be helpful here.

eventlet/green/ssl.py
```diff
@@ -24,6 +22,7 @@
 
 _original_sslsocket = __ssl.SSLSocket
 _original_wrap_socket = __ssl.wrap_socket
+_original_sslcontext = __ssl.SSLContext if hasattr(__ssl, 'SSLContext') else None
```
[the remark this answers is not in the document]
jstasiak: `getattr` is also an option

eventlet/greenio/base.py
luanjunyi: Done

eventlet/green/ssl.py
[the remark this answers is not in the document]
luanjunyi: Added more details to the background section in the PR description.

luanjunyi:
@jstasiak @nat-goodspeed I've make more intensive research of this issue. I've put what I found in the details and background section in this PR's description. PTAL and let me know your thoughts.

eventlet/green/ssl.py
luanjunyi: done

nat-goodspeed:
Thank you for your comprehensive writeup in the PR description!

I skimmed it quickly, so I may have missed something -- but when you mention the difficulty of having a function's `__globals__` be the same `dict` as its containing module's `__dict__`, have you looked at (e.g.) `eventlet.green.subprocess`'s use of `types.FunctionType` to create a new function object with the original code but replaced globals?

luanjunyi:
@nat-goodspeed Thanks for sharing the tips of using FunctionType! It's a great feature I didn't know. However, I did some research and found that it can't solve this problem. To create new SSLSocket, the call chains is

ssl.wrap_socket -> ssl.SSLContext.verify_mode -> ...

Indeed we can create a function that is a clone of ssl.wrap_socket. In this clone, all globals are preserved. However, we can't control the globals in verify_mode and other functions downstream. Those functions' globals are still 'polluted' by the monkey patch. And we have to swap them back as we did in this PR.

luanjunyi:
@nat-goodspeed @jstasiak bump

nat-goodspeed:
Oh sorry, didn't realize the ball was back in our court. Will scan through once more ... please be patient...

luanjunyi:
@nat-goodspeed friendly reminder

maxin93:
@luanjunyi Thanks for your modification, it do works for me.
But I found another problem that your changes might be conflict with the monkey_patch for socket in **eventlet/patcher.py**, it will raise **ssl.SSLWantReadError: The operation did not complete (read) (_ssl.c:2460)**
I got the test code from https://github.com/eventlet/eventlet/issues/371, by Renmusxd, and if I set the socket as False, it works. 
`eventlet.monkey_patch(socket = False)`

Another problem that bothers me is: Does Python3.7's SSL.py not support custom context, and also not support for loading encrypted keyfile?

Looking forward to your reply.

luanjunyi:
@maxin93 Yes this fix doesn't work with monkey_patch(). In fact, many features are broken with eventlet.monkey_patch(). I think this function should be removed and eventlet.import_patched() should be the only right way for monkey patching. Read on if you want to know why.

First some contexts. monkey_patch() swap out all system modules that contains blocking IO and replace them with eventlet's implementation with non-blocking IO. You call monkey_patch at the very beginning of your execution, likely in the first line of you __init__.py. This will MODIFY the behavior of original modules like socket and select. 

import_patched is far less obstructive. It temporarily swap out system modules containing blocking IO and replace with eventlet equivalents. Then import the module you wanna 'patch'. This way this module binds with the eventlet replacement instead of the original modules. After the import, all original modules are swapped back. So that only the patched module will use the modified version of the replaced system modules, but other modules and symbols are unaffected.

Sadly, as you can see both method is really hacking the system in a undocumented way. This will introduce bugs like this PR wanna address and another bug I'm fixing(https://github.com/eventlet/eventlet/pull/540). The good news is those kind of bug usually will cause a crash with minimal amount of program runtime.

I personally recommend using import_patched and never use monkey_patch(). Also make import_patched as the first line of you code execution if possible.

I hope this answers you question.

nat-goodspeed:
Would you consider (maybe as a separate PR) a documentation patch stating that monkey_patch() is deprecated, with the above explanation?
I don't know if there will come a time when it's safe to actually remove it, but we could consider a deprecation warning?

luanjunyi:
@nat-goodspeed Sure thing, happy to have a more thorough look at patcher and update code/document accordingly. I think it should be separate PRs.

maxin93:
@luanjunyi How could I create SSLContext with eventlet by loading encrypted keyfile?

maxin93:
Awesome, thx.

eventlet/green/ssl.py
```diff
@@ -40,59 +53,56 @@ class GreenSSLSocket(_original_sslsocket):
     settimeout(), and to close/reopen the connection when a timeout
     occurs at an unexpected juncture in the code.
     """
+    def __new__(cls, sock=None, keyfile=None, certfile=None,
+                server_side=False, cert_reqs=CERT_NONE,
+                ssl_version=PROTOCOL_SSLv23, ca_certs=None,
+                do_handshake_on_connect=True, *args, **kw):
+        if sys.version_info < (3, 7):
+            return super(GreenSSLSocket, cls).__new__(cls)
+        else:
```
nat-goodspeed: I'm a little uneasy about the fact that the default value for `sock` is `None`. I realize you copied that argument list from `ssl.SSLSocket.__init__()`, but I don't know under what circumstances we could actually arrive at this point with `sock is None`. Is that case supported by `ssl.SSLSocket`? Passing `None` to `GreenSocket` doesn't seem likely to be rewarding -- or trying to reference `None.fd`, etc.
I think what I'm asking is that you either defend against that case, or explain in comments why that won't happen.

eventlet/green/ssl.py
```diff
@@ -40,59 +53,56 @@ class GreenSSLSocket(_original_sslsocket):
     settimeout(), and to close/reopen the connection when a timeout
     occurs at an unexpected juncture in the code.
     """
+    def __new__(cls, sock=None, keyfile=None, certfile=None,
+                server_side=False, cert_reqs=CERT_NONE,
+                ssl_version=PROTOCOL_SSLv23, ca_certs=None,
+                do_handshake_on_connect=True, *args, **kw):
+        if sys.version_info < (3, 7):
```
kashirin-alex: suggestion,
all the conditions of `sys.version_info < (3, 7)` can be pre-defined with

`python
is_under_3_7 = sys.version_info < (3, 7)
`
in the global scope

eventlet/green/ssl.py
```diff
@@ -40,59 +53,56 @@ class GreenSSLSocket(_original_sslsocket):
...
+            return super(GreenSSLSocket, cls).__new__(cls)
```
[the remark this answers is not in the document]
luanjunyi: I'm sure you need to pass cls to __new__. See this example:
```
In [151]: SSLSocket.__new__()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-151-fc85ecb27b2b> in <module>()
----> 1 SSLSocket.__new__()

TypeError: _socket.socket.__new__(): not enough arguments

In [152]: SSLSocket.__new__(SSLSocket)
Out[152]: <ssl.SSLSocket fd=-1, family=AddressFamily.AF_UNSPEC, type=0, proto=0>
```

sock, keyfile should not be sent to __new__ because they are not used. You can do that, but there's actually no effect. It will confuse the maintainer. There are tons of great post explaining the difference between __init__ and __new__. They worth a good read if you have no idea how do __new__ and __init__ work, but feel free to let me know if you want my own understanding.

eventlet/green/ssl.py
luanjunyi: This is the orginal SSLSocket.__init__ signature. 

```
In [157]: ssl.SSLSocket.__init__
Out[157]: <function ssl.SSLSocket.__init__(self, sock=None, keyfile=None, certfile=None, server_side=False, cert_reqs=<VerifyMode.CERT_NONE: 0>, ssl_version=<_SSLMethod.PROTOCOL_TLS: 2>, ca_certs=None, do_handshake_on_connect=True, family=<AddressFamily.AF_INET: 2>, type=<SocketKind.SOCK_STREAM: 1>, proto=0, fileno=None, suppress_ragged_eofs=True, npn_protocols=None, ciphers=None, server_hostname=None, _context=None, _session=None)>
```

eventlet/green/ssl.py
luanjunyi: Done

luanjunyi:
@kashirin-alex @jstasiak Ping

eventlet/green/ssl.py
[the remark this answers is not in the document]
nat-goodspeed: > I'm sure you need to pass cls to `__new__`.

I'm sure of that too. What I'm suggesting is that I believe `super(GreenSSLSocket, cls).__new__()` binds `cls` to the callable returned by `super()` such that you shouldn't need to pass it *again* in the argument list.

But you're saying that this `__new__()` ignores its argument list -- which probably makes it harmless to pass `cls` redundantly.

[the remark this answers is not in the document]
luanjunyi: I read through the Googled results again and this is the recommended way of calling parent class's __new__. What's your suggested way of writing the statement? @nat-goodspeed

luanjunyi:
Removed the blank line. Other blank lines are required by PEP-8

luanjunyi:
@jstasiak Would you merge this? I've another project that depends on this fix.

licht1stein:
Hi, I'm still getting ```TypeError: wrap_socket() got an unexpected keyword argument '_context'``` when trying to use requests with eventlet in nameko. Is there something I can do to fix this?

When trying to use httpx instead of requests everything just hangs.

nat-goodspeed:
I've been (slowly) trying to hack through #526 myself in PR #565. You might take a look at that fork+branch. I do not yet have a solution that passes all tests, but the current state of that branch is that it permits my use of requests with monkey_patch(socket=True) to work.

mttmtt:
> Removed the blank line. Other blank lines are required by PEP-8

Is there currently an official evenlet version that solves the problem "TypeError: wrap_socket () got an unexpected keyword argument '_context"?  

Looking forward to reply as soon as possible, thank you!

mttmtt:
Is there currently an official evenlet version that solves the problem "TypeError: wrap_socket () got an unexpected keyword argument '_context"?

Looking forward to reply as soon as possible, thank you!

PR #532, opened 2018-12-13
wsgi: Catch and swallow IOErrors during discard()

tipabu:


nat-goodspeed:
I'm puzzled by your description "IOErrors during discard()" -- the code you added seems to be in `HttpProtocol.handle_one_response()`.

To be honest, I don't have the background to evaluate the benefit of this change. I'm happy that @charmster is pleased (though I'm also puzzled by his reference to Swift).

And yes, thank you for including a test for your new functionality.

tipabu:
Sorry for the original terseness; I think I had a few things dividing my attention when I submitted this.

> I'm puzzled by your description "IOErrors during discard()" -- the code you added seems to be in HttpProtocol.handle_one_response().

I'd meant it as "Catch and swallow [IOErrors during discard()]", rather than "[Catch and swallow IOErrors] during discard()" -- the trouble I'd see was this:

- client starts a chunked PUT, either using HTTP/1.1 or sending an explicit `Connecton: keep-alive`
- client probably even sends some data
- application wraps its read()s with Timeouts
- client is slow enough that the Timeout pops
- application sends back a `408 Request Timeout` without any further reads
- client sees the early response (because it has separate threads for input vs output), kills the transfer, and closes the connection

Now ideally, as soon as the application sends the 408, it could also include some indication to the WSGI server to say "this connection is dead to me, please just close it and clean it up" -- but I don't think there are any provisions for this in the spec short of raising an exception [which would trigger a 500](https://github.com/eventlet/eventlet/blob/v0.24.1/eventlet/wsgi.py#L581-L590) instead of the 408. We can't even include a `Connection: close` header in the response, because

> Applications and middleware are forbidden from using HTTP/1.1 "hop-by-hop" features or headers, any equivalent features in HTTP/1.0, or any headers that would affect the persistence of the client's connection to the web server. These features are the exclusive province of the actual web server, and a server or gateway **should** consider it a fatal error for an application to attempt sending them, and raise an error if they are supplied to `start_response()`.

So the application just sends the error response and hopes for the best. From the WSGI server's perspective, though, everything's humming along just fine: client sent a request, we passed it on to the application; application read however much of the input as was necessary to generate a response; we passed that response back to the client just fine. Since the client indicated that it might send more requests, let's keep serving!

The application didn't read everything, so it's on us to discard the rest of the body. But when we do, we start getting back empty byte strings, since the client hung up! This would be fine if it weren't a chunked transfer: we'd just pass that back up to `discard()` and it'd think it was done. But [reading chunked requests can raise `IOError`s](https://github.com/eventlet/eventlet/blob/v0.24.1/eventlet/wsgi.py#L186-L188) which won't be caught and we'll get a traceback like when the new test fails after backing out the changes to wsgi.py:

```
  File ".../eventlet/eventlet/hubs/hub.py", line 461, in fire_timers
    timer()
  File ".../eventlet/eventlet/hubs/timer.py", line 59, in __call__
    cb(*args, **kw)
  File ".../eventlet/eventlet/greenthread.py", line 221, in main
    result = function(*args, **kwargs)
  File ".../eventlet/eventlet/wsgi.py", line 789, in process_request
    proto.__init__(conn_state, self)
  File ".../eventlet/eventlet/wsgi.py", line 343, in __init__
    self.handle()
  File ".../eventlet/eventlet/wsgi.py", line 376, in handle
    self.handle_one_request()
  File ".../eventlet/eventlet/wsgi.py", line 450, in handle_one_request
    self.handle_one_response()
  File ".../eventlet/eventlet/wsgi.py", line 600, in handle_one_response
    request_input.discard()
  File ".../eventlet/eventlet/wsgi.py", line 248, in discard
    while self.read(buffer_size):
  File ".../eventlet/eventlet/wsgi.py", line 217, in read
    return self._chunked_read(self.rfile, length)
  File ".../eventlet/eventlet/wsgi.py", line 188, in _chunked_read
    raise IOError("unexpected end of file while parsing chunked data")
IOError: unexpected end of file while parsing chunked data
```

Note that we [behave more sensibly](https://github.com/eventlet/eventlet/blob/v0.24.1/eventlet/wsgi.py#L601-L607) on `ChunkReadError`s such as might happen with bad request framing. Maybe an alternative approach would be to `raise ChunkReadError("unexpected end of file while parsing chunked data")` instead? I was just nervous to

1. broaden the scope of `ChunkReadError` like that and
2. change the `Input.read()` contract in case any applications [were _expecting_](https://github.com/openstack/swift/blob/2.20.0/swift/common/utils.py#L4178-L4181) that it might raise `IOError`.

Note that we _also_ [behave more sensibly](https://github.com/eventlet/eventlet/blob/v0.24.1/eventlet/wsgi.py#L387-L399) if the client hangs up between requests despite the client indicating that the connection should persist -- after `_read_request_line()` returns an empty byte string, we just flag that the connection should be closed and bail.

> I'm also puzzled by his reference to Swift

This came up in the context of (someone building something on top of) [OpenStack Swift](https://github.com/openstack/swift/), which uses eventlet in its servers. I'm a maintainer there and was helping @charmster investigate the issue, but otherwise it's a red herring.

PR #540, opened 2019-01-21
[bug] reimport submodule as well in patcher.inject

luanjunyi:
# What

patcher.inject() has a bug that won't handle modules that introduce objects from sub modules. For example, [websocket-client](https://pypi.org/project/websocket-client/). In it's __init__.py, the objects are imported from sub modules like _socket.

In inject, sys.modules['websocket'] is reimported, however, websocket._socket won't be reimported thus all the functions still has the original non-greened version of modules in their __globals__. Making the inject not useful.

This bug will be triggered when import_patched/inject is called after the module(like websocket) is imported.

@ nat-goodspeed


eventlet/patcher.py
```diff
@@ -90,6 +90,10 @@ def inject(module_name, new_globals, *additional_modules):
     # Remove the old module from sys.modules and reimport it while
     # the specified modules are in place
     sys.modules.pop(module_name, None)
+    # Also remove sub modules and reimport
+    for imported_module_name in list(sys.modules.keys()):
```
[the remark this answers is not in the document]
jstasiak: Yeah, I believe a test case is necessary here, the monkey patching mechanism is tricky so it'd help keeping it working correctly in the future.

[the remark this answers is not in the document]
luanjunyi: Added unit test and comments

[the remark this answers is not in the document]
jstasiak: Since there's some global state involved I think making this an isolated test would be better. See `tests/patcher_test.py` for examples – in general you just add a simple test function calling `run_isolated('xxx.py')` and move your actual test into `tests/isolated/xxx.py`

tests/patcher/test_sub_module_in_import_patched/test.py
```diff
@@ -0,0 +1,11 @@
...
+from tests import LimitedTestCase
+import tests.patcher.test_sub_module_in_import_patched.sample_main_module
+
+
+class ImportPatchedHandlesSubmoduleTest(LimitedTestCase):
+    def test_sub_module_is_patched(self):
+        patched_module = eventlet.import_patched(
+            'tests.patcher.test_sub_module_in_import_patched.sample_main_module')
+        print(patched_module.function_use_socket())
+        assert patched_module.function_use_socket() is eventlet.green.socket
```
jstasiak: Can you also assert that the original `tests.patcher.test_sub_module_in_import_patched.sample_main_module.function_use_socket()` result is *not* `eventlet.green.socket` just to make sure we don't leak monkey patched objects outside?

luanjunyi: 👍

eventlet/patcher.py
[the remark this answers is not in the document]
luanjunyi: Done

tests/isolated/test_sub_module_in_import_patched/test.py
```diff
@@ -0,0 +1,10 @@
+import eventlet
+import tests.isolated.test_sub_module_in_import_patched.sample_main_module as test_module
+
+
+if __name__ == '__main__':
+    assert test_module.function_use_socket() is not eventlet.green.socket
```
jstasiak: One last nitpick – can you move this *after` the `import_patched` call? We're mostly interested if `import_patched` doesn't break things.

jstasiak: Sure:
* this assert doesn't check anything useful right now. Why would the returned object be anything other than the standard `socket` module?
* `import_patched` can potentially mutate modules that it should not mutate, therefore moving the assert *after* the call makes sure the non-monkey-patched module is ok

```python
    patched_module = eventlet.import_patched(
        'tests.isolated.test_sub_module_in_import_patched.sample_main_module')
    assert patched_module.function_use_socket() is eventlet.green.socket
    assert test_module.function_use_socket() is not eventlet.green.socket
```

luanjunyi: That makes sense. I intended to test the net effect of import_patched is changing the socket symbol in __globals__. But this is tested with the assert moving to the end as well. Added the change. PTAL.

PR #542, opened 2019-01-23
Increase Travis slop factor for ZMQ CPU usage. 

nat-goodspeed:
The comment in check_idle_cpu_usage() reads:

    # This check is reliably unreliable on Travis, presumably because of CPU
    # resources being quite restricted by the build environment. The workaround
    # is to apply an arbitrary factor that should be enough to make it work nicely.

Empirically -- it's not. Over the last few months there have been many Travis
"failures" that boil down to this one spurious error. Increase from a slop
factor of 1.2 to 5. If that's still unreliable, may bypass this test entirely
on Travis.


nat-goodspeed:
I disagree: this change does not disable the test. It just boosts the threshold from 12% to 50%. Arguably the calls should pass keyword arguments:

    tests.check_idle_cpu_usage(duration=0.2, allowed_part=0.1)

But given how consistently that test causes spurious Travis failures, I'm about ready to disable it altogether anyway.

nat-goodspeed:
@jstasiak, please review?

luanjunyi:
It used to be *= 1.2. You changed it to be *= 5. This is not from 12% to 50%. It's from 120% to 500%!

nat-goodspeed:
@luanjunyi, help me understand the math here.

`allowed_part` is passed as 0.1 (10%).

The previous code multiplies `allowed_part` by 1.2 (producing 0.12, or 12%).

I want to change that factor to 5 (which would produce 0.5, or 50%).

What part of the above am I getting wrong?

luanjunyi:
You need to understand what does the 'percentage' mean in your 'math', with the context. The indent of check_idle_cpu_usage if to make sure CPU is idle. To do that, it make sure CPU time is less than a threshold. Whatever threshold is passed in, you change it to 5 times the original value if the running envrionment is TRAVIS. Do you call this 500% increase or 50% increase?

It's up to you what you call. But the reason it works is to make sure there's little CPU usage while sleeping. That's why `tests.check_idle_cpu_usage(0.2, 0.1)` makes sense. Now you simply say let's do tests.check_idle_cpu_usage(0.2, 0.5). Meaning while we sleep 0.2 seconds, feel free to have the CPU run 0.5 seconds and we say the CPU is idle.

Does this make sense?

luanjunyi:
I see what you mean. I thought the compare predicate is cpu < allowed_part instead of cpu < duration * allowed_part

nat-goodspeed:
@temoto @jstasiak I would love an approving review -- this issue continues to block otherwise-valid pull requests.

nat-goodspeed:
Thank you!!

nat-goodspeed:
I never again want to have a pull request blocked due to spurious Travis failures of these tests. A cautious increase in the slop factor would probably reduce the incidence of spurious failures, while still tossing us the occasional "gotcha."

The way I see it, bumping the limit is actually more conservative than disabling those tests altogether. This way if ZMQ actually does go nuts and spin the CPU, we still have some hope that Travis could catch *that.*

temoto:
You are very correct. I'm sorry for not leaving any better option.

PR #544, opened 2019-01-30
Proper way to handle websocket WSGI apps

### LINKED ISSUE #543: Websocket connections are not closed properly when we use them with WSGI middleware.
I'm getting following error whenever websocket connections are closed.

``` python
  File "/usr/local/lib/python3.6/site-packages/eventlet/wsgi.py", line 580, in handle_one_response
    write(b'')
  File "/usr/local/lib/python3.6/site-packages/eventlet/wsgi.py", line 472, in write
    raise AssertionError("write() before start_response()")
AssertionError: write() before start_response()
```

More precisely, I'm using [Flask-SocketIO](http://flask-socketio.readthedocs.io) with eventlet, with [Flask integration of sentry-sdk](https://docs.sentry.io/platforms/python/flask/).

As I see it, the problem occurred by following reason.

In `eventlet.websocket`, you made a special object called `ALREADY_HANDLED` which makes special functionalities for websocket handlers
https://github.com/eventlet/eventlet/blob/05d613d8ed123b87f64f8d1987968c5bf825b07a/eventlet/websocket.py#L136-L138

And, in `eventlet.wsgi`

The websocket-specific functionalities are handled by type of the return value of WSGI app. 

https://github.com/eventlet/eventlet/blob/05d613d8ed123b87f64f8d1987968c5bf825b07a/eventlet/wsgi.py#L547-L550

However, In sentry-sdk, the plugin wraps the WSGI responses (which are iterable objects) with there wrapper called [`_ScopedResponse`](https://github.com/getsentry/sentry-python/blob/3ac9097165180b4a53e5d483f73a3abfc2d239e6/sentry_sdk/integrations/wsgi.py#L134).

This is fine, since [PEP 3333](https://www.python.org/dev/peps/pep-3333/) specifies the return value of WSGI app just have to be an iterable object.

But it breaks the type check on `eventlet.wsgi`. So eventlet doesn't finish at the if statement, and emits the above error message.

I think, you may have to use another way to handle websocekt specific features.

Hardtack:
As I reported at #543, using return value to handle websocket-specific functionalities does not work well.
In this PR, it uses a coroutine-local flag called `wsgi.WSGI_LOCAL.already_handled`.
Just set flag on in websocket app, and then check that in `wsgi.py`

It fixes #543

### DIFF
--- a/tests/websocket_test.py
+++ b/tests/websocket_test.py
@@ -552,7 +552,8 @@ class TestWebSocket(tests.wsgi_test._TestBase):
         headers, result = resp.split(b"\r\n\r\n")
         msgs = [result.strip(b"\x00\xff")]
         msgs.extend(sock.recv(20).strip(b"\x00\xff") for _ in range(10))
-        assert msgs == [six.b("msg {}".format(i)) for i in range(10)]
+        expect = [six.b("msg {}".format(i)) for i in range(10)] + [b""]
+        assert msgs == expect
         # In case of server error, server will write HTTP 500 response to the socket
         msg = sock.recv(20)
         assert not msg
--- a/tests/websocket_test.py
+++ b/tests/websocket_test.py
@@ -552,7 +552,7 @@ class TestWebSocket(tests.wsgi_test._TestBase):
         headers, result = resp.split(b"\r\n\r\n")
         msgs = [result.strip(b"\x00\xff")]
         msgs.extend(sock.recv(20).strip(b"\x00\xff") for _ in range(10))
-        assert msgs == [six.b("msg %d" % i) for i in range(10)]
+        assert msgs == [six.b("msg {}".format(i)) for i in range(10)]
         # In case of server error, server will write HTTP 500 response to the socket
         msg = sock.recv(20)
         assert not msg
--- a/tests/websocket_test.py
+++ b/tests/websocket_test.py
@@ -528,6 +528,37 @@ class TestWebSocket(tests.wsgi_test._TestBase):
         with eventlet.Timeout(1):
             pool.waitall()
 
+    def test_wrapped_wsgi(self):
+        site = self.site
+
+        def wrapper(environ, start_response):
+            for chunk in site(environ, start_response):
+                yield chunk
+
+        self.site = wrapper
+        self.spawn_server()
+        connect = [
+            "GET /range HTTP/1.1",
+            "Upgrade: WebSocket",
+            "Connection: Upgrade",
+            "Host: {}:{}".format(*self.server_addr),
+            "Origin: http://{}:{}".format(*self.server_addr),
+            "WebSocket-Protocol: ws",
+        ]
+        sock = eventlet.connect(self.server_addr)
+
+        sock.sendall(six.b("\r\n".join(connect) + "\r\n\r\n"))
+        resp = sock.recv(1024)
+        headers, result = resp.split(b"\r\n\r\n")
+        msgs = [result.strip(b"\x00\xff")]
+        msgs.extend(sock.recv(20).strip(b"\x00\xff") for _ in range(10))
+        assert msgs == [six.b("msg %d" % i) for i in range(10)]
+        # In case of server error, server will write HTTP 500 response to the socket
+        msg = sock.recv(20)
+        assert not msg
+        sock.close()
+        eventlet.sleep(0.01)
+
 
 class TestWebSocketSSL(tests.wsgi_test._TestBase):
     def set_site(self):
--- a/tests/wsgi_test.py
+++ b/tests/wsgi_test.py
@@ -103,7 +103,8 @@ def chunked_post(env, start_response):
 
 def already_handled(env, start_response):
     start_response('200 OK', [('Content-type', 'text/plain')])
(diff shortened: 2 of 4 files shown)


Hardtack:
Anyone watching this?

tests/wsgi_test.py
```diff
@@ -114,8 +115,7 @@ def __call__(self, env, start_response):
 
 
 class IterableApp(object):
-
-    def __init__(self, send_start_response=False, return_val=wsgi.ALREADY_HANDLED):
+    def __init__(self, send_start_response=False, return_val=[]):
```
jstasiak: Let's not use a mutable default (just in case `return_val` is modified by something further down the line in the future)

Hardtack: Thanks!

Hardtack:
Hi, Do you have any plan to merge this? or should I have to do something more for this PR?

DenisOH:
@jstasiak 
Any plans on merging this one?

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+            for chunk in site(environ, start_response):
+                yield chunk
+
+        self.site = wrapper
+        self.spawn_server()
+        connect = [
+            "GET /range HTTP/1.1",
+            "Upgrade: WebSocket",
+            "Connection: Upgrade",
+            "Host: %s:%s" % self.server_addr,
```
temoto, line 544:
```suggestion
            "Host: {}:{}".format(*self.server_addr),
```

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+            "Origin: http://%s:%s" % self.server_addr,
```
temoto, line 545:
```suggestion
            "Origin: http://{}:{}".format(*self.server_addr),
```

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+            "WebSocket-Protocol: ws",
+        ]
+        sock = eventlet.connect(self.server_addr)
+
+        sock.sendall(six.b('\r\n'.join(connect) + '\r\n\r\n'))
```
temoto, line 550:
```suggestion
        sock.sendall(six.b("\r\n".join(connect) + "\r\n\r\n"))
```

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+        resp = sock.recv(1024)
+        headers, result = resp.split(b'\r\n\r\n')
```
temoto, line 552:
```suggestion
        headers, result = resp.split(b"\r\n\r\n")
```

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+        msgs = [result.strip(b'\x00\xff')]
```
temoto, line 553:
```suggestion
        msgs = [result.strip(b"\x00\xff")]
```

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+        cnt = 9
+        while cnt:
+            msgs.append(sock.recv(20).strip(b'\x00\xff'))
+            cnt -= 1
```
temoto, line 557:
```suggestion
        msgs.extend(sock.recv(20).strip(b"\x00\xff") for _ in range(10))
```

tests/websocket_test.py
```diff
@@ -528,6 +528,40 @@ def test_close_idle(self):
...
+        self.assertEqual(msgs, [six.b('msg %d' % i) for i in range(10)])
```
temoto, line 558:
```suggestion
        assert msgs == [six.b("msg {}".format(i)) for i in range(10)]
```

temoto (changes requested):
Downside of this change is that WSGI implementation becomes strictly bound to eventlet greenthreads. But it's unlikely someone uses eventlet.wsgi without the rest of library.

Please accept style fixes to make future auto-format touch less code.

Hardtack:
@temoto I closed this PR by mistake :(

I accepted your suggestion and applied recent commits by rebasing origin's master branch. Thank you for reviewing!

tests/websocket_test.py
```diff
@@ -528,6 +528,37 @@ def test_close_idle(self):
...
+            "WebSocket-Protocol: ws",
+        ]
+        sock = eventlet.connect(self.server_addr)
+
+        sock.sendall(six.b("\r\n".join(connect) + "\r\n\r\n"))
+        resp = sock.recv(1024)
+        headers, result = resp.split(b"\r\n\r\n")
+        msgs = [result.strip(b"\x00\xff")]
+        msgs.extend(sock.recv(20).strip(b"\x00\xff") for _ in range(10))
+        assert msgs == [six.b("msg %d" % i) for i in range(10)]
```
temoto, line 555:
```suggestion
        assert msgs == [six.b("msg {}".format(i)) for i in range(10)]
```

temoto (approved):

PR #548, opened 2019-02-20
wsgi: fix Input.readline on Python 3

tipabu:
Previously, we would compare the last item of a byte string with a newline in a native string. On Python 3, getting a single item from a byte string give you an integer (which will not be equal to any string), so `readline()` would return the entire request body.

While we're at it, fix the return type when the caller requests that zero bytes be read.


tipabu:
Bonus fix for `readlines()` -- though I'd be happy to split them up if you'd prefer.

Issue #551, opened 2019-02-25
Spurious wake up after select calls


gogglesguy:
Please see the code below. 

We expect the program to sleep for at least 4 seconds, however, if you run the program as is, the assert will trigger and the program will exit immediately. If you change the second select to either

> select([sd],[],[sd],0)    # check read, except

or

> select([], [sd], [sd],0)  # check write, except

the program will sleep as expected and no assert is triggered. 
```
from __future__ import print_function, division, unicode_literals
import eventlet
import socket
import select
import time

eventlet.monkey_patch(socket=True, select=True)


def main():
    sd = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
    sd.bind(('127.0.0.1', 0))
    addr = sd.getsockname()
    sd.sendto(b'test', addr)

    select.select([sd], [], [])
    select.select([sd], [sd], [sd], 0)

    now = time.time()
    eventlet.sleep(4)  # This sleep returns immediately for some reason
    elapsed = time.time() - now

    assert elapsed > 1.0


if __name__ == '__main__':
    main()
```

dralley:
I'd like to see this get more attention, it causes problems for redis-py

kashirin-alex:
Interesting, It work correct on my branch, https://github.com/kashirin-alex/eventlet/blob/master/eventlet/green/select.py

just I'm unable to find the difference for the fix required to be applied,
pip install https://github.com/kashirin-alex/eventlet/archive/master.zip

EVENTLET_HUB=v1_  followed epolls_linuxfd /  epolls / kqueue / poll / selects 
eg:
```bash 
EVENTLET_HUB=v1_epolls  python test.py 
```
add a
```python 
print (eventlet.hubs.get_hub())
```
to confirm the hub that is been used, if it is "epolls_linuxfd " it is a completely different type of timer process

gogglesguy:
Still broken after 3 years in eventlet 0.33. We moved on to gevent.

PR #552, opened 2019-02-26
Fix for select read+write and spurious wakeup

erickj00001:
Addresses #551 

The green select function relies on removing the listeners in the "finally" block to ensure that no other callbacks will be issued.  But in some hubs (poll and epolls, but not selects), callbacks are queued up in a local variable (called "callbacks"), so will not be prevented even if we remove the corresponding listeners.  So in the case where both read and write are applicable, one of those callbacks gets issued after the select call returns, causing a spurious wakeup.

This addresses the issue by scheduling another callback to call current.switch, so that Hub.wait has a chance to finish issuing all callbacks.

This also has the benefit that select will now return both the read and write flags if they both apply, which is more consistent with the behavior of the original select function.


dralley:
RQ users also, via redis-py

Issue #53, opened 2013-09-30
Empty GreenPile iterator hangs


dmcnelis:
Here is a simplified recreation of the problem:

```
from eventlet import GreenPile, GreenPool
pool = GreenPool()
pile = GreenPile(pool)
for x in pile:
    print x
```

This will just sit and spin, without throwing any sort of exception / error.

nat-goodspeed:
I hit this one in my own code a couple days ago. It was a nasty surprise.

I reviewed all references to `GreenPile` on the master branch. Most of them are tests, and those tests unconditionally spawn at least one greenthread on the `GreenPile`.

It seems to me that the `feedscraper.py` example is susceptible to this bug if an empty (or all whitespace) body is posted. To me, that example clearly implies that iterating over a `GreenPile` with no `spawn()` calls should produce an empty sequence. Otherwise, the web application hangs.

`greenpool.GreenMap`, used by `GreenPool.starmap()` and `imap()`, is a subclass of `GreenPile`. But `starmap()` should (!) never return an "empty" `GreenMap`: it spawns `GreenPool._do_map()`, which spawns a greenthread per iterable item and then unconditionally spawns `return_stop_iteration()`. The same is true of `imap()`, which calls `starmap()`.

I think I know where the undesirable empty-`GreenPile` behavior comes from. Please sanity-check my reasoning.

[This commit from 2009](https://github.com/eventlet/eventlet/commit/3ddbba23de2ad1485d968bddbbb3e2866b4429c2), by one of the original eventlet authors, explicitly adds `GreenPile.used`. The only effect of the `used` flag is to force the bad behavior of which this bug (correctly) complains. The same commit introduces the `GreenPile` subclass (then called) `GreenImap`, reimplementing the `GreenPool` helper method (then called) `_do_imap()` to use it.

[...]

PR #555, opened 2019-03-03
 #53: Make a GreenPile with no spawn()s an empty sequence.

nat-goodspeed:
Remove the GreenPile.used flag. This was probably originally intended to
address the potential problem of GreenPool.starmap() returning a (still empty)
GreenPile, a problem that has since been addressed by using GreenMap which
requires an explicit marker from the producer. Currently the only effect of
GreenPile.used is to make GreenPile hang when used with an empty sequence.

Test the empty GreenPile case.

Even though GreenMap is probably not intended for general use, make it
slightly more consumer-friendly by adding a done_spawning() method instead of
requiring the consumer to spawn(return_stop_iteration). GreenPool._do_map()
now calls done_spawning(). Remove return_stop_iteration().

Since done_spawning() merely spawns a function that returns StopIteration, any
existing consumer that explicitly does the same will still work. However, this
is potentially a breaking change if any consumers specifically reference
eventlet.greenpool.return_stop_iteration() for that or any other purpose.

Refactor GreenPile.next(), breaking out the bookkeeping detail to new _next()
method. Make subclass GreenMap call base-class _next(), eliminating the need
to replicate that bookkeeping detail.


nat-goodspeed:
The failing check is, of course, the spurious Travis ZMQ CPU fail, which will be addressed by #542.

PR #557, opened 2019-03-18
wsgi: Only send 100 Continue response if no response has been sent yet

tipabu:
Some applications may need to perform some long-running operation during a client-request cycle. To keep the client from timing out while waiting for the response, the application issues a status _pro tempore_, dribbles out whitespace (or some other filler) periodically, and expects the client to parse the final response to confirm success or failure.

Previously, if the application was *too* eager and sent data before ever reading from the request body, we would write headers to the client, send that initial data, but then **still send the `100 Continue`** when the application finally read the request. Since this would occur on a chunk boundary, the client cannot parse the size of the next chunk, and everything goes off the rails.

Now, only be willing to send the 100 Continue response if we have not sent headers to the client.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -92,7 +92,8 @@ class Input(object):
                  sock,
                  wfile=None,
                  wfile_line=None,
-                 chunked_input=False):
+                 chunked_input=False,
+                 headers_sent=None):
 
         self.rfile = rfile
         self._sock = sock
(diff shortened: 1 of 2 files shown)


tests/wsgi_test.py
```diff
@@ -988,6 +988,67 @@ def wsgi_app(environ, start_response):
...
+
+        def expect_chunk(data):
+            expected = chunkify(data)
+            self.assertEqual(expected, fd.read(len(expected)))
+
+        # Can even see that initial whitespace
+        expect_chunk(b' ')
+
+        # Send message
+        body = b'some data'
```
tipabu: Bah; left this in accidentally.

eventlet/wsgi.py
```diff
@@ -446,23 +458,23 @@ def handle_one_request(self):
                 self.close_connection = 1
                 return
 
-        self.environ = self.get_environ()
+        headers_sent = []
```
tipabu: I'm not entirely sure whether I like moving this up and out to `handle_one_request`, but I didn't see a good way to plumb the flag down into `Input` otherwise. If anyone else has a better idea, I'm all ears...

PR #558, opened 2019-03-19
Clean up some warnings in wsgi_test

tipabu:
Previously, running tests with pytest would catch

    DeprecationWarning: cgi.parse_qs is deprecated, use urllib.parse.parse_qs instead

and

    DeprecationWarning: capitalize_response_headers is disabled...

The latter one was actually part of what was being tested, so make assertions about the warning being properly raised.


temoto (approved):

jstasiak (approved):

temoto:
@tipabu it doesn't pass `len(catched)==1` check. I've merged master, but it also didn't pass before that.

temoto (changes requested):
LGTM but tests need to be fixed

PR #563, opened 2019-04-26
Don’t import deprecated dns.hash

flying-sheep:


flying-sheep:
No idea why your checks fail. From the messages I assume my code is not the problem.

temoto:
Thank you, but it's not clear why change is needed. Please run a test that fails on master first.

Feel free to reopen if anyone wants to continue work here.

PR #564, opened 2019-05-03
tests: Improve ECONNREFUSED checks

iv-m:
Only the value for the current platform should be considered
valid here, so this check uses the constant from `errno`
module as expected output, instead of hardcoded ints.

Also, this fixes build on MIPS, where ECONNREFUSED is defined
as 146.

Signed-off-by: Ivan A. Melnikov <iv@altlinux.org>

### DIFF
--- a/tests/test__socket_errors.py
+++ b/tests/test__socket_errors.py
@@ -1,3 +1,4 @@
+import errno
 import unittest
 import socket as _original_sock
 from eventlet.green import socket
@@ -18,7 +19,7 @@ class TestSocketErrors(unittest.TestCase):
             self.fail("Shouldn't have connected")
         except socket.error as ex:
             code, text = ex.args
-            assert code in [111, 61, 10061], (code, text)
+            assert code == errno.ECONNREFUSED, 'Expected ECONNREFUSED, got {0} ({1})'.format(code, text)
             assert 'refused' in text.lower(), (code, text)
 
     def test_timeout_real_socket(self):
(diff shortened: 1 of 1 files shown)


tests/test__socket_errors.py
```diff
@@ -54,9 +55,9 @@ def test_timeout(self, socket=socket):
...
 
 def test_create_connection_refused():
-    errno = None
     try:
         socket.create_connection(('127.0.0.1', 1))
     except socket.error as ex:
-        errno = ex.errno
-    assert errno in [111, 61, 10061], 'Expected socket.error ECONNREFUSED, got {0}'.format(errno)
+        code, text = ex.args
+        self.assertEqual(code, errno.ECONNREFUSED,
```
jstasiak: Let's keep the previous indentation level here I think, instead this test can silently fail if `socket.create_connection()` succeeds for some reason

iv-m: Yes, I missed that case, sorry.

I think it's better to be consistent with the above code in handling it, though, and add one more assert before `except`. I'll update the pull request.

tests/test__socket_errors.py
```diff
@@ -18,7 +19,7 @@ def test_connection_refused(self):
             self.fail("Shouldn't have connected")
         except socket.error as ex:
             code, text = ex.args
-            assert code in [111, 61, 10061], (code, text)
+            assert code == errno.ECONNREFUSED, 'Expected ECONNREFUSED, got {0} ({1})'.format(code, text))
```
temoto, line 22: typo
```suggestion
            assert code == errno.ECONNREFUSED, 'Expected ECONNREFUSED, got {0} ({1})'.format(code, text)
```

tests/test__socket_errors.py
```diff
@@ -54,9 +55,9 @@ def test_timeout(self, socket=socket):
...
 def test_create_connection_refused():
-    errno = None
     try:
         socket.create_connection(('127.0.0.1', 1))
+        assert False, "Shouldn't have connected"
     except socket.error as ex:
-        errno = ex.errno
-    assert errno in [111, 61, 10061], 'Expected socket.error ECONNREFUSED, got {0}'.format(errno)
+        code, text = ex.args
+        assert code == errno.ECONNREFUSED, 'Expected ECONNREFUSED, got {0} ({1})'.format(code, text))
```
temoto, line 63:
```suggestion
        assert code == errno.ECONNREFUSED, 'Expected ECONNREFUSED, got {0} ({1})'.format(code, text)
```

temoto:
CI test blocked by #636

temoto:
rebased on master

PR #569, opened 2019-05-16
wsgi: minimize API changes for 100-continue fix

tipabu:
In commit b9bf369778ec9798b3b6cffe59b7fd15f6159013, we stopped sending `100 Continue` responses in the middle of a response when the application was over-eager to start sending back bytes, but I did that by pulling the `headers_sent` state out of `handle_one_response()`, up into `handle_one_request()`, and plumbing it through to

  * `get_environ()`,
  * `Input.__init__()`, and finally
  * `handle_one_response()`.

This works, but updates a whole bunch of `HttpProtocol` APIs in ways that consumers may not have been expecting. For example, if someone wanted to subclass `HttpProtocol` and override `get_environ()`, they [may not have bothered to include `*args` and `**kwargs`](https://github.com/openstack/swift/blob/2.21.0/swift/common/wsgi.py#L509-L517) to accommodate future API changes. That code should certainly be fixed, but we shouldn't break them gratuitously.

Now, wait to introduce the `headers_sent` state until `handle_one_response()` once more, and push it directly into the request's `Input`. All the same protections with minimal API disruption.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -92,8 +92,7 @@ class Input(object):
                  sock,
                  wfile=None,
                  wfile_line=None,
-                 chunked_input=False,
-                 headers_sent=None):
+                 chunked_input=False):
 
         self.rfile = rfile
         self._sock = sock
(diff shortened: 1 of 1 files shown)


eventlet/wsgi.py
```diff
@@ -458,23 +458,26 @@ def handle_one_request(self):
...
             self.server.outstanding_requests -= 1
 
-    def handle_one_response(self, headers_sent):
+    def handle_one_response(self):
         start = time.time()
         headers_set = []
+        headers_sent = []
+        # Push the headers-sent state into the Input so it won't send a
+        # 100 Continue response if we've already started a response.
+        self.environ['wsgi.input'].headers_sent = headers_sent
```
clayg: it's a little tricky how this mutable/non-local becomes truthy below when we `headers_sent.append(1)` - I can see why you were trying to do the plumbing but agree the API breaking signature change was probably a mis-step.

PR #572, opened 2019-05-24
Prepping for v0.25.0 release

tipabu:
It's been some nine months or so since 0.24.1 -- I figured it might be nice to have another release. Hopefully I did OK in curating the changelog?


jstasiak:
Yup!

PR #578, opened 2019-06-26
wsgi: Don't break HTTP framing during 100-continue handling

tipabu:
`Expect: 100-continue` is a funny beast -- the client sends it to indicate that it's willing to wait for an early error, but

- the client has no guarantee that the server supports `100 Continue`,
- the server gets no indication of how long the client's willing to wait for the go/no-go response, and
- even if it did, the server has no way of knowing that the response it *emitted* within that time was actually *received* within that time
- so the client may have started sending the body regardless of what the server's done.

As a result, the server only has two options when it *does not* send the `100 Continue` response:

- close the socket
- read and discard the request body

Previously, we did neither of these things; as a result, a request body could be interpreted as a new request. Now, close the connection. <s>read and discard the body (in part because we already have tests that exercise pipelined requests after an early response).

Note that this is some pretty *old* bad behavior; however, prior to 5b5afd1859dd181c14d5dc1be6da1bc1039e52bd we actually did the right thing!</s>

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -134,8 +134,12 @@ class Input(object):
         # Reinitialize chunk_length (expect more data)
         self.chunk_length = -1
 
+    @property
+    def should_send_hundred_continue(self):
+        return self.wfile is not None and not self.is_hundred_continue_response_sent
+
     def _do_read(self, reader, length=None):
-        if self.wfile is not None and not self.is_hundred_continue_response_sent:
+        if self.should_send_hundred_continue:
             # 100 Continue response
             self.send_hundred_continue_response()
             self.is_hundred_continue_response_sent = True
@@ -152,7 +156,7 @@ class Input(object):
         return read
 
     def _chunked_read(self, rfile, length=None, use_readline=False):
-        if self.wfile is not None and not self.is_hundred_continue_response_sent:
+        if self.should_send_hundred_continue:
             # 100 Continue response
             self.send_hundred_continue_response()
             self.is_hundred_continue_response_sent = True
@@ -464,9 +468,11 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
         start = time.time()
         headers_set = []
         headers_sent = []
+        # Grab the request input now; app may try to replace it in the environ
+        request_input = self.environ['eventlet.input']
         # Push the headers-sent state into the Input so it won't send a
         # 100 Continue response if we've already started a response.
-        self.environ['wsgi.input'].headers_sent = headers_sent
+        request_input.headers_sent = headers_sent
 
         wfile = self.wfile
         result = None
@@ -563,10 +569,15 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
                 result = self.application(self.environ, start_response)
 
                 # Set content-length if possible
-                if headers_set and \
-                        not headers_sent and hasattr(result, '__len__') and \
-                        'Content-Length' not in [h for h, _v in headers_set[1]]:
-                    headers_set[1].append(('Content-Length', str(sum(map(len, result)))))
+                if headers_set and not headers_sent and hasattr(result, '__len__'):
+                    # We've got a complete final response
+                    if 'Content-Length' not in [h for h, _v in headers_set[1]]:
+                        headers_set[1].append(('Content-Length', str(sum(map(len, result)))))
+                    if request_input.should_send_hundred_continue:
+                        # We've got a complete final response, and never sent a 100 Continue.
+                        # There's no chance we'll need to read the body as we stream out the
+                        # response, so we can be nice and send a Connection: close header.
+                        self.close_connection = 1
 
                 towrite = []
                 towrite_size = 0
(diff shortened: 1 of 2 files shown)


tipabu:
> It would be nice to add some protection to how much we'll discard when the client and wsgi.app are having a hard time working out what they want to do.

That's a good idea! I wrote up https://github.com/tipabu/eventlet/commit/21a61d9160581968deb890f8c11a373c4fa98a8c on top of this to implement it, but I'm still trying to decide whether I like my approach. Like, it feels weird that it would have to be set on every (rejected) request, but I'm not quite sure where to stuff it _otherwise_. It also feels awkward that the app would have to do the book-keeping if/when it does some reads before rejecting -- at that point, maybe it should just be a binary flag to say "trash this connection **now**" and _skip_ the `read()`s...

And that's even _before_ I try to figure out where I ought to _document_ such a feature :stuck_out_tongue:

clayg:
> clients that try to incorrectly skip sending a body after they get a early error on 100-Continue

It may not have been clear - I don't care about such clients; that was mostly a missive about how this corner of HTTP is wonky

> It may be safer in all cases to just close the socket after sending the 100-continue error response.

This would probably best for everyone, does tipabu@21a61d9 do that or something else?

We can merge this in the meantime.

clayg:
I've cooled on this patch.

Despite lots of resources making it clear that discarding the body is the right thing to do, the way that master works (guarding the discard so that it only happens when client did NOT send 100-Continue) actually seems to have the broader compatibility with how clients that send 100-Continue actually *behave*

https://curl.se/mail/lib-2004-08/0002.html

That is to say "clients that try to incorrectly skip sending a body after they get a early error on 100-Continue" are seemingly incredibly common (e.g. python boto, aws-sdk-go) and this change makes things worse.

We should just close the connection in all cases; it's the only way to make sure the client and server agree on what to do next.  Even if the body might be small, the cost of getting HTTP framing wrong is too high.  There's no need to optimize; just close the connection.

temoto:
Agree with "just disconnect" solution.

tipabu:
Updated to close the connection.

eventlet/wsgi.py
```diff
@@ -607,11 +618,22 @@ def cap(x):
...
+                # We just sent the final response, no 100 Continue. Client may or
+                # may not have started to send a body, and if we keep the connection
+                # open we've seen clients either
+                #   * send a body, then start a new request
+                #   * skip the body and go straight to a new request
+                # Looks like the most broadly compatible option is to close the
+                # connection and let the client retry.
+                # https://curl.se/mail/lib-2004-08/0002.html
+                # Note that we likely *won't* send a Connection: close header at this point
+                self.close_connection = 1
```
clayg, line 631: this is really the crux of it; we've handled this response - if we didn't send a 100-continue and the client was expecting one that's it for this connection

does it matter if it was a zero byte POST/PUT? can you even set expect-100 in that case?

so up above where we're being cute with hasattr `__len__` and connection close headers; do we really want to close the connection, or do we want to set "should_send_hundred_continue" to False and let it discard?

eventlet/wsgi.py
```diff
@@ -607,11 +618,22 @@ def cap(x):
...
+
             if (request_input.chunked_input or
                     request_input.position < (request_input.content_length or 0)):
-                # Read and discard body if there was no pending 100-continue
-                if not request_input.wfile and self.close_connection == 0:
+                # Read and discard body if connection is going to be reused
+                if self.close_connection == 0:
```
clayg, line 636: this could probably be collapsed into a single expression:

there's body on the wire and we're still using this connection

and it's intirely independent from should_send_hundred_continue, which is nice

clayg (commented):
I think the tests are great, the finally block handling looks pretty good

The work to get the connection: close header on the wire raises the question for me how many apps were responding without reading the body and expect the server to discard it - outside of the corner case with 100-continue (and wonky clients that maybe won't send the body) it seems straight forward and might have been working well

temoto (approved):
LGTM

eventlet/wsgi.py
temoto: > does it matter if it was a zero byte POST/PUT? can you even set expect-100 in that case?

Expect-100 with zero body is valid, with especially clear semantic for PUT (rewrite with empty). It's not economic to close connection in zero body request case, but probably it's also very rare.

> so up above where we're being cute with hasattr __len__ and connection close headers; do we really want to close the connection, or do we want to set "should_send_hundred_continue" to False and let it discard?

My vote for close as more robust option.

temoto:
Unsure about lot of comments. Waiting for hard do or don't merge for few days.

clayg:
We shipped it exactly like this and it rolled out yesterday - no fires yet :rofl: 

I like everything except for the tricky bit where we get fancy right after calling the applicaiton trying to see if we can send headers - it's such a ninche case [1] - marginal value at the expense of duplicated `if should send 100: close = 1` logic

If we ever did want to improve our expect-100-continue connection handling, i.e.:

1) let 0-byte bodies stay open (impossible to get framing wrong)
2) let chunked-transfer maybe send `0\r\n\r\n` (impossible to confuse with next HTTP request)
3) interpolate next bytes and attempt to catagorize as body or request, allowing small bodies to be discarded (kinda sketch)

... we'd have to make sure we remember to do that in two places.  I don't see an obvious way to DRY it out - it's necessary complexity if we want to handle the speical case.

1. maybe not so niche?  When we return a 429 to an expect-100 PUT our resp probably passes in the "list-of-strings" test.

temoto:
@clayg right, the two places. `eventlet.wsgi` code could use refactoring for easier understanding, including breaking the big request processing function into few parts. It could be a single `close=1` then. Thanks for review and testing.

Issue #580, opened 2019-07-10
cqi.parse_qs was removed in Python 3.8


hroncok:
https://github.com/eventlet/eventlet/blob/b9bf369778ec9798b3b6cffe59b7fd15f6159013/tests/wsgi_test.py#L317-L321

This fails on Python 3.8.0b2 with:

```
======================================================================
FAIL: test_007_get_arg (tests.wsgi_test.TestHttpd)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/builddir/build/BUILD/eventlet-0.25.0/tests/wsgi_test.py", line 338, in test_007_get_arg
    self.assertEqual(result.body, b'a is a, body is a=a')
AssertionError: b'Traceback (most recent call last):\n  Fil[489 chars]\'\n' != b'a is a, body is a=a'

----------------------------------------------------------------------
Ran 694 tests in 56.076s

FAILED (SKIP=95, failures=1)
```

And the traceback is:

```
Traceback (most recent call last):
  File "/builddir/build/BUILD/eventlet-0.25.0/eventlet/wsgi.py", line 566, in handle_one_response
    result = self.application(self.environ, start_response)
  File "/builddir/build/BUILD/eventlet-0.25.0/tests/wsgi_test.py", line 113, in __call__
    return self.application(env, start_response)
  File "/builddir/build/BUILD/eventlet-0.25.0/tests/wsgi_test.py", line 321, in new_app
    a = cgi.parse_qs(body).get('a', [1])[0]
AttributeError: module 'cgi' has no attribute 'parse_qs'
```

See:

https://docs.python.org/3.7/library/cgi.html#cgi.parse_qs
This function is deprecated in this module. Use urllib.parse.parse_qs() instead. It is maintained here only for backward compatibility.

https://docs.python.org/3.8/whatsnew/3.8.html#api-and-feature-removals
parse_qs, parse_qsl, and escape are removed from cgi module. They are deprecated from Python 3.2 or older.

hroncok:
https://github.com/eventlet/eventlet/pull/581

PR #581, opened 2019-07-10
Stop using deprecated cgi.parse_qs() to support Python 3.8

hroncok:
Fixes https://github.com/eventlet/eventlet/issues/580


jstasiak:
Cheers! For what it's worth it's not necessary to create an issue if you have a pull request ready.

hroncok:
> For what it's worth it's not necessary to create an issue if you have a pull request ready.

Ack. I filed the issue and only after that I started to work on a fix.

PR #582, opened 2019-07-10
Use _imp instead of deprecated imp

onovy:
Module imp is deprecated in favour of importlib. But importlib doesn't
support acquire_lock/release_lock/lock_held. Use internal _imp module
instead.

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -1,4 +1,7 @@
-import imp
+try:
+    import _imp as imp
+except ImportError:
+    import imp
 import sys
 
 import eventlet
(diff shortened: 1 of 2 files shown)


tito:
+1 for this PR.

This is happening from a long time, i just put it here as a reference:
```
tests/test_analyzer.py::test_analyzer_get_list
  /home/xxx/.conda/envs/xxx/lib/python3.7/site-packages/eventlet/patcher.py:1: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp
```

temoto:
Obviously not relying on internal module is a better option. I wonder if _imp is even present on PyPy or other interpreters.

This seems like a harmless change for now though. Waiting on input from other people for now.

PR #584, opened 2019-08-19
workaround for pathlib on py 3.7

### LINKED ISSUE #534: Eventlet broke pathlib on Python 3.7 when patches `os`
Script to reproduce:

```python
import eventlet
eventlet.monkey_patch(os=True)
from pathlib import Path
Path('.').open()

# Python 3.x: IsADirectoryError: [Errno 21] Is a directory: '.'
# ^ it's ok
# Python 3.7.1: TypeError: open: path should be string, bytes or os.PathLike, not _NormalAccessor
# ^ it's not ok
```

As you can see, `Path.open` doesn't work on Python 3.7. It's because `pathlib` uses `_NormalAccessor` instance, that stores functions from `os` as attributes:

```python
class _NormalAccessor(_Accessor):
    open = os.open
```

Python <3.7 wraps this functions in `staticmethod` like this:

```python
class _NormalAccessor(_Accessor):
    open = staticmethod(os.open)
```

However Python 3.7 doesn't. Python doesn't bound `os.open` to `_NormalAccessor` because `os.open` isn't a function. But when eventlet patches `os.open` by the custom python function then interpreter bound this function (`eventlet.green.os.open`) to `_NormalAccessor`.

I can't report this bug to Python because I can't register on bugs.python.org >.<

Possible solution: patch `_NormalAccessor` too. But don't forget: in Python 3.6 passed argument has to converted to string:

```python
class _NormalAccessor(_Accessor):

    def _wrap_strfunc(strfunc):
        @functools.wraps(strfunc)
        def wrapped(pathobj, *args):
            return strfunc(str(pathobj), *args)
        return staticmethod(wrapped)

    def _wrap_binary_strfunc(strfunc):
        @functools.wraps(strfunc)
        def wrapped(pathobjA, pathobjB, *args):
            return strfunc(str(pathobjA), str(pathobjB), *args)
        return staticmethod(wrapped)

    open = _wrap_strfunc(os.open)
```

References:
* [Python 3.6 pathlib source](https://github.com/python/cpython/blob/3.6/Lib/pathlib.py#L382)
* [Python 3.7 pathlib source](https://github.com/python/cpython/blob/3.7/Lib/pathlib.py#L381)

davidszotten:
pathlib._NormalAccessor wraps `open` in `staticmethod` for py < 3.7 but
not 3.7. That means we `Path.open` calls `green.os.open` with `file`
being a pathlib._NormalAccessor object, and the other arguments shifted.
Fortunately pathlib doesn't use the `dir_fd` argument, so we have space
in the parameter list. We use some heuristics to detect this and adjust
the parameters (without importing pathlib)

fixes #534

### DIFF
--- a/eventlet/green/os.py
+++ b/eventlet/green/os.py
@@ -103,6 +103,15 @@ def open(file, flags, mode=0o777, dir_fd=None):
         This behaves identically, but collaborates with
         the hub's notify_opened protocol.
     """
+    # pathlib workaround #534 pathlib._NormalAccessor wraps `open` in
+    # `staticmethod` for py < 3.7 but not 3.7. That means we get here with
+    # `file` being a pathlib._NormalAccessor object, and the other arguments
+    # shifted.  Fortunately pathlib doesn't use the `dir_fd` argument, so we
+    # have space in the parameter list. We use some heuristics to detect this
+    # and adjust the parameters (without importing pathlib)
+    if type(file).__name__ == '_NormalAccessor':
+        file, flags, mode, dir_fd = flags, mode, dir_fd, None
+
     if dir_fd is not None:
         fd = __original_open__(file, flags, mode, dir_fd=dir_fd)
     else:
--- a/tests/os_test.py
+++ b/tests/os_test.py
@@ -1,0 +1,11 @@
+import eventlet
+import six
+
+
+if six.PY3:
+    def test_pathlib_open_issue_534():
+        pathlib = eventlet.import_patched('pathlib')
+        path = pathlib.Path(__file__)
+        with path.open():
+            # should not raise
+            pass


orsinium:
Thank you for your work! Smart solution.

davidszotten:
Thanks @jstasiak . Any chance of a point version release with this fix to unblock kombu? Anything I can do to help with that?

jstasiak:
If you could prepare something like https://github.com/eventlet/eventlet/pull/572 for version 0.25.1 I can merge and release this within an hour.

davidszotten:
sure, raise #585

jstasiak:
Done, just released now.

davidszotten:
Thanks!

PR #587, opened 2019-08-26
wsgi: Fix header capitalization on py3

tipabu:
`str.capitalize()` and `str.upper()` respect unicode capitalization rules on py3, while py2 just translates `a-z` to `A-Z`.

At best, this may cause confusion and unexpected behaviors, such as when `'\xdf'` (a Latin1-encoded ß) becomes `'SS'`; at worst, this causes `UnicodeEncodeErrors` and the server fails to reply, such as when `'\xff'` (a Latin1-encoded ÿ) becomes `'\u0178'` which does not map back into Latin1.

Now, convert everything to bytes before capitalizing so just `a-z` and `A-Z` are affected on both py2 and py3.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -554,8 +554,15 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
             # Per HTTP RFC standard, header name is case-insensitive.
             # Please, fix your client to ignore header case if possible.
             if self.capitalize_response_headers:
+                if six.PY2:
+                    def cap(x):
+                        return x.capitalize()
+                else:
+                    def cap(x):
+                        return x.encode('latin1').capitalize().decode('latin1')
+
                 response_headers = [
-                    ('-'.join([x.capitalize() for x in key.split('-')]), value)
+                    ('-'.join([cap(x) for x in key.split('-')]), value)
                     for key, value in response_headers]
 
             headers_set[:] = [status, response_headers]
--- a/tests/wsgi_test.py
+++ b/tests/wsgi_test.py
@@ -190,7 +190,7 @@ def read_http(sock):
         x = x.strip()
         if not x:
             continue
-        key, value = bytes_to_str(x).split(':', 1)
+        key, value = bytes_to_str(x, encoding='latin1').split(':', 1)
         key = key.rstrip()
         value = value.lstrip()
         key_lower = key.lower()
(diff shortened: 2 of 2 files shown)


eventlet/wsgi.py
```diff
@@ -554,8 +554,15 @@ def start_response(status, response_headers, exc_info=None):
             # Per HTTP RFC standard, header name is case-insensitive.
             # Please, fix your client to ignore header case if possible.
             if self.capitalize_response_headers:
+                if six.PY2:
+                    def cap(x):
+                        return x.capitalize()
+                else:
+                    def cap(x):
+                        return x.encode('latin1').capitalize().decode('latin1')
```
4383, line 562: What do you think about reuse the encoding given in the response header (if exist), and why not allow the possibility to pass the encoding as a function param?

tipabu:
No, I hadn't bothered writing up an issue for it yet.

The choice of Latin-1 comes down to [PEP-3333](https://www.python.org/dev/peps/pep-3333/#unicode-issues) which dictates that any WSGI environment variables or `start_response` arguments be Latin-1. It's actually been a pretty constant [thorn](https://github.com/openstack/swift/commit/06e4533c2e04333e6215fcaed969118cb758d3d3) [in](https://github.com/openstack/swift/commit/76fde89261e1940daadb720c41df1a3595314a97) [my](https://github.com/openstack/swift/commit/60c27d3aef6e7412b41e04445d72aa62ea9883d3) [side](https://github.com/openstack/swift/commit/b5c9dc1c9f68afe8083b74eaf8080e17a9ef8f8b) while porting a sizeable project that uses UTF-8 to Python3.

And FWIW, [when we write headers](https://github.com/eventlet/eventlet/blob/v0.25.1/eventlet/wsgi.py#L498) we use `six.b()` which [uses Latin-1](https://github.com/benjaminp/six/blob/1.15.0/six.py#L627-L628).

4383:
> No, I hadn't bothered writing up an issue for it yet.
> 
> The choice of Latin-1 comes down to [PEP-3333](https://www.python.org/dev/peps/pep-3333/#unicode-issues) which dictates that any WSGI environment variables or `start_response` arguments be Latin-1. It's actually been a pretty constant [thorn](https://github.com/openstack/swift/commit/06e4533c2e04333e6215fcaed969118cb758d3d3) [in](https://github.com/openstack/swift/commit/76fde89261e1940daadb720c41df1a3595314a97) [my](https://github.com/openstack/swift/commit/60c27d3aef6e7412b41e04445d72aa62ea9883d3) [side](https://github.com/openstack/swift/commit/b5c9dc1c9f68afe8083b74eaf8080e17a9ef8f8b) while porting a sizeable project that uses UTF-8 to Python3.
> 
> And FWIW, [when we write headers](https://github.com/eventlet/eventlet/blob/v0.25.1/eventlet/wsgi.py#L498) we use `six.b()` which [uses Latin-1](https://github.com/benjaminp/six/blob/1.15.0/six.py#L627-L628).

So the usage of `latin-1` make sense here.

Thanks for your responses.

PR #590, opened 2019-10-31
Fix wsgi.server shutdown for in-flight requests

dbishop:
While doing some work on OpenStack Swift, I noticed that in-flight
requests were getting their sockets closed by eventlet when the listen
socket was closed.  In that case, there is a server process which bound
the listen socket and N fork'ed off worker processes which all run the
wsgi.server loop calling accept() and handling requests.

I noticed that nothing was ever setting connection state to
STATE_REQUEST (i.e. unused constant in previous patch).  Upon closer
inspection, it turns out that the socket cleanup code for STATE_IDLE
sockets got applied to the socket with an in-progress request.  This
caused code using the socket to get an EPIPE and exit the request's
greenthread without completing the request.

With some instrumentation in the code, without this patch's fix, the new
test observed the code doing this:

(7544) wsgi starting up on http://127.0.0.1:36745
(7544) calling accept
(7544) accept got <eventlet.greenio.base.GreenSocket object at 0x7f7650d73290>, ('127.0.0.1', 50194)
(7544) accepted ('127.0.0.1', 50194)
(7544) completed req ('127.0.0.1', 50194)
(7544) calling accept
(7544) got exception 22
(7544) re-raising!
(7544) wsgi exiting, calling pool.waitall()
  File "/tmp/swiftstack-eventlet/eventlet/wsgi.py", line 602, in handle_one_response
    write(b'')
  File "/tmp/swiftstack-eventlet/eventlet/wsgi.py", line 541, in write
    wfile.flush()
  File "/usr/lib64/python2.7/socket.py", line 303, in flush
    self._sock.sendall(view[write_offset:write_offset+buffer_size])
  File "/tmp/swiftstack-eventlet/eventlet/greenio/base.py", line 403, in sendall
    tail = self.send(data, flags)
  File "/tmp/swiftstack-eventlet/eventlet/greenio/base.py", line 397, in send
    return self._send_loop(self.fd.send, data, flags)
  File "/tmp/swiftstack-eventlet/eventlet/greenio/base.py", line 384, in _send_loop
    return send_method(data, *args)
error: [Errno 32] Broken pipe

127.0.0.1 - - [31/Oct/2019 01:08:05] "PUT /foo-bar HTTP/1.1" 200 0 0.000696
(7544) wsgi exited, is_accepting=True

In the new tests, not having the fix manifests as all greenthreads
exiting before the client finishes the in-flight request.

For the WebsocketWSGI app, it has a private class attribute that tells
the wsgi.server() code to consider its "reqeusts" always idle, which I
believe was the primary intent of the original patch.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -390,6 +390,8 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
             self.handle_one_request()
             if self.conn_state[2] == STATE_CLOSE:
                 self.close_connection = 1
+            else:
+                self.conn_state[2] = STATE_IDLE
             if self.close_connection:
                 break
 
@@ -417,6 +419,9 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
             self.protocol_version = self.server.max_http_version
 
         self.raw_requestline = self._read_request_line()
+        # WebSocketWSGI needs us, here, to consider the connection always idle
+        if not getattr(self.server.app, '_WSGI_APP_ALWAYS_IDLE', False):
+            self.conn_state[2] = STATE_REQUEST
         if not self.raw_requestline:
             self.close_connection = 1
             return
--- a/tests/wsgi_test.py
+++ b/tests/wsgi_test.py
@@ -101,6 +101,17 @@ def chunked_post(env, start_response):
         return [x for x in iter(lambda: env['wsgi.input'].read(1), b'')]
 
 
+def echo_server(env, start_response):
+    exp_len = int(env['CONTENT_LENGTH'])
+    got_body = env['wsgi.input'].read(exp_len)
+    start_response('200 OK', [
+        ('Content-type', 'application/octet-stream'),
+        ('Content-Length', str(len(got_body))),
+        ('X-Path', env['PATH_INFO']),
+    ])
+    return [got_body]
+
+
 def already_handled(env, start_response):
     start_response('200 OK', [('Content-type', 'text/plain')])
     return wsgi.ALREADY_HANDLED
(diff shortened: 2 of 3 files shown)


tests/wsgi_test.py
```diff
@@ -1747,21 +1758,127 @@ def test_log_disable(self):
...
+            'Content-Length: %d\r\n\r\n' % len(body))
+        sock.sendall(method_and_headers.encode('utf8'))
+        eventlet.sleep(0)
+
+        listen_sock.shutdown(socket.SHUT_RDWR)
+        eventlet.sleep(0)
+        listen_sock.close()
+        eventlet.sleep(0)
+        try:
+            with eventlet.Timeout(1):
```
clayg: this is a pretty good test!  it forces the test to slow down and make sure the socket doesn't get closed.  If our test infra was running too slow it's possible we'd let a regression through, but a false negative is certainly less brittle.

tests/wsgi_test.py
```diff
@@ -1747,21 +1758,127 @@ def test_log_disable(self):
...
+        self.assertEqual('/foo-bar', result.headers_lower['x-path'])
+        eventlet.sleep(0)
+
+        try:
+            eventlet.sleep(0)
+            with eventlet.Timeout(1):
+                eventlet.sleep(0)
+                pool.waitall()
+        except (eventlet.Timeout, Exception):
+            assert False, self.logfile.getvalue()
```
clayg: could we assert that a send to the socket here raises like a ioerror (epipe?) and it's for-realzy closed at this point?

dbishop: Youbetcha!

tests/wsgi_test.py
dbishop: I bumped it to 2s to reduce the chance of missing a regression.

eventlet/wsgi.py
```diff
@@ -417,6 +419,9 @@ def handle_one_request(self):
             self.protocol_version = self.server.max_http_version
 
         self.raw_requestline = self._read_request_line()
+        # WebSocketWSGI needs us, here, to consider the connection always idle
+        if not getattr(self.server.app, '_WSGI_APP_ALWAYS_IDLE', False):
+            self.conn_state[2] = STATE_REQUEST
```
tipabu: Oh hey, we weren't actually _using that_ before, were we...

dbishop:
looks much better to me!

temoto (approved):
thanks

tipabu:
Resolved conflicts and re-opened as #912 (since the source repo got deleted).

PR #598, opened 2020-02-12
Fix #508: Py37 Deadlock ThreadPoolExecutor

### LINKED ISSUE #508: py3.7 deadlock with monkeypatch of stdlib thread modules + use of ThreadPoolExecutor
There's a py3.7 deadlock when eventlet.patcher.monkey_patch(thread=True) is used with concurrent.futures.ThreadPoolExecutor().

Here's some sample code that deadlocks: https://bugs.python.org/file47709/bpo34173-recreate.py

Notice that it does not deadlock if the monkey_patch is removed.

I'm not sure if this is a bug in eventlet or concurrent.futures of the py3.7 standard library. I also have a py3.7 issue opened here: https://bugs.python.org/issue34173

If I switch /usr/lib/python3.7/concurrent/futures/thread.py back to use queue.Queue() (which was the case for py3.6) the deadlock goes away. This was changed in py3.7 from Queue to SimpleQueue:
```
- self._work_queue = queue.Queue()
+ self._work_queue = queue.SimpleQueue()
```

Akrog:
Python 3.7 and later implement queue.SimpleQueue in C, causing a
deadlock when using ThreadPoolExecutor with eventlet.

To avoid this deadlock we now replace the C implementation with the
Python implementation on monkey_patch for Python versions 3.7 and
higher.

Fix #508

### DIFF
--- a/tests/isolated/patcher_threadpoolexecutor.py
+++ b/tests/isolated/patcher_threadpoolexecutor.py
@@ -1,0 +1,18 @@
+# Issue #508: test ThreadPoolExecutor with monkey-patching
+__test__ = False
+
+if __name__ == '__main__':
+    import eventlet
+    eventlet.monkey_patch()
+
+    import sys
+
+    # Futures is only included in 3.2 or later
+    if sys.version_info >= (3, 2):
+        from concurrent import futures
+
+        with futures.ThreadPoolExecutor(max_workers=1) as executor:
+            future = executor.submit(pow, 2, 3)
+            res = future.result()
+        assert res == 8, '2^3 should be 8, not %s' % res
+    print('pass')
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -328,6 +328,12 @@ def monkey_patch(**on):
         import threading
         threading.RLock = threading._PyRLock
 
+    # Issue #508: Since Python 3.7 queue.SimpleQueue is implemented in C,
+    # causing a deadlock.  Replace the C implementation with the Python one.
+    if sys.version_info >= (3, 7):
+        import queue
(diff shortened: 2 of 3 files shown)


Akrog:
CI test failures are caused by the incompatibility of the six version with the virtualenv version, as [can be seen in the logs](https://travis-ci.org/eventlet/eventlet/jobs/649431675#L319).

tests/isolated/patcher_threadpoolexecutor.py
```diff
@@ -0,0 +1,18 @@
...
+
+    # Futures is only included in 3.2 or later
+    if sys.version_info >= (3, 2):
+        from concurrent import futures
+
+        with futures.ThreadPoolExecutor(max_workers=1) as executor:
+            future = executor.submit(pow, 2, 3)
+            res = future.result()
+        assert res == 8, '2^3 should be 8, not %s' % res
+    print('pass')
```
tipabu, line 18: This feels like a bit of an indirect test; something like
```
import eventlet
eventlet.monkey_patch()
import threading
try:
    import queue
except ImportError:
    import Queue as queue  # py2

def tgt(q):
    print(q.get())

# SimpleQueue is sufficient, but was only added in 3.7
q = getattr(queue, 'SimpleQueue', queue.Queue)()
t = threading.Thread(target=tgt, args=(q,))
t.start()
q.put('pass')
t.join()
```
seems a little more obvious -- could even tighten it to just use `eventlet.monkey_patch(all=False, thread=True)`.

OTOH, this more-clearly demonstrates the fact that we had a thing that was working when monkey-patched prior to 3.7 and stopped working after. \*shrug\*

PR #601, opened 2020-02-12
Remove unnecessary assignment in _recv_loop

Akrog:
In `_recv_loop` we assign `self.fd` to local variable `fd`, but then we don't use it within the method, so we can remove it.


Akrog:
CI test failures are caused by the incompatibility of the six version with the virtualenv version, as [can be seen in the logs](https://travis-ci.org/eventlet/eventlet/jobs/649451220#L321).

Issue #614, opened 2020-05-29
test_isolate_from_socket_default_timeout fails on SPARC


mgorny:
`test_isolate_from_socket_default_timeout` fails on SPARC due to timeout.

```
$ nosetests tests/tpool_test.py 
.............................Program /home/mgorny/eventlet/tests/isolated/tpool_isolate_socket_default_timeout.py output:
---

---
F.
======================================================================
FAIL: tests.tpool_test.test_isolate_from_socket_default_timeout
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mgorny/eventlet/tests/__init__.py", line 355, in run_python
    output, _ = p.communicate(timeout=timeout)
  File "/usr/lib/python3.7/subprocess.py", line 964, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.7/subprocess.py", line 1716, in _communicate
    self._check_timeout(endtime, orig_timeout, stdout, stderr)
  File "/usr/lib/python3.7/subprocess.py", line 1011, in _check_timeout
    stderr=b''.join(stderr_seq) if stderr_seq else None)
subprocess.TimeoutExpired: Command '['/usr/bin/python3.7', '/home/mgorny/eventlet/tests/isolated/tpool_isolate_socket_default_timeout.py']' timed out after 1 seconds

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/mgorny/eventlet/tests/tpool_test.py", line 369, in test_isolate_from_socket_default_timeout
    tests.run_isolated('tpool_isolate_socket_default_timeout.py', timeout=1)
  File "/home/mgorny/eventlet/tests/__init__.py", line 381, in run_isolated
    run_python(prefix + path, **kwargs)
  File "/home/mgorny/eventlet/tests/__init__.py", line 361, in run_python
    assert False, 'timed out'
AssertionError: timed out

----------------------------------------------------------------------
Ran 31 tests in 6.873s

FAILED (failures=1)
```

It seems that the default value of 1 s is probably too slow for non-x86 hardware.

```
$ time python /home/mgorny/eventlet/tests/isolated/tpool_isolate_socket_default_timeout.py
pass

real	0m1.894s
user	0m1.590s
sys	0m0.110s
```

[...]

PR #615, opened 2020-05-29
Fixes for the test suite on SPARC

mgorny:
This PR combines 4 fixes needed for the test suite to pass on SPARC. Don't ask me why Linux kernel behaves so differently between architectures.

1. Fixes the test case wrongly assuming that `F_SETFL` will return flags. It always returns 0, so we need to call `F_GETFL` explicitly to verify the new flags. While this technically doesn't break anything (the `assert` always passes), the fix made it easier to figure out why the test was failing.
2. Fixes disabling non-blocking mode to unset `O_NDELAY` as well. Unsetting just `O_NONBLOCK` and leaving `O_NDELAY` didn't do anything on SPARC (the flags stayed as-is).
3. Fixes the assertion when setting non-blocking mode to assume that `O_NDELAY` may be set as well as `O_NONBLOCK`. This is what happens on SPARC.
4. Increases test timeout per #614. I'm not sure if this is the correct solution, or if the test needs a major refactoring but expecting it to pass under 1 second when it takes almost 2 seconds to import modules is impossible.

### DIFF
--- a/tests/greenio_test.py
+++ b/tests/greenio_test.py
@@ -925,7 +925,10 @@ def test_set_nonblocking():
     assert orig_flags & os.O_NONBLOCK == 0
     greenio.set_nonblocking(sock)
     new_flags = fcntl.fcntl(fileno, fcntl.F_GETFL)
-    assert new_flags == (orig_flags | os.O_NONBLOCK)
+    # on SPARC, O_NDELAY is set as well, and it might be a superset
+    # of O_NONBLOCK
+    assert (new_flags == (orig_flags | os.O_NONBLOCK)
+            or new_flags == (orig_flags | os.O_NONBLOCK | os.O_NDELAY))
 
 
 def test_socket_del_fails_gracefully_when_not_fully_initialized():
--- a/tests/greenio_test.py
+++ b/tests/greenio_test.py
@@ -634,7 +634,8 @@ class TestGreenSocket(tests.LimitedTestCase):
         sock1 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         fd = sock1.fd.fileno()
         flags = fcntl.fcntl(fd, fcntl.F_GETFL)
-        flags = fcntl.fcntl(fd, fcntl.F_SETFL, flags & ~os.O_NONBLOCK)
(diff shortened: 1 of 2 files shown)


tests/greenio_test.py
```diff
@@ -634,7 +634,10 @@ def test_skip_nonblocking(self):
         sock1 = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         fd = sock1.fd.fileno()
         flags = fcntl.fcntl(fd, fcntl.F_GETFL)
-        flags = fcntl.fcntl(fd, fcntl.F_SETFL, flags & ~os.O_NONBLOCK)
+        # on SPARC, nonblocking mode sets O_NDELAY as well
+        fcntl.fcntl(fd, fcntl.F_SETFL, flags & ~(os.O_NONBLOCK
+                                                 | os.O_NDELAY))
+        flags = fcntl.fcntl(fd, fcntl.F_GETFL)
         assert flags & os.O_NONBLOCK == 0
```
jstasiak, line 641: I'm wondering if maybe we should `assert flags & (os.O_NONBLOCK | os.O_NDELAY) == 0` here as well?

tests/tpool_test.py
```diff
@@ -366,7 +366,7 @@ def test_leakage_from_tracebacks(self):
 
 
 def test_isolate_from_socket_default_timeout():
-    tests.run_isolated('tpool_isolate_socket_default_timeout.py', timeout=1)
+    tests.run_isolated('tpool_isolate_socket_default_timeout.py', timeout=5)
```
jstasiak, line 369: Would it be reasonable to conditionally set 5 here only on SPARC64, since it seems to be working fine elsewhere?

mgorny: I suspected that it could hit the same issue e.g. on ARM, or simply on systems under heavy load (we run tests as part of package installation, so other packages may be compiling in parallel).

tests/greenio_test.py
mgorny: I suppose that makes sense.

jstasiak:
Cheers!

PR #621, opened 2020-07-01
Fix compatibility with SSLContext usage >= Python 3.7

javacruft:
For SSL sockets created using the SSLContext class under Python >= 3.7,
eventlet incorrectly passes the context as '_context' to the top
level wrap_socket function in the native ssl module.

This causes:

  wrap_socket() got an unexpected keyword argument '_context'

as the context cannot be passed this way.

If a context is provided, use the underlying sslsocket_class to
wrap the socket, mirroring the implementation of the wrap_socket
method in the native SSLContext class.

Fixes issue #526

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -64,17 +64,28 @@ class GreenSSLSocket(_original_sslsocket):
             if not isinstance(sock, GreenSocket):
                 sock = GreenSocket(sock)
             with _original_ssl_context():
-                ret = _original_wrap_socket(
-                    sock=sock.fd,
-                    keyfile=keyfile,
-                    certfile=certfile,
-                    server_side=server_side,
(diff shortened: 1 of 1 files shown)


jstasiak:
Thank you. I think this one will require a unit test demonstrating the failure conditions, do you mind creating one?

tipabu:
The test at https://github.com/tipabu/eventlet/commit/7cac8379e0da4885ded3735107cc075e24611d43 should be a decent head-start -- I think I was maybe supposed to pass `ssl.PROTOCOL_TLS` to `ssl.SSLContext()` for the sake of py2, though.

jstasiak:
Merged as 02fa826d22d0df7f9379c3a1e5cd2192e0ffadd8 squashed with @tipabu's test.

PR #622, opened 2020-07-01
Fix misc SyntaxWarning's under Python 3.8

javacruft:
Use '==' instead of 'is':

 eventlet/db_pool.py:78: SyntaxWarning: "is" with a literal. Did you mean "=="?
   if self.max_age is 0 or self.max_idle is 0:


jstasiak:
Thanks!

PR #626, opened 2020-07-08
github linter

temoto:


temoto:
Currently, pylint is disabled and black is set to continue-on-error because there are just too many failures. Most pylint failures are also false positive.
But github/super-linter will catch syntax error in `.travis.yml` for example.

jstasiak:
I'm unsure about this. I think it'd good if the all the linting could be run locally before pushing to GH and the GH linter almost guarantees that's not possible, as it requires Docker on developers' machines. And the core of the linter for Python (pylint) is disabled in this patch anyway. And I don't really like the things pylint reports which is why I never pushed for its inclusion – are you actually using it for other things and like its behavior?

I'm a big fan of black but in a mode that enforces the formatting, maybe we should just reformat everything but this would generates conflicts with the existing pull requests. :/

Switching from Travis to GH Actions for CI could be nice to get Mac OS/Windows coverage on the other hand.

temoto:
ok thanks

PR #640, opened 2020-08-15
Test on Python 3.8

jayvdb:


jstasiak:
Allowing Python 2.7 failures is a separate issue, worth both discussing and explaining the motivation in the commit message (one-liner does help understanding why this is being done).

jayvdb:
Note https://github.com/eventlet/eventlet/pull/642 sites on top of this PR, and hits one of the bigger py27 breakages I found, but there are lots of little breakages, and complex to fix, and cause increased complexity at least in the test suite, so I can guess it is going to be a hard unrewarded battle getting tests working on py27 again.

jstasiak:
A change allowing Python 2.7 failures in a PR titles "Test on Python 3.8" is out of scope.

temoto:
Thanks for contribution, py3.8 should have been tested for a while.

Please remove any changes to py27, py37 and try py38 without sudo. We don't obsess with green CI, only that patches must be helpful and not break something working.

jayvdb:
@temoto the change removing `dist: xenial` is because that is useless.  `dist: xenial` is superflous - Travis uses xenial or later by default now.

I'll move the py27 commit to a different PR.

temoto:
@jayvdb thanks for education.

Please don't send patches with allow_failures. I will fix the problem with py27, it's super simple, just have to silence one particular warning.

temoto:
And while dist:xenial may be superficial, there is still no need to change py37 config in "test of python 3.8"

tox.ini
```diff
@@ -63,6 +64,7 @@ deps =
     py27: subprocess32==3.2.7
     pypy: psycopg2cffi-compat==1.1
     py{27,35,36,37}: psycopg2-binary==2.7.4
+    py38: psycopg2-binary
```
temoto, line 67:
```suggestion
    py{27,35,36,37,38}: psycopg2-binary==2.7.4
```

does it work?

.travis.yml
```diff
@@ -18,9 +18,13 @@ matrix:
     - {python: 3.6, env: TOXENV=py36-poll}
     - {python: 3.6, env: TOXENV=py36-selects}
 
-    - {python: 3.7, sudo: required, dist: xenial, env: TOXENV=py37-epolls}
-    - {python: 3.7, sudo: required, dist: xenial, env: TOXENV=py37-poll}
-    - {python: 3.7, sudo: required, dist: xenial, env: TOXENV=py37-selects}
+    - {python: 3.7, sudo: required, env: TOXENV=py37-epolls}
```
temoto, line 21: not related to 3.8

temoto (changes requested):

tox.ini
jayvdb: No of course it doesnt work.  That is why I added this.

.travis.yml
jayvdb: It is routine tidy up.  I will split it into a separate commit in this PR.

jstasiak: Tidying up like this belongs in a separate PR as far as I'm concerned, pollutes the history otherwise.

.travis.yml
```diff
@@ -18,9 +18,13 @@ matrix:
...
     - {python: 3.6, env: TOXENV=py36-selects}
 
-    - {python: 3.7, sudo: required, dist: xenial, env: TOXENV=py37-epolls}
-    - {python: 3.7, sudo: required, dist: xenial, env: TOXENV=py37-poll}
-    - {python: 3.7, sudo: required, dist: xenial, env: TOXENV=py37-selects}
+    - {python: 3.7, env: TOXENV=py37-epolls}
+    - {python: 3.7, env: TOXENV=py37-poll}
+    - {python: 3.7, env: TOXENV=py37-selects}
+
+    - {python: 3.8, env: TOXENV=py36-epolls}
```
temoto, line 25: 3.8 ~ 36 typo

.travis.yml
temoto: I was angry at first, but few deep breaths later, it even makes sense. In the end it would be a 3.8 commit and tidy commit, like after proper merge (which is of course fast forward without merge commit). So github PR history is polluted, but git history is fine.

@jstasiak let's agree to make it quicker this one time.
@jayvdb please make unrelated changes in separate patches in future, no PR-on-PR thing. It may seem stupid rules, but keeping patches with small scope really helps to make fast and correct decisions. It's related to limited time we can invest into volunteer maintaining, I suppose you understand.

jstasiak: Sure, fair enough.

jayvdb:
Done

Note `tests.event_test.test_wait_timeout_ok` is an intermittent fault at https://travis-ci.org/github/eventlet/eventlet/jobs/719232704#L1686

tox.ini
temoto: thanks for pinning

temoto:
@jayvdb thanks. If you like to keep git commiter header, you may rebase on master, where py27 is also fixed. Otherwise I will rebase with your ack or in one hour.

temoto (approved):

PR #642, opened 2020-08-18
tests: Avoid using Python 3.3+ subprocess on py27

jayvdb:
This PR sits on top of https://github.com/eventlet/eventlet/pull/640 , making the overall CI pass even though there are still errors in the py27 jobs.


temoto:
Duplicate of #342

PR #645, opened 2020-09-23
Listener removes wrong guy

clayg:
I really only intended this to be a followup to @charmster 's comment on #498 

I'm not actually sure if this is better, but I do think it would also work


temoto (approved):
LGTM thanks a lot

PR #648, opened 2020-09-30
Clean up TypeError in __del__

tipabu:
At the end of the py2 tests, we would see (ignored) errors like

    Exception TypeError: "'NoneType' object is not callable" in <bound method _SocketDuckForFd.__del__ of _SocketDuckForFd:14> ignored

### DIFF
--- a/eventlet/greenio/py2.py
+++ b/eventlet/greenio/py2.py
@@ -207,7 +207,10 @@ class _SocketDuckForFd(object):
         was_closed = self._mark_as_closed()
         if was_closed:
             return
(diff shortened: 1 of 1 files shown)


tipabu:
Compare a [recent py27 build on master](https://travis-ci.org/github/eventlet/eventlet/jobs/733079289#L2130) vs the [same test job but for this change](https://travis-ci.org/github/eventlet/eventlet/jobs/731712386#L2129).

temoto (approved):
Thank you.

PR #649, opened 2020-09-30
Update .gitignore

harshsingh121098:


tipabu:
Thanks for the thought, but both `.tox` and `.coverage` should already be ignored by the `.*` rule at https://github.com/eventlet/eventlet/blob/v0.28.0/.gitignore#L4

PR #650, opened 2020-10-02
mqtt testing for IoT

galuh404:
script mqtt testing for IoT


jstasiak:
Can you provide some details and explanation? "script mqtt testing for IoT" is not enough when it's not clear what this does, why, and how it relates to this project.

temoto:
It's likely that hacktoberfest spam thing. I would be glad to be proven wrong. Reopen at will.

Issue #644, opened 2020-09-11
Monkey_patch throws AttributeError on Windows (using Celery and Eventlet)


Galsor:
Hello,

I am trying to run Celery with Eventlet as pool in a Django Project running periodic tasks.
When I run `celery worker --app=app -l info` everything is fine but when I run `celery worker --app=app-l info -P eventlet`
I got this exception: 

> Traceback (most recent call last):
>  File "c:\users\anaconda3\envs\paper_parser\lib\runpy.py", line 193, in _run_module_as_main  "__main__", mod_spec)
>  File "c:\users\anaconda3\envs\paper_parser\lib\runpy.py", line 85, in _run_code   exec(code, run_globals)
>  File "C:\Users\Anaconda3\envs\paper_parser\Scripts\celery.exe\__main__.py", line 7, in <module>
>  File "c:\users\anaconda3\envs\paper_parser\lib\site-packages\celery\__main__.py", line 14, in main
>    maybe_patch_concurrency()
>  File "c:\users\anaconda3\envs\paper_parser\lib\site-packages\celery\__init__.py", line 152, in maybe_patch_concurrency
>    patcher()
>  File "c:\users\anaconda3\envs\paper_parser\lib\site-packages\celery\__init__.py", line 109, in_patch_eventlet
>    eventlet.monkey_patch()
>  File "c:\users\anaconda3\envs\paper_parser\lib\site-packages\eventlet\patcher.py", line 334, in monkey_patch
>    fix_threading_active()
>  File "c:\users\anaconda3\envs\paper_parser\lib\site-packages\eventlet\patcher.py", line 331, in fix_threading_active
>    _os.register_at_fork(
>AttributeError: module 'os' has no attribute 'register_at_fork'

After a few research it seems that `os.register_at_fork` is only available in Unix systems. ([_source_in_python_doc_](https://docs.python.org/3.8/library/os.html#os.register_at_fork))

Is it due to recent changes? Eventlet seemed to be the preferred solution to make celery run with Windows.
Is there any workaround?

Configuration: `Windows10`, `Celery Version: 4.4.7`, `Eventlet Version: 0.27.0`, `Django Version: 3.1.1`, `django-celery-beat Version: 2.0.0`

Thanks,

temoto:
Yes, 115103d5608cbe8f15df10e27eba1644f5364e95 made eventlet incompatible with Windows.

[...]

PR #654, opened 2020-10-17
patcher: no os.register_at_fork on Windows

temoto:
https://github.com/eventlet/eventlet/issues/644

Eventlet 0.27.0 and 0.28.0 were incompatible with Windows (and other non-Unix) platforms


tipabu (approved):
Looks entirely reasonable, and I remember thinking at the time that I should maybe do a `hasattr` check. Sorry for the breakage.

Issue #651, opened 2020-10-05
SSLSocket.accept fails when monkey-patched


Xandaros:
Currently setting up a fairly simple SSL server and I wanted to use eventlet to spawn handler threads for the clients. Unfortunately, if I use `eventlet.monkey_patch`, the code stops working (with no other modification).

```
Traceback (most recent call last):
  File "./test.py", line 45, in <module>
    main()
  File "./test.py", line 38, in main
    client_sock, addr = ssl_sock.accept()
  File "/home/xandaros/workspace/company/test/server/venv/lib/python3.8/site-packages/eventlet/green/ssl.py", line 405, in accept
    new_ssl = type(self)(
  File "/home/xandaros/workspace/company/test/server/venv/lib/python3.8/site-packages/eventlet/green/ssl.py", line 79, in __new__
    ret = _original_wrap_socket(
  File "/usr/lib/python3.8/ssl.py", line 1393, in wrap_socket
    raise ValueError("certfile must be specified for server-side "
ValueError: certfile must be specified for server-side operations
```

The monkey-patching happens right at the top of the file:
```python
import eventlet
eventlet.monkey_patch()
```

To create the SSL socket, I don't use `ssl.wrap_socket`, but instead create my own SSL socket and load the certificates:
```python
    context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
    context.load_cert_chain(certfile="server.crt", keyfile="server.key")
```

Then I wrap the socket:
```python
    sock = socket.create_server(('0.0.0.0', 1234))
    ssl_sock = context.wrap_socket(sock, server_side=True)
```

And finally, I call `ssl_sock.accept()`, which causes aforementioned error when a client connects; but only if the environment is monkey-patched.

temoto:
@Xandaros thank you, that's a shame to fail such basic interaction. Sorry.

I've found the culprit of problem, it's a one line fix, but I'm afraid it may potentially break a lot of other things. So asking you to participate in as extensive testing as you can afford before it may be released to public. Please read shortest instruction and report problems here https://github.com/eventlet/eventlet/pull/655

Xandaros:
Due to time constraints I had to move away from eventlet, so I don't actually have a use-case to test this on, sorry.

At this point it would be too much work to move everything back to eventlet, especially when using an unpublished version we wouldn't be able to deploy.

temoto:
[...]

PR #655, opened 2020-10-18
ssl: context wrapped listener failed to supply _context in accept()

temoto:
https://github.com/eventlet/eventlet/issues/651

Seems correct, but could potentially break a lot of code, so calling for extensive public testing.

`pip install -U https://github.com/eventlet/eventlet/archive/i651.zip` and try to run your software.

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -404,14 +404,11 @@ class GreenSSLSocket(_original_sslsocket):
 
         new_ssl = type(self)(
             newsock,
-            keyfile=self.keyfile,
-            certfile=self.certfile,
             server_side=True,
-            cert_reqs=self.cert_reqs,
-            ssl_version=self.ssl_version,
-            ca_certs=self.ca_certs,
             do_handshake_on_connect=False,
-            suppress_ragged_eofs=self.suppress_ragged_eofs)
+            suppress_ragged_eofs=self.suppress_ragged_eofs,
+            _context=self._context,
+        )
         return (new_ssl, addr)
 
     def dup(self):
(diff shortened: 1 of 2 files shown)


temoto:
Last version fails on Python 2.7 - 3.6, works on 3.7+.

I'm tired today, will continue later. Error with keyfile argument.

```
  File "/home/travis/build/eventlet/eventlet/tests/ssl_test.py", line 342, in test_context_wrapped_accept
    peer, _ = server_tls.accept()
  File "/home/travis/build/eventlet/eventlet/eventlet/green/ssl.py", line 407, in accept
    keyfile=self.keyfile,
AttributeError: 'GreenSSLSocket' object has no attribute 'keyfile'
```

jstasiak (approved):
Works for me on Python 3.8.5 and looks reasonable.

temoto:
Works now.

jstasiak:
How does the code work on Python older than 3.8 with `keyfile`, `certfile` not being passed through?

temoto:
@jstasiak 
> How does the code work on Python older than 3.8 with `keyfile`, `certfile` not being passed through?

Sorry, it's a working black magic for me, currently. Test passes, so apparently they are read from context. So I was wondering is it redundant to pass them at all. And can you believe it, all key/cert related kwargs were redundant. It passes py2.7 and 3.7 without extra 3.7 condition. Hope it will pass all tests.

Thank you.

temoto:
@jstasiak and for shorter diff too, I've only noticed it too late after push, was too much excited about dropping cert/key kwargs. Done.

jstasiak (approved):
Yeah it was just a detail, LGTM regardless.

PR #656, opened 2020-10-19
travis: update according to their validator

temoto:
- deprecated key sudo (The key `sudo` has no effect anymore.)
- missing os/dist
- key matrix is an alias for jobs, using jobs
- eventlet slack channel was abandoned

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
-sudo: required
+os: linux
+dist: xenial
 language: python
 
-matrix:
+jobs:
   fast_finish: true
   include:
     - {python: 3.6, env: TOXENV=pep8}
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,7 +1,8 @@
-sudo: required
+os: linux
+dist: xenial
```
tipabu, line 2: This matches what we've got currently, but I wonder if we should move to bionic...

tipabu (approved):

.travis.yml
temoto: Yes, I wanted to add another job with bionic and their default Python but then I thought we specify exact Python version anyway and Python is provided by special means of Travis, so dist should play little to no difference anyway.

PR #657, opened 2020-10-19
hubs: drop pyevent hub

temoto:
This is a big change, but I honestly don't think anyone was using pyevent hub.


temoto:
And such round commit id `d9000b00` definitely proves this is the right rebase attempt.

jstasiak (commented):
I think it'd be better to have it deprecated for a release or two, but I also don't think it's a big deal.

temoto:
> I think it'd be better to have it deprecated for a release or two, but I also don't think it's a big deal.

I thought it was already broken, spent about hour to satisfy effin libevent and pyevent versions and build flags. And what do you know? pyevent hub still works! Hard to build but works after that.

So that deserves deprecation time, yes.

This PR is to be reopened after version 0.31.

temoto:
merged in master

PR #658, opened 2020-10-19
doc index: (mostly) remove dead links

temoto:
Eventlet channel on Freenode IRC is probably still alive, so I left it in bottom links list, but it's also probably ghost, so removed whole Discussion section, every other communication was dead for a while now.

### DIFF
--- a/doc/real_index.html
+++ b/doc/real_index.html
@@ -65,18 +65,6 @@ eventlet==x.y</pre></code></p>
 </ul>
 
 
-<h3>Discussion</h3>
-
-<ul>
-<li>
-  <p><a class="reference external" target="_blank" href="https://lists.secondlife.com/cgi-bin/mailman/listinfo/eventletdev">eventletdev at lists.secondlife.com</a></p>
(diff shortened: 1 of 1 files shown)


doc/real_index.html
```diff
@@ -65,18 +65,6 @@ <h3>Installation</h3>
 </ul>
 
 
-<h3>Discussion</h3>
-
-<ul>
-<li>
-  <p><a class="reference external" target="_blank" href="https://lists.secondlife.com/cgi-bin/mailman/listinfo/eventletdev">eventletdev at lists.secondlife.com</a></p>
```
tipabu, line 72: We should maybe update the topic in IRC, too:
```
Eventlet discussion, ask away! | http://eventlet.net/ | Mailing list: http://urlenco.de/usbo |Latest release: http://pypi.python.org/pypi/eventlet
```
Looks like urlenco.de isn't a thing anymore?

tipabu (approved):

jstasiak (approved):

doc/real_index.html
temoto: I have registered project channel and updated topic.

@jstasiak @nat-goodspeed or anybody who needs access to something related to eventlet channel on IRC, please write me.

Issue #660, opened 2020-10-20
maximum recursion depth exceeded error in `monkey_patch`


lobsterkatie:
Hello!

[This PR](https://github.com/eventlet/eventlet/pull/654), merged yesterday, removed a version check on Python, and now running under 2.7 (which we do as part of our test suite) throws this error:

```
_______ ERROR at setup of test_transport_works[eventlet-True-close-True] _______
tests/conftest.py:292: in maybe_monkeypatched_threading
    eventlet.monkey_patch()
.tox/py2.7/lib/python2.7/site-packages/eventlet/patcher.py:316: in monkey_patch
    _os = original('os')
.tox/py2.7/lib/python2.7/site-packages/eventlet/patcher.py:201: in original
    real_mod = __import__(modname, {}, {}, modname.split('.')[:-1])
.tox/py2.7/lib/python2.7/fnmatch.py:40: in fnmatch
    import os
.tox/py2.7/lib/python2.7/fnmatch.py:40: in fnmatch
    import os
E   RuntimeError: maximum recursion depth exceeded in __instancecheck__
!!! Recursion detected (same locals & position)
```

That version check was added [here](https://github.com/eventlet/eventlet/commit/115103d5608cbe8f15df10e27eba1644f5364e95). Is it incompatible with the change made yesterday, or could it be added back in?

Thanks!

temoto:
Fixed in 0.29.1

lobsterkatie:
Thanks for the quick turn-around!

PR #661, opened 2020-10-21
patcher: [py27] recursion error in pytest/python2.7 installing register_at_fork

temoto:
https://github.com/eventlet/eventlet/issues/660
https://github.com/getsentry/sentry-python/pull/880

Could not reproduce problem without pytest, so normal production code was probably safe.

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -1,5 +1,11 @@
 import imp
 import sys
+try:
+    # Only for this purpose, it's irrelevant if `os` was already patched.
+    # https://github.com/eventlet/eventlet/pull/661
+    from os import register_at_fork
+except ImportError:
+    register_at_fork = None
 
 import eventlet
 import six
@@ -298,6 +304,7 @@ def monkey_patch(**on):
             # tell us whether or not we succeeded
             pass
 
+    _threading = original('threading')
     imp.acquire_lock()
     try:
         for name, mod in modules_to_patch:
@@ -313,11 +320,13 @@ def monkey_patch(**on):
                 if hasattr(orig_mod, attr_name):
                     delattr(orig_mod, attr_name)
 
-            _os = original('os')
-            if name == 'threading' and hasattr(_os, 'register_at_fork'):
+            # https://github.com/eventlet/eventlet/issues/592
+            if name == 'threading' and register_at_fork:
                 def fix_threading_active(
-                    _global_dict=original('threading').current_thread.__globals__,
-                    _patched=orig_mod
+                    _global_dict=_threading.current_thread.__globals__,
+                    # alias orig_mod as patched to reflect its new state
+                    # https://github.com/eventlet/eventlet/pull/661#discussion_r509877481
+                    _patched=orig_mod,
                 ):
                     _prefork_active = [None]
 
@@ -328,7 +337,7 @@ def monkey_patch(**on):
                     def after_fork():
                         _global_dict['_active'] = _prefork_active[0]
 
-                    _os.register_at_fork(
+                    register_at_fork(
                         before=before_fork,
                         after_in_parent=after_fork)
                 fix_threading_active()


temoto:
Guys I tried to create test, was hoping for trigger if os/threading was previously imported or double execute anything, but no luck. Maybe pytest does something wonderful. Can't put too much time into this. But I need working python-sentry :-)

eventlet/patcher.py
```diff
@@ -229,7 +231,6 @@ def monkey_patch(**on):
 
     It's safe to call monkey_patch multiple times.
     """
-
```
jstasiak, line 232: This removal (and and `#_os = original('os')` line above) would be better left out of this PR.

eventlet/patcher.py
```diff
@@ -1,5 +1,7 @@
 import imp
+import os
```
jstasiak, line 2: I don't think we can be sure that at the time of accessing this `os` it's not patched already (multiple `monkey_patch` calls are probably legal if they patch separate things?), correct me if I'm wrong

jstasiak (commented):
LGTM with one comment.

eventlet/patcher.py
temoto: Yes, it even broke pep8 checker. Sorry, wasn't careful after trying random stuff with testing.

eventlet/patcher.py
temoto: Ya, but `register_at_fork` is not patched, so either system or patched `os` of same value for this particular change.

jstasiak: Fair enough. I'd import `register_at_fork` only from the module to not leave the whole maybe-patched-maybe-not `os` around for someone to use accidentally.

jstasiak (approved):

eventlet/patcher.py
```diff
@@ -313,10 +315,9 @@ def monkey_patch(**on):
                 if hasattr(orig_mod, attr_name):
                     delattr(orig_mod, attr_name)
 
-            _os = original('os')
-            if name == 'threading' and hasattr(_os, 'register_at_fork'):
+            if name == 'threading' and hasattr(os, 'register_at_fork'):
                 def fix_threading_active(
-                    _global_dict=original('threading').current_thread.__globals__,
+                    _global_dict=_threading.current_thread.__globals__,
                     _patched=orig_mod
```
temoto, line 321: @tipabu it's called patched but refers to pseudo-original module (that is original at first `monkey_patch` call but patched at second). Which module `._active` should be used?

tipabu: Sorry, I think this was born out of confusion -- honestly, I'm probably *still* a fair bit confused. IIRC, the discussion at #592 made it seem that the core problem was the original threading's `current_thread` having the wrong notion of what the active threads were, and we needed to replace it with the patched active threads. So I waited until *after* `orig_thread` becomes the patched version (i.e., after all the `setattr`s), and that's why I named it as I did.

I'm also a little confused about the options you presented: "original" and "patched" make sense enough, but what do you mean by "current but with updated name"?

What's **really** confusing me now, though, is that I can't seem to reproduce the original problem. I comment out the call to `fix_threading_active` and my test still passes :confused: I *know* I tested that it would fail at the time. I also know I've upgraded my Python since then (I'm currently on 3.7.9 and 3.8.6 which were only released in August and September) -- maybe some fix made its way in upstream?

temoto: @tipabu so it should be patched, as named. Thank you, clear now.

Third option is related to `monkey_patch` getting "original" from `sys.modules`, it would actually contain patched modules if `monkey_patch` was ever to be run twice.

I believe you did test, that's frustrating to not reproduce, yeah. Even if there is upstream fix, we can't force users to upgrade Python.

eventlet/patcher.py
temoto: done

Issue #646, opened 2020-09-24
tests.patcher_test.test_fork_after_monkey_patch fails on py3.9


mgorny:
This looks like a regression due to a new test.

```
$ python3.9 --version
Python 3.9.0rc2
```

```
tests.patcher_test.test_fork_after_monkey_patch ... Program /tmp/portage/dev-python/eventlet-0.28.0/work/eventlet-0.28.0/tests/isolated/patcher_fork_after_monkey_patch.py output:
---
Exception ignored in: <function _after_fork at 0x7f2c51f7d430>
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 1492, in _after_fork
    thread._reset_internal_locks(False)
  File "/usr/lib/python3.9/threading.py", line 827, in _reset_internal_locks
    self._started._at_fork_reinit()
  File "/usr/lib/python3.9/threading.py", line 527, in _at_fork_reinit
    self._cond._at_fork_reinit()
  File "/usr/lib/python3.9/threading.py", line 253, in _at_fork_reinit
    self._lock._at_fork_reinit()
AttributeError: 'Semaphore' object has no attribute '_at_fork_reinit'
Exception ignored in: <function _after_fork at 0x7f2c51f4a670>
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 1492, in _after_fork
    thread._reset_internal_locks(False)
  File "/usr/lib/python3.9/threading.py", line 827, in _reset_internal_locks
    self._started._at_fork_reinit()
  File "/usr/lib/python3.9/threading.py", line 527, in _at_fork_reinit
    self._cond._at_fork_reinit()
  File "/usr/lib/python3.9/threading.py", line 253, in _at_fork_reinit
    self._lock._at_fork_reinit()
AttributeError: 'Semaphore' object has no attribute '_at_fork_reinit'
Exception ignored in: <function _after_fork at 0x7f2c50813550>
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 1486, in _after_fork
    thread._reset_internal_locks(True)
  File "/usr/lib/python3.9/threading.py", line 827, in _reset_internal_locks
    self._started._at_fork_reinit()
  File "/usr/lib/python3.9/threading.py", line 527, in _at_fork_reinit
    self._cond._at_fork_reinit()
  File "/usr/lib/python3.9/threading.py", line 253, in _at_fork_reinit
    self._lock._at_fork_reinit()
[...]

PR #664, opened 2020-11-03
py39: Add _at_fork_reinit method to Semaphores

tipabu:
CPython expects to be able to call such a method on `RLock`s, `Condition`s, and `Event`s in `threading`; since we may monkey-patch `threading` to use `Semaphore`s as locks, they need the method, too.

Addresses #646


temoto (approved):
Thank you.

jstasiak:
LGTM!

PR #666, opened 2020-11-04
Test with more Pythons

jstasiak:
CPython should be officially supported so let's verify this. Also we
need to see if PyPy 3 builds work (and, if not, how do they break).


jstasiak:
Likely blocked by https://github.com/psycopg/psycopg2/issues/1183.

jstasiak:
Oh nice, I forgot I opened this.

Issue #616, opened 2020-06-18
GreenSocket conflict with pysocks


terryzhu:
Combine `eventlet` and `pysocks`, it will cause `RecursionError: maximum recursion depth exceeded
`

## Code example
```
import eventlet
eventlet.monkey_patch()

import socket
import socks

def test_block_connect():
    s = socks.socksocket(socket.AF_INET, socket.SOCK_STREAM)
    s.connect(('example.com', 80))

gt = eventlet.spawn(test_block_connect)
gt.wait()
```
### Actual behaviour
```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/eventlet/hubs/hub.py", line 461, in fire_timers
    timer()
  File "/usr/lib/python3.6/site-packages/eventlet/hubs/timer.py", line 59, in __call__
    cb(*args, **kw)
  File "/usr/lib/python3.6/site-packages/eventlet/greenthread.py", line 221, in main
    result = function(*args, **kwargs)
  File "eventlet_sample.py", line 10, in test_block_connect
    s.connect(('example.com', 80))
  File "/usr/lib/python3.6/site-packages/socks.py", line 47, in wrapper
    return function(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/socks.py", line 774, in connect
    super(socksocket, self).settimeout(self._timeout)
  File "/usr/lib/python3.6/site-packages/eventlet/greenio/base.py", line 418, in settimeout
    self.setblocking(True)
  File "/usr/lib/python3.6/site-packages/socks.py", line 296, in setblocking
    self.settimeout(None)
  File "/usr/lib/python3.6/site-packages/socks.py", line 287, in settimeout
    super(socksocket, self).settimeout(self._timeout)
.....
.....
.....
.....
  File "/usr/lib/python3.6/site-packages/eventlet/greenio/base.py", line 418, in settimeout
    self.setblocking(True)
  File "/usr/lib/python3.6/site-packages/socks.py", line 296, in setblocking
    self.settimeout(None)
RecursionError: maximum recursion depth exceeded
```
### Root cause
[...]

PR #680, opened 2020-12-24
Fix GreenSocket conflict with pysocks

terryzhu:
#616 
https://github.com/Anorov/PySocks/issues/121


eventlet/greenio/base.py
```diff
@@ -414,7 +414,8 @@ def setblocking(self, flag):
 
     def settimeout(self, howlong):
         if howlong is None or howlong == _GLOBAL_DEFAULT_TIMEOUT:
-            self.setblocking(True)
+            if not (self.act_non_blocking == False and self._timeout is None):
```
temoto, line 417:
```suggestion
            if self.act_non_blocking or self._timeout is not None:
```

temoto:
Also needs a test.

PR #682, opened 2021-01-19
Stop reraising global exceptions in greendns (alternative to #811)

jjohnson42:
I was working on a user who was having a problem with memory consumption over time.

I found that that most of their memory consumption was the EAI_NONAME_ERROR object in greendns. It had accumulated massive memory through accumulating many stack frames each time it was raised.

In the user's case, they had getaddrinfo calls to an IPv4 name with AF_UNSPEC, so it tried INET6 and INET and squashed the error on INET6 (to be expected). However the raise would grow the reused exception object each time.

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -77,18 +77,19 @@ socket = _socket_nodns
 DNS_QUERY_TIMEOUT = 10.0
 HOSTS_TTL = 10.0
 
-EAI_EAGAIN_ERRORARGS = (socket.EAI_AGAIN, 'Lookup timed out')
-EAI_NONAME_ERRORARGS = (socket.EAI_NONAME, 'Name or service not known')
-EAI_NONAME_ERROR = socket.gaierror(*EAI_NONAME_ERRORARGS)
+_EAI_EAGAIN_ARGS = (socket.EAI_AGAIN, 'Lookup timed out')
+EAI_EAGAIN_ERROR = socket.gaierror(*_EAI_EAGAIN_ARGS)
+_EAI_NONAME_ARGS = (socket.EAI_NONAME, 'Name or service not known')
+EAI_NONAME_ERROR = socket.gaierror(*_EAI_NONAME_ARGS)
 # EAI_NODATA was removed from RFC3493, it's now replaced with EAI_NONAME
 # socket.EAI_NODATA is not defined on FreeBSD, probably on some other platforms too.
 # https://lists.freebsd.org/pipermail/freebsd-ports/2003-October/005757.html
 EAI_NODATA_ERROR = EAI_NONAME_ERROR
-EAI_NODATA_ERRORARGS = EAI_NONAME_ERRORARGS
+_EAI_NODATA_ARGS = _EAI_NONAME_ARGS
 if (os.environ.get('EVENTLET_DEPRECATED_EAI_NODATA', '').lower() in ('1', 'y', 'yes')
         and hasattr(socket, 'EAI_NODATA')):
-    EAI_NODATA_ERRORARGS = (socket.EAI_NODATA, 'No address associated with hostname')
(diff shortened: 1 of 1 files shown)


jstasiak:
This is *really* interesting  – can you provide some details as to how the stack frames are being accumulated? (I'm interested in what object reference chain leads to this)

jjohnson42:
I didn't quite get the tight understanding, and I failed to produce the same
behavior in my own copy of their environment and also failed to produce a unit test to demonstrate the issue.

However, guppy shortest path to the huge bunch of stackframes was always attached to that single instance of the socket.gaierror (well, tied between NODATA and NONAME, which is a reference to the same single exception).

I am not quite familiar enough to have dug in and more precisely understand the linkages in more detail. The DNS server triggered ENODATA on an IPv6 lookup, and that gets swallowed in getaddrinfo since they were IPv4 based anyway, but with the side-effect of bigger and bigger global EAI_NODATA_ERROR object. I did confirm low memory usage after change, and then reverted the change and the memory consumption shot back up as expected.

Sorry I don't quite understand things enough to be more detailed...

temoto:
@jjohnson42 memory consumption is good enough unit test for me. Check this out https://code.activestate.com/recipes/577504/ (linked in stdlib doc https://docs.python.org/3/library/sys.html#sys.getsizeof )

eventlet/support/greendns.py
```diff
@@ -77,15 +77,18 @@ def import_patched(module_name):
 DNS_QUERY_TIMEOUT = 10.0
 HOSTS_TTL = 10.0
 
-EAI_EAGAIN_ERROR = socket.gaierror(socket.EAI_AGAIN, 'Lookup timed out')
-EAI_NONAME_ERROR = socket.gaierror(socket.EAI_NONAME, 'Name or service not known')
+EAI_EAGAIN_ERRORARGS = (socket.EAI_AGAIN, 'Lookup timed out')
+EAI_NONAME_ERRORARGS = (socket.EAI_NONAME, 'Name or service not known')
+EAI_NONAME_ERROR = socket.gaierror(*EAI_NONAME_ERRORARGS)
```
temoto, line 82: - (the only major comment here) hide exception args from public API (`_FOO_ARGS`)
- `FOO_ERROR_ARGS` or `FOO_ARGS`
- `EAI_EAGAIN_ERROR` was deleted, probably good idea to leave it like NONAME, NODATA errors.

```suggestion
_EAI_EAGAIN_ARGS = (socket.EAI_AGAIN, 'Lookup timed out')
EAI_EAGAIN_ERROR = socket.gaierror(*EAI_EAGAIN_ARGS)
_EAI_NONAME_ARGS = (socket.EAI_NONAME, 'Name or service not known')
EAI_NONAME_ERROR = socket.gaierror(*EAI_NONAME_ARGS)
```

temoto (changes requested):

jjohnson42:
Ok, made suggested changes.

The problem I have is less about how to measure it, it's not being able to produce the same behavior. When I try to write a test case, the size of the exception does not seem to grow on re-raise. It clearly grew in their environment, and it clearly came from the exception and Frame objects.
So as far as me trying to synthetically test, it seems like a fine method, but in their environment, somehow it grew and I don't understand why it misbehaved there and I can't make it misbehave even knowing what was getting bigger.

eventlet/support/greendns.py
jjohnson42: The desired change is in the lattest commit

itamarst:
We want to merge either this or #811, but not both.

4383:
> We want to merge either this or #811, but not both.

I'd suggest to go with #811 as unit tests are also provided within.

4383:
Also #811 comes with a clear description of the bug and with a reproducer (#810)

itamarst:
Thank you; this has been addressed by #811.

Issue #567, opened 2019-05-14
SSL client certificates ValueError: check_hostname needs server_hostname argument


plockaby:
I'm not sure if this worked in previous versions and just broke or if it is has always been broken but here it goes. Using Python 3.6.8 and Eventlet 0.24.1. This code works without eventlet and successfully connects to a remote host using a client certificate:

```
import socket
import ssl

ctx = ssl.SSLContext()
ctx.verify_mode = ssl.CERT_REQUIRED
ctx.check_hostname = True
ctx.load_verify_locations("/usr/local/ssl/certs/ca.pem")
ctx.load_cert_chain("/usr/local/ssl/certs/near.example.com.pem")

with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
    with ctx.wrap_socket(sock, server_hostname="far.example.com") as ssock:
        ssock.connect(("far.example.com", 3278))
        ssock.sendall(bytes("hello", "utf-8"))
```

But if this is prefixed then it no longer works:

```
import eventlet
eventlet.monkey_patch()

import socket
import ssl
...
```

I get this exception:
```
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/home/plockaby/venv/lib/python3.6/site-packages/eventlet/green/ssl.py", line 329, in connect
    self.do_handshake()
  File "/home/plockaby/venv/lib/python3.6/site-packages/eventlet/green/ssl.py", line 260, in do_handshake
    super(GreenSSLSocket, self).do_handshake)
  File "/home/plockaby/venv/lib/python3.6/site-packages/eventlet/green/ssl.py", line 109, in _call_trampolining
    return func(*a, **kw)
  File "/usr/local/python-3.6.8/lib/python3.6/ssl.py", line 1077, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/local/python-3.6.8/lib/python3.6/ssl.py", line 692, in do_handshake
    raise ValueError("check_hostname needs server_hostname "
ValueError: check_hostname needs server_hostname argument
```

Any insight would be helpful.

mutesplash:
https://github.com/eventlet/eventlet/blob/05d613d8ed123b87f64f8d1987968c5bf825b07a/eventlet/green/ssl.py#L345

Changing this to:

sslobj = context._wrap_socket(self, server_side, server_hostname = self.server_hostname)

Seems to fix the problem using the example

plockaby:
[...]

PR #575, opened 2019-06-06
fix client certificates

plockaby:
This is for #567.

PR #705, opened 2021-05-11
fix check_hostname needs server_hostname argument by @plockaby

temoto:
Fixes https://github.com/eventlet/eventlet/issues/567

Moved from https://github.com/eventlet/eventlet/pull/575

### DIFF
--- a/tests/ssl_test.py
+++ b/tests/ssl_test.py
@@ -1,6 +1,7 @@
 import contextlib
 import random
 import socket
+import sys
 import warnings
 
 import eventlet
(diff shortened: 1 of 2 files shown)


tests/ssl_test.py
```diff
@@ -371,3 +372,24 @@ def client(addr):
         peer, _ = server_tls.accept()
         assert peer.recv(64) == expected
         peer.close()
+
+    @tests.skip_if(sys.version_info[:2] < (3, 7))
```
tipabu, line 376: Why only run the test on py37+? Looks like the issue was reported on py36...

tests/ssl_test.py
```diff
@@ -371,3 +372,24 @@ def client(addr):
...
+            sock, addr = listener.accept()
+            sock.recv(64)
+            sock.sendall(b"response")
+            sock.close()
+
+        listener = listen_ssl_socket()
+        server_coro = eventlet.spawn(serve, listener)
+        client = ssl.wrap_socket(
+            eventlet.connect(listener.getsockname()),
+            check_hostname=True,
```
tipabu, line 390: This doesn't seem to be how CPython's `ssl.wrap_socket()` works:
``` python
>>> import socket, ssl
>>> sock = socket.socket()
>>> sock.connect(('github.com', 443))
>>> ssock = ssl.wrap_socket(sock, check_hostname=True, verify_mode=ssl.CERT_REQUIRED)
TypeError: wrap_socket() got an unexpected keyword argument 'check_hostname'
```
WDYT about something like
``` diff
diff --git a/tests/ssl_test.py b/tests/ssl_test.py
index c5b5b49..dda82ff 100644
--- a/tests/ssl_test.py
+++ b/tests/ssl_test.py
@@ -385,11 +385,14 @@ class SSLTest(tests.LimitedTestCase):
 
         listener = listen_ssl_socket()
         server_coro = eventlet.spawn(serve, listener)
-        client = ssl.wrap_socket(
-            eventlet.connect(listener.getsockname()),
-            check_hostname=True,
-            verify_mode=ssl.CERT_REQUIRED,
-        )
+        ctx = ssl.SSLContext()
+        ctx.verify_mode = ssl.CERT_REQUIRED
+        ctx.check_hostname = True
+        ctx.load_verify_locations(tests.certificate_file)
+        client = ctx.wrap_socket(
+            socket.socket(socket.AF_INET, socket.SOCK_STREAM),
+            server_hostname="Test")
+        client.connect(listener.getsockname())
         client.send(b"check_hostname works")
         client.recv(64)
         server_coro.wait()

```
?

tipabu (changes requested):
I'm not seeing the test pop if I back out the change to eventlet/green/ssl.py -- and I'm kinda surprised the test doesn't pass a `server_hostname` arg anywhere :confused:

The diff I included got me a failing test for py36, but (surprisingly?) it passes on master for py37-39 -- not sure what's up with that.

temoto:
@tipabu thanks. It actually breaks on <3.7, the opposite of my version skipping.

Issue #638, opened 2020-08-13
WIP: Remove dependency on nose


mcepl:
I am a maintainer of Python packages in openSUSE, and I am on my crusade of eliminating nose1 from our distribution. When I look at its repository on https://github.com/nose-devs/nose, the last release 1.3.7 was on 2 Jun 2015, and even the last commit on the master branch was on 4 Mar 2016. Moreover, it seems that nose1 will stop working with upcoming Python 3.9 (or 3.10).

[This patch](https://github.com/eventlet/eventlet/files/5070080/remove_nose.patch.txt) eliminates dependency on nose, which has been unnecessary for a long time. Resulting test suite depends on pytest, which is currently perfectly developed and maintained.

Unfortunately, the resulting testsuite (with tarball 0.26.1) doesn’t pass for me: there are 61 on collection of the tests by pytest (which use as the test runner, without explicit dependency on it), see [this complete detailed build log](https://github.com/eventlet/eventlet/files/5070117/_log.txt) for more.

Do you have any idea what I do wrong? Is there a way how to make the test suite to pass even in the environment without the network connection (no Linux distro build machines have network access)? Would it be possible to [mark](https://docs.pytest.org/en/latest/mark.html) some tests as requiring network, so they can be *en mass* skipped?

Thank you for any help.

bnavigator:
Hi @mcepl, it's great that you work on the tests for this package. Just about now there arose problems because the package skipped all the tests even with nose. And so it was not detected that the update to dnspython 2.0.0 broke it.
https://build.opensuse.org/request/show/826300#comment-1285835

dnspython wants to read from /etc/resolv.conf by default which does not exist on offline OBS hosts.

My crude workaround I came up with just a few hours ago, is to provide a dummy package `python-eventlet-resolvdummy` but I guess that won't be accepted into Factory.
https://build.opensuse.org/package/show/home:bnavigator:branches:devel:languages:python/python-eventlet

At least before your nose removal patch got pulled in, I got the failures down to a few dozen. The biggest fail factor was the unclean teardown with `verify_hub_empty`

mcepl:
Is there bug for this on the dnspython project?

bnavigator:
- https://github.com/eventlet/eventlet/issues/629
- https://github.com/rthalley/dnspython/issues/559
- dnspython itself does not read resolv.conf in their own tests anymore https://github.com/rthalley/dnspython/pull/519
[...]

PR #707, opened 2021-05-11
Add classifiers for py38 and py39

tipabu:


tipabu:
FWIW, I started looking at Python 3.10 support based off the recent beta, but it looks like it'll take some effort. We'll want to address #638 -- otherwise nosetests will bomb out collecting tests with
```
  File ".../nose/suite.py", line 106, in _set_tests
    if isinstance(tests, collections.Callable) and not is_suite:
AttributeError: module 'collections' has no attribute 'Callable'
```
Even if we [move to pytest](https://github.com/tipabu/eventlet/commit/892ac1ead3a74356779d72b897fdf749fc4dcb2b) or something, though, we'll start seeing errors like
```
  File ".../eventlet/greenio/base.py", line 32, in <module>
    socket_timeout = eventlet.timeout.wrap_is_timeout(socket.timeout)
  File ".../eventlet/timeout.py", line 166, in wrap_is_timeout
    base.is_timeout = property(lambda _: True)
TypeError: cannot set 'is_timeout' attribute of immutable type 'TimeoutError'
```

jstasiak (approved):

PR #721, opened 2021-08-08
fix #697 add Lock class and make API compatible

### LINKED ISSUE #697: Monkey patch changes threading.Lock().release() behaviour
After monkey patching `threading.Lock().release()` returns `True` instead of raising an exception.

Behaviour before patch:

```sh
$ python
Python 3.8.5 (default, Jan 27 2021, 15:41:15) 
[GCC 9.3.0] on linux
>>> import eventlet
>>> import threading
>>> threading.Lock().release()
RuntimeError: release unlocked lock
```

After patch:

```sh
>>> eventlet.monkey_patch()
>>> threading.Lock().release()
True
```

Noticed this because it broke `readerwriterlock` as `rwlock.py` does this at import (which obviously fails because no exception is raised by `release` any more):

```python3
try:
	threading.Lock().release()
except BaseException as exc:
	RELEASE_ERR_CLS = type(exc)  # pylint: disable=invalid-name
	RELEASE_ERR_MSG = str(exc)
else:
	raise AssertionError()  # pragma: no cover
```

Installed:

```sh
$ pip freeze
dnspython==1.16.0
eventlet==0.30.2
greenlet==1.0.0
six==1.15.0
$
```

Work around for my immediate problem is obviously enough `from readerwriterlock import rwlock` before `import eventlet`.

mwri:
Fixes #697. Replace patched Lock implementation with new Lock class which extends the existing Semaphore class, overriding release to restore API compatible behaviour.


temoto (commented):
Please add this regression to `tests/thread_test.py`

```python
def test_compat_lock_release():
    # https://github.com/eventlet/eventlet/issues/697
    for mod in (patcher.original("threading"), thread):
        try:
            mod.Lock().release()
        except RuntimeError as e:
            assert "release unlocked lock" in str(e).lower(), str((mod, e))
```

mwri:
> Please add this regression to `tests/thread_test.py`
> 
> ```python
> def test_compat_lock_release():
>     # https://github.com/eventlet/eventlet/issues/697
>     for mod in (patcher.original("threading"), thread):
>         try:
>             mod.Lock().release()
>         except RuntimeError as e:
>             assert "release unlocked lock" in str(e).lower(), str((mod, e))
> ```

Very good yes, agree. I had to make some other changes for that, but it's all to the good; more compatible.

I did get a py27-epolls fail with this revision, but it seems to be transient, I've not had it locally and the CI didn't get it the second time around:

```
FAIL: tests.patcher_test.test_fork_after_monkey_patch
  File "/home/runner/work/eventlet/eventlet/.tox/py27-epolls/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/runner/work/eventlet/eventlet/tests/patcher_test.py", line 521, in test_fork_after_monkey_patch
    tests.run_isolated('patcher_fork_after_monkey_patch.py')
  File "/home/runner/work/eventlet/eventlet/tests/__init__.py", line 381, in run_isolated
    run_python(prefix + path, **kwargs)
  File "/home/runner/work/eventlet/eventlet/tests/__init__.py", line 374, in run_python
    assert ok, 'Expected single line "pass" in stdout'
AssertionError: Expected single line "pass" in stdout
```

temoto:
Seems that LockType -> Lock rename could break some code. https://github.com/search?l=Python&q=thread%5C.LockType&type=Code

mwri:
> Seems that LockType -> Lock rename could break some code. https://github.com/search?l=Python&q=thread%5C.LockType&type=Code

Ok, I guess that was naive of me. It's not part of the API right? Maybe it should have been `_LockType`. How about exporting both then, like so...

eventlet/green/thread.py
```diff
@@ -3,13 +3,14 @@
 import six
 from eventlet.support import greenlets as greenlet
 from eventlet import greenthread
-from eventlet.semaphore import Semaphore as LockType
+from eventlet.lock import Lock as LockType
+from eventlet.lock import Lock as Lock
```
temoto, line 7:
```suggestion
from eventlet.lock import Lock
```

eventlet/green/thread.py
```diff
@@ -3,13 +3,14 @@
...
 import sys
 
 
 __patched__ = ['get_ident', 'start_new_thread', 'start_new', 'allocate_lock',
                'allocate', 'exit', 'interrupt_main', 'stack_size', '_local',
-               'LockType', '_count']
+               'LockType', 'Lock', '_count']
 
 error = __thread.error
```
temoto, line 15:
```suggestion
LockType = Lock
error = __thread.error
```

temoto (commented):
thanks for fixing test for 2.7

temoto:
> Ok, I guess that was naive of me. It's not part of the API right? Maybe it should have been `_LockType`. How about exporting both then, like so...

Yeah it's not crystal clear. Recently we had situation where we broke something that wasn't considered public API, but important external project relied on.

temoto (approved):
LGTM

temoto:
@tipabu @jstasiak @edwardgeorge please have a look

mwri:
Rebased.

tipabu (approved):
Sorry it took so long; seems reasonable. I wonder a bit if we ought to move `_at_fork_reinit` over to _just_ be on the new `Lock` class, since that's the only context in which we need it. See also: https://github.com/eventlet/eventlet/commit/087ba743c7af8a40ac1e4e2ec89409eee3b4233e

temoto:
@tipabu yes, agree, please observe updated patch.

tipabu:
I see that the test moved, but I was expecting the `_at_fork_reinit` method to move, too, like https://github.com/tipabu/eventlet/commit/22e72b085fb9924cccd7661bb6d5d17faea3f354

It also simplifies the implementation a bit, since we don't need to track the original value -- a semaphore used in place of `threading.Lock` should always start with `counter=1`. That said, it's a nit; I'd be content to have this merge as-is.

FWIW, I don't think we need to worry about breaking the Semaphore API by removing this method, since the only caller should be buried down in stdlib and expecting a Lock-like, not a Semaphore in particular.

PR #684, opened 2021-02-04
greendns: Patch ssl

tipabu:
Newer (2.0+) `dnspython` [imports `ssl`](https://github.com/rthalley/dnspython/blob/v2.0.0/dns/query.py#L48) (and [`requests`](https://github.com/rthalley/dnspython/blob/v2.0.0/dns/query.py#L40), which would have eventually imported `ssl`), so `greendns` needs a monkey-patched `ssl`, too. This should prevent a `RecursionError` with `SSLContext`.

Note that this *does not* bring full `dnspython>=2.0` support, but our version pin hasn't stopped people from trying to run with newer `dnspython` anyway. May as well make it hurt a little less.

Fixes #677, but see also #619.

Issue #632, opened 2020-07-20
python 3.6 TypeError: _wrap_socket() argument 1 must be _socket.socket, not GreenSSLSocket


SeanMooney:
recently in the openstack ci we started to see the following error
TypeError: _wrap_socket() argument 1 must be _socket.socket, not GreenSSLSocket
https://zuul.opendev.org/t/openstack/build/1180d2313ade4d5480550313bcd20b79/log/controller/logs/screen-n-novnc-cell1.txt#29

this seams similar to https://github.com/eventlet/eventlet/issues/201
there has not been a new eventlet release or websockify release in several months so its not clear why this would start failing now. we have not actully debuged this much but i was wondering if this seam familar or is a known issue?

what is alittle bit confusing is we have a passing build from 4 days ago https://zuul.opendev.org/t/openstack/build/1d23dbdf6fa84702ab543602cae1ba67 which seams to be useing more or less the same versions
as the failing build https://zuul.opendev.org/t/openstack/build/1180d2313ade4d5480550313bcd20b79

the eventlet version used was 0.25.2
https://zuul.opendev.org/t/openstack/build/1180d2313ade4d5480550313bcd20b79/log/controller/logs/pip3-freeze.txt#31

we were also using openssl version  1.1.1-1ubuntu2.1~18.04.6 in both builds.

comparing the python package difference between the two runs there are some but nothing that looks related
http://paste.openstack.org/show/796119/

and it looks like we were using the same python build  python 3.6.7-1~18.04 in both cases
https://zuul.opendev.org/t/openstack/build/1180d2313ade4d5480550313bcd20b79/log/controller/logs/devstacklog.txt#656
https://zuul.opendev.org/t/openstack/build/1d23dbdf6fa84702ab543602cae1ba67/log/controller/logs/devstacklog.txt#656

so at first glance its not clear why one would pass and the the other would fail which is why i tought i would ask the experts.

----------------error traceback -------------------

Jul 20 10:44:42.425350 ubuntu-bionic-rax-ord-0018472470 nova-novncproxy[22231]: INFO nova.console.websocketproxy [None req-4c318998-3a74-4929-beab-b774a72844cf None None] handler exception: _wrap_socket() argument 1 must be _socket.socket, not GreenSSLSocket
[...]

Issue #619, opened 2020-06-25
eventlet is incompatible with dnspython 2.0.0rc1


frenzymadness:
Hello.

dnspython version 2.0.0 will be released soon (it's a release candidate now, https://github.com/rthalley/dnspython/issues/466) and it seems that eventlet is not compatible with it.

```
# pip list
Package    Version
---------- --------
dnspython  2.0.0rc1
eventlet   0.25.2
greenlet   0.4.16
monotonic  1.5
pip        20.1.1
setuptools 47.3.1
six        1.15.0
wheel      0.34.2

# python
Python 3.8.3 (default, May 29 2020, 00:00:00) 
[GCC 10.1.1 20200507 (Red Hat 10.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import eventlet
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/__init__.py", line 10, in <module>
    from eventlet import convenience
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/convenience.py", line 7, in <module>
    from eventlet.green import socket
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/green/socket.py", line 21, in <module>
    from eventlet.support import greendns
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/support/greendns.py", line 64, in <module>
    setattr(dns, pkg, import_patched('dns.' + pkg))
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/support/greendns.py", line 59, in import_patched
    return patcher.import_patched(module_name, **modules)
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/patcher.py", line 123, in import_patched
    return inject(
  File "/home/lbalhar/.virtualenvs/tmp-722a19f0bf97ea6/lib/python3.8/site-packages/eventlet/patcher.py", line 100, in inject
    module = __import__(module_name, {}, {}, module_name.split('.')[:-1])
[...]

PR #722, opened 2021-08-16
greendns: Use _compute_times on dnspython 2 (Updated)

felixonmars:
This PR is based on https://github.com/eventlet/eventlet/pull/639

Rebased against latest master, and updated the GitHub CI instead of Travis.

I have also updated the tests for the last test failure left.


eventlet/support/greendns.py
```diff
@@ -120,6 +120,15 @@ def is_ip_addr(host):
     return is_ipv4_addr(host) or is_ipv6_addr(host)
 
 
+def compute_expiration(query, timeout):
+    # NOTE(ralonsoh): in dnspython v2.0.0, "_compute_expiration" was replaced
+    # by "_compute_times".
+    if hasattr(query, '_compute_expiration'):
```
temoto, line 126: Think you can do this attribute check at module level to avoid runtime cost?

tox.ini
```diff
@@ -16,7 +16,7 @@ statistics = 1
 [tox]
 minversion=2.5
 envlist =
-    ipv6, pep8, py{27,35,36,37,38,39,py2,py3}-{selects,poll,epolls}, py38-openssl
+    ipv6, pep8, py{27,35,36,37,38,39,py2,py3}-{selects,poll,epolls,dnspython1}, py38-openssl
```
temoto, line 19:
```suggestion
    dnspython1, ipv6, pep8, py{27,35,36,37,38,39,py2,py3}-{selects,poll,epolls}, py38-openssl
```

IMHO, it's sufficient to check against dnspython1 on any single Python version.

.github/workflows/test.yaml
```diff
@@ -32,22 +32,28 @@ jobs:
...
+          - { py: 3.7, toxenv: py37-dnspython1, ignore-error: false }
           - { py: 3.8, toxenv: py38-epolls, ignore-error: false }
           - { py: 3.8, toxenv: py38-openssl, ignore-error: false }
           - { py: 3.8, toxenv: py38-poll, ignore-error: false }
           - { py: 3.8, toxenv: py38-selects, ignore-error: false }
+          - { py: 3.8, toxenv: py38-dnspython1, ignore-error: false }
           - { py: 3.9, toxenv: py39-epolls, ignore-error: false }
           - { py: 3.9, toxenv: py39-poll, ignore-error: false }
           - { py: 3.9, toxenv: py39-selects, ignore-error: false }
+          - { py: 3.9, toxenv: py39-dnspython1, ignore-error: false }
```
temoto, line 56:
```suggestion
          - { py: 3.x, toxenv: dnspython1, ignore-error: false }
```
+ remove combination explosion above

IMHO, it's sufficient to check against dnspython1 on any single Python version.

felixonmars: Should I literally write `3.x` here? (Sorry I'm not so familiar)

Otherwise I have updated to only test on Python 2.7 & 3.9.

temoto: Yeah 27 is fine even if redundant. Also you can press "commit suggestion" button if you agree with proposed code.

temoto: 3.x selects "latest of 3.any"

felixonmars: I see. I thought 3.x was invalid so I didn't click the button. Sorry :P

eventlet/support/greendns.py
felixonmars: Updated too :)

felixonmars:
Okay, found the cause. They are checking whether the QR flag is set now.

All green now :)

temoto (approved):
LGTM, **but** I'm pretty sure someone reported that compute_time wasn't enough for compatibility with dnspython2.

So to actually merge this, we need to run some real world application using updated code. Testing volunteers are very welcome.

felixonmars:
> LGTM, **but** I'm pretty sure someone reported that compute_time wasn't enough for compatibility with dnspython2.
> 
> So to actually merge this, we need to run some real world application using updated code. Testing volunteers are very welcome.

I took a look at the other related issue: #632

The original logs are not accessible now, but based on on nova's bug report this seems to be caused by a module import order issue: https://bugs.launchpad.net/nova/+bug/1888237

In #684 we are patching ssl now, not sure if it helps on this. But IMHO if it's indeed caused by a module import order issue, it should be fixed at the corresponding project (by running eventlet monkeypatch in prior to importing dnspython), and we can do very little here.

temoto:
Maintainer reminder before merge: add co-authored-by @ralonsoh and @jayvdb

rthalley:
See my comments in #619 [here](https://github.com/eventlet/eventlet/issues/619#issuecomment-660250478).  There were more changes than to just compute_times, as the API of dns.query.udp() and dns.query.tcp() changed too, and eventlet monkeypatching them back to old versions will cause failures with the resolver.

felixonmars:
@rthalley Thank you, and sorry for missing that comment before. I have added all the missing parameters for udp() and tcp(), up to master. Hopefully the implementation makes sense.

filippog:
I'm using python-eventlet from Debian Bullseye (`0.26.1-7`) and can confirm that with these changes DNS resolution works again with dnspython `2.0.0` and Openstack Swift `2.26.0-10`

PR #724, opened 2021-09-11
"/etc/resolv.conf" assumed default path is wrong for termux

rawrrawrfox:
termux: the default path for resolv.conf file is not "/etc/resolv.conf", but the code presumes that to be the case

termux is a terminal app for android capable of running many linux utilites

the file resolv.conf is not found at the default location which therefore breaks other projects which use this one as a dependency

i ran accross this problem when trying to run the termux version of onionshare (https://github.com/oleanderxoxo/onidroid) which uses this library as a dependency somewhere downstream

i'm not sure what would be the most elegant solution to the resolv.conf location problem, but i've tested this code, and it works as a quickfix nonetheless

it checks if the $PREFIX environment variable is set and if it points to the termux app's "...data/files/usr" system path, then specifies explicitly the location of the resolv.conf file when creating the resolver object, which needs to be $PREFIX + "/etc/resolv.conf"

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -408,7 +408,24 @@ class ResolverProxy(object):
         return aliases
 
 
-resolver = ResolverProxy(hosts_resolver=HostsResolver())
+"""
+Some systems' default resolv.conf file is not located in /etc/resolv.conf
+
+for instance, in termux (terminal app in android) there are system configurations wich can not
+be mounted or symbolically linked to the android's /etc/ path, due to how android works
+and therefore have their own locations in a subdirectory in the app's data directory
+
+the termux environment can be identified via the $PREFIX environment variable, and so
+the following code provides a patch for this specific case, while maintaining the default 
+functionality if the system is not termux
+"""
+resolver = None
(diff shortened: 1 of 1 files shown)


temoto:
@rawrrawrfox thank you. Configurable location of resolv.conf is a great feature.

I think we should try to support one of:
- `EVENTLET_RESOLV_CONF=/path/to/resolv.conf` or just `RESOLV_CONF` at first, becase second option plain `$PREFIX/etc/resolv.conf` may break unexpectedly in a lot of build environments.
- `$PREFIX/etc/resolv.conf` seems most reasonable after specific path option for resolv.conf file.

Unless required, I'm strongly against hardcode integration with one particular platform+application (Android/termux). It should work for other applications that set PREFIX like termux but to another value. IMHO, it would be awkward to limit this useful feature to Android.

@eventlet/developers @jstasiak @tipabu @clayg @edwardgeorge @nat-goodspeed @vikt0rs  please say

clayg:
@rawrrawrfox I agree with @temoto strong preference for `EVENTLET_RESOLV_CONF=/path/to/resolv.conf` - I found no such facility in [dns.resolver.Resolver](https://github.com/rthalley/dnspython/blob/master/dns/resolver.py#L743)

```

PR #743, opened 2022-01-07
Prevent deadlock on logging._lock

### LINKED ISSUE #742: Monkey-patched RLocks can deadlock
I've been trying to run down some CI timeouts I've seen in a project that uses eventlet. I finally managed to catch it in the act and extracted some stack frames that looked like
```
...
  File "/home/zuul/src/opendev.org/openstack/swift/swift/common/utils.py", line 2402, in get_logger
    logger = logging.getLogger(log_route)
  File "/usr/lib/python3.9/logging/__init__.py", line 2042, in getLogger
    return Logger.manager.getLogger(name)
  File "/usr/lib/python3.9/logging/__init__.py", line 1315, in getLogger
    _releaseLock()
  File "/usr/lib/python3.9/logging/__init__.py", line 232, in _releaseLock
    _lock.release()
  File "/usr/lib/python3.9/threading.py", line 189, in release
    self._block.release()
  File "/home/zuul/src/opendev.org/openstack/swift/.tox/py39/lib/python3.9/site-packages/eventlet/lock.py", line 24, in release
    return super(Lock, self).release(blocking=blocking)
  File "/usr/lib/python3.9/logging/__init__.py", line 831, in _removeHandlerRef
    acquire()
  File "/usr/lib/python3.9/logging/__init__.py", line 225, in _acquireLock
    _lock.acquire()
  ...
```
(Full stacks are available at https://paste.opendev.org/show/811957/, but I'm fairly certain this is the only relevant bit.) The trouble seems to be that GC snuck in while we were in the process of releasing `logging._lock`, dropped the refcount for some handler to zero, and triggered a weakref callback which also wants to take the lock. Since the `_PyRLock` already cleared its ownership information, it can't re-enter and deadlocks.

It sounds a bit like a CPython bug; maybe `_PyRLock` has always had this possibility to deadlock if weakref callbacks also want to take the lock. Indeed, if I use `sys.settrace` I can reproduce it without eventlet _at all_, but that feels a bit like cheating: https://paste.opendev.org/show/811970/

However, since upgrading eventlet 0.32.0 to 0.33.0, there's been a noticeable up-tick in CI timeouts: what used to happen maybe once or twice a week now happens daily, often multiple times. I strongly suspect that the extra stack frame introduced in #721 has increased the likelihood of having GC run during that key time between clearing ownership information in the `_PyRLock` and releasing the underlying `Lock`. Maybe just duplicating the `Semaphore.release` code into `Lock.release` (instead of using `super`) would be enough to significantly reduce the likelihood of deadlocks?

FWIW, I'm fairly certain this is only an issue for tests (and specifically, tests which create and discard a bunch of loggers), not production code (where presumably logger handlers are rarely removed). I've also only noticed the issue on py3 (specifically, we test against py36-py39), not on py27 for whatever reason.

tipabu:
Or at least, reduce the likelihood of it.

We've seen deadlocks when GC runs while in the process of releasing an `RLock`; it results in a stack that looks like

```
  File ".../logging/__init__.py", line 232, in _releaseLock
    _lock.release()
  File ".../threading.py", line 189, in release
    self._block.release()
  File ".../eventlet/lock.py", line 24, in release
    return super(Lock, self).release(blocking=blocking)
  File ".../logging/__init__.py", line 831, in _removeHandlerRef
    acquire()
  File ".../logging/__init__.py", line 225, in _acquireLock
    _lock.acquire()
```

That is, we try to release the lock, but in between clearing the `RLock` ownership information and releasing the underlying `Lock`, GC runs and invokes a weakref callback, which in turn tries to acquire the `RLock`. Since the ownership information has already been cleared, the lock's no longer re-entrant and everything seizes.

This seems to have become more of a problem since we separated `Lock` and `Semaphore`; apparently the extra stack frame makes it much more likely that GC can sneak in during that critical moment. So, inline the release inside of `Lock` rather than punting to `Semaphore`; the implementation is simple anyway, and hasn't changed for at least 12 years (since `Semaphore` was pulled out to its own module).

Closes #742

### DIFF
--- a/eventlet/lock.py
+++ b/eventlet/lock.py
@@ -1,3 +1,4 @@
+from eventlet import hubs
 from eventlet.semaphore import Semaphore
 
 
@@ -21,7 +22,15 @@ class Lock(Semaphore):
         if self.counter > 0:
             raise RuntimeError("release unlocked lock")
 
-        return super(Lock, self).release(blocking=blocking)
+        # Consciously *do not* call super().release(), but instead inline
+        # Semaphore.release() here. We've seen issues with logging._lock
+        # deadlocking because garbage collection happened to run mid-release
+        # and eliminating the extra stack frame should help prevent that.
+        # See https://github.com/eventlet/eventlet/issues/742
+        self.counter += 1
+        if self._waiters:
+            hubs.get_hub().schedule_call_global(0, self._do_acquire)
+        return True
 
     def _at_fork_reinit(self):
         self.counter = 1


temoto:
@tipabu hi, did this work well?

tipabu:
So I hacked up my CI jobs in https://review.opendev.org/c/openstack/swift/+/826950/ -- with 10+ py3 jobs per run that all focused on tests that create and destroy a lot of loggers, I'd see 3-5 deadlocks per run (see the three builds for patchset 3).

Switching to installing from this branch, all the py3 jobs passed with no deadlocks (patchset 4).

temoto (approved):

tipabu:
Any chance we could get a release that includes this? My CI has a strong preference for tagged releases, so I'm still seeing some deadlocks. Or would it be preferable for me to put together the changelog?

temoto:
@tipabu yeah 0.33.1 is on PyPI

PR #747, opened 2022-01-13
wsgi: server MUST NOT send Content-Length/Transfer-Encoding header in response with a status code of 1xx, 204 or (2xx to CONNECT request)

### LINKED ISSUE #746: wsgi response header handling disregards RFC7230 with regard to an HTTP 204 response code
[RFC7230 specifies that transfer encoding](https://httpwg.org/specs/rfc7230.html#header.transfer-encoding) must not include a Transfer-Encoding header in an HTTP 204 response. However, eventlet appears to always do so by default based upon the class of request, and if the response does not have a content-length field explicitly defined. Then regardless of the response code, it is stamping the reply with a Transfer-Encoding header.

> A server MUST NOT send a Transfer-Encoding header field in any response with a status code of 1xx (Informational) or 204 (No Content). A server MUST NOT send a Transfer-Encoding header field in any 2xx (Successful) response to a CONNECT request (Section 4.3.6 of [RFC7231]).

Specifically, as noted, it is basing it's response on the presence of a [Content-Length header](https://httpwg.org/specs/rfc7230.html#header.content-length). However, HTTP 204 responses must not have a Content-Length header based upon the RFC.

> A server MUST NOT send a Content-Length header field in any response with a status code of 1xx (Informational) or 204 (No Content). A server MUST NOT send a Content-Length header field in any 2xx (Successful) response to a CONNECT request (Section 4.3.6 of [RFC7231]).

This creates a situation where, as previously mentioned the Transfer-Encoding header is injected, and a HTTP/1.1 client can find itself confused, and actually issue an error or exception as a result of the response because a client interpreting the Transfer-Encoding before the result code is expecting more data. That data, however, will never come and depending on the application, the socket may close as well, ultimately generating additional errors.

@dtantsur was kind enough to whip up a reproducer utilizing Flask and eventlet wsgi:

>import eventlet
>import eventlet.wsgi
>import flask
>
>
>app = flask.Flask(__name__)
>
>
>@app.route('/', methods=['PUT', 'DELETE'])
>def test():
>    return flask.make_response('', 204)
>
>
> if __name__ == '__main__':
>    sock = eventlet.listen(('127.0.0.1', 5000))
>     eventlet.wsgi.server(sock, app, debug=True)

Test and result:

> $ curl -vX DELETE http://localhost:5000/
> *   Trying ::1:5000...
> * connect to ::1 port 5000 failed: Connection refused
> *   Trying 127.0.0.1:5000...
> * Connected to localhost (127.0.0.1) port 5000 (#0)
> > DELETE / HTTP/1.1
> > Host: localhost:5000
> > User-Agent: curl/7.74.0
> > Accept: */*
> > 
> * Mark bundle as not supporting multiuse
> < HTTP/1.1 204 NO CONTENT
> < Content-Type: text/html; charset=utf-8
> < Date: Thu, 13 Jan 2022 18:33:20 GMT
> < Transfer-Encoding: chunked
> < 
> * Excess found: excess = 5 url = / (zero-length body)
> * Connection #0 to host localhost left intact
> $

Curl doesn't return a non-zero return code in this precise reproducer case, but you can see that the header is injected inappropriately when compared to RFC7230. This has varying effects on different client implementations. In one case we encountered this was aiding in Apache httpd returning a incomplete response error to the end client. In another we found python-requests trying to parse the response body and an exception being raised as a result there.

This was observed on Eventlet 0.30.2 and comparing the code to the present development branch indicates it is still present.
Both encountered on Python 3.8.5 and Python 3.6.8. Debian and Red Hat Enterprise Linux, respectively.

Ideally, the injection of headers would be skipped if the HTTP result code is 204.

### LINKED ISSUE #862: wsgi response with status code 304 contains a body
`eventlet.wsgi.server` adds an empty chunked body to 304 responses. This violates [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-304-not-modified) which states that:

> A 304 response is terminated by the end of the header section; **it cannot contain content** or trailers.

292266831-080441bb-e88a-437a-b4f8-f2f93662be20

This causes errors in reverse proxies like caddy or traefik, which tend to strictly adhere to the HTTP spec:

```
Dez 21 11:42:59 proliant caddy[48932]: {"level":"info","ts":1703155379.7331445,"msg":"Unsolicited response received on idle HTTP channel starting with \"0\\r\\n\\r\\n\"; err=<nil>"}
Dez 21 11:43:04 proliant caddy[48932]: {"level":"info","ts":1703155384.3559086,"msg":"Unsolicited response received on idle HTTP channel starting with \"0\\r\\n\\r\\n\"; err=<nil>"}
Dez 21 11:43:04 proliant caddy[48932]: {"level":"info","ts":1703155384.5456474,"msg":"Unsolicited response received on idle HTTP channel starting with \"0\\r\\n\\r\\n\"; err=<nil>"}
Dez 21 11:43:04 proliant caddy[48932]: {"level":"info","ts":1703155384.5492866,"msg":"Unsolicited response received on idle HTTP channel starting with \"0\\r\\n\\r\\n\"; err=<nil>"}
```

The content describes an empty chunk as outlined in https://en.wikipedia.org/wiki/Chunked_transfer_encoding#Format

```
0\r\n\r\n
```

A possible solution might be to drop the `Transfer-Encoding` header for 304 responses.

Minimal reproducible example: https://github.com/dgtlmoon/flask-304-modified-extra-in-reply-test

see https://github.com/dgtlmoon/changedetection.io/issues/2053 for more details

temoto:
https://httpwg.org/specs/rfc7230.html#rfc.section.3.3.1

Fixes #746
Fixes #862


eventlet/wsgi.py
```diff
@@ -479,6 +479,9 @@ def handle_one_response(self):
         use_chunked = [False]
         length = [0]
         status_code = [200]
+        # Status code of 1xx or 204 or 2xx to CONNECT request MUST NOT send body and related headers
+        # https://httpwg.org/specs/rfc7230.html#rfc.section.3.3.1
+        bodyless = [False]
```
tipabu, line 484: It bugs me a little that we're not actually _enforcing_ the absence of a body -- if application code responds `204 No Content` but then also returns an app iter that has some bytes, we're going to yield them out and mess up message framing. Bit of a slippery slope to go down, though, I suppose.

eventlet/wsgi.py
```diff
@@ -510,7 +513,7 @@ def write(data):
                     self.close_connection = 1
 
                 if 'content-length' not in header_list:
-                    if self.request_version == 'HTTP/1.1':
+                    if not bodyless[0] and self.request_version == 'HTTP/1.1':
```
tipabu, line 516: I think this will interfere with pipelining requests/responses. We may want something more like
```
if bodyless[0]:
    pass  # client doesn't expect a body anyway
elif self.request_version == 'HTTP/1.1':
    use_chunked[0] = True
    towrite.append(b'Transfer-Encoding: chunked\r\n')
elif 'content-length' not in header_list:
    self.close_connection = 1
```
(Side note: can that last block become an `else:`? Didn't we just check that `'content-length'` is not in `header_list`?)

tipabu (commented):
Makes sense, though I think there are two more cases from [section 3.3.3](https://httpwg.org/specs/rfc7230.html#rfc.section.3.3.3) where we may want to skip our standard attempt at imposing a content-length or transfer-encoding on the response:

1. responses to `HEAD` requests
2. responses with status `304 Not Modified`

While the RFC allows servers to include these headers in such responses, I don't think _we're_ in a position to know what appropriate values are -- only application code can say for sure.

eventlet/wsgi.py
temoto: Yes, agreed, thank you. But also we have a funny hack to allow random header case for compatibility with some popular server.

IMHO, we should claim WSGI and HTTP are separated at eventlet.wsgi, it takes care of transfer headers, may drop body and proper application expects that anyway. But also allow things like 204 with body via feature flag.

But it seems a fair amount of change. Suggest to extract this as separate task and implement later.

tipabu: Agreed; it can certainly be a follow-up -- or we just tell application developers to be more careful. Definitely not a blocker to getting this fix merged.

JojoBr0:
@4383 would it be possible to merge this request any time soon, it effects [this](https://github.com/dgtlmoon/changedetection.io/issues/2053) issue?

4383:
@JojoBr0 sure, let me check with Tim if this PR is ready to go.
@tipabu: Can we merged this pull request? I saw you updated it several time, so I prefer to check with you first.

tipabu:
Yep, this still looks good & makes sense. There's probably more that *could* be done, but this should fix up some known cases.

Pushed a merge from master and resolved the conflict in tests.

tests/wsgi_test.py
```diff
@@ -1994,6 +1996,48 @@ def test_close_idle_connections(self):
...
+        # Connection header, so we can keep going
+        sock.sendall(b"DELETE /foo HTTP/1.1\r\n\r\n")
+        response = read_http(sock)
+        assert "transfer-encoding" not in response.headers_lower
+        assert "content-length" not in response.headers_lower
+        assert "connection" not in response.headers_lower
+
+        sock.close()
+
+
```
4383, line 2040: extra blank line. it broke job

4383 (changes requested):

tests/wsgi_test.py
4383: I fixed it on my side

4383 (approved):

PR #752, opened 2022-02-24
ci: remove redundant basepython decls from tox.ini

mgorny:
Remove basepython declarations from tox.ini where they match what tox
already infers from the environment name.  This is NFC for existing
testenvs, and it makes it possible to run tests on newer Python targets,
e.g. py310-*.


temoto (approved):
LGTM

Issue #546, opened 2019-02-14
monkey patching of existing RLock is broken on CPython3


antocuni:
`eventlet.patcher._green_existing_locks` tries to monkey-patch locks which have been created before the call to monkey patch. To do so, it relies on `gc.get_objects()` to find them.

This works well on CPython2.7, but it is broken on CPython3.x, as the following example demonstrate (it prints `Found!` on 2.7 and `Not found` on 3.6):
```python
import threading
import gc
lock = threading.RLock()
for obj in gc.get_objects():
    if obj is lock:
        print('Found!')
        break
else:
    print('Not found')
```

This happens because in 3.6 `threading.RLock()` returns an object of type `_thread._RLock`, which is implemented in C, inside [_threadmodule.c](https://github.com/python/cpython/blob/master/Modules/_threadmodule.c).
In particular, [RLockType](https://github.com/python/cpython/blob/master/Modules/_threadmodule.c#L487) does not have the flag `Py_TPFLAGS_HAVE_GC`, and as such it is never returned by `gc.get_objects()`.

Another more direct demonstration of the problem:
```python
import threading
lock = threading.RLock()
print('my lock [1]:', type(lock), lock, id(lock))

import eventlet
eventlet.monkey_patch()
print('my lock [2]:', type(lock), lock, id(lock))
```
On my machine, with the latest eventlet HEAD (commit a915bb6), running it on `python3.6` shows that `id(lock)` is still the same even after the `monkey_patch()`.

The fact that it went unnoticed until now probably means that this specific functionality lacks a test.

However, I don't know how to fix the problem: it seems to me that there is not reasonable way to implement this functionality in Python3 (not counting the fact that even if we find the existing locks, monkey-patching them is hard: `_fix_py3_lock` looks *very* fragile and there are probably cases in which it doesn't work at all).

A possible, brutal "solution" would be to fail and complain loudly if you try to call `monkey_patch` after `threading` has already been imported. It probably breaks a lot of existing code, but it's still better than leaving it half-broken as it is now

antocuni:
To explain better what are the concrete consequences of this, here is an example which shows the wrong behavior in python3.6:

```python
from __future__ import print_function
import threading
lock = threading.RLock()

import eventlet
eventlet.sleep()
eventlet.monkey_patch()

def a():
    with lock:
        print('A1')
        eventlet.sleep()
    print('A2')

def b():
    with lock:
        print('B1')
        eventlet.sleep()
    print('B2')

def main():
[...]

PR #754, opened 2022-04-20
py<=3.10: Green logging locks

### LINKED ISSUE #730: python 3.10 issue with RLock patcher existing locks
As reported at https://github.com/eventlet/eventlet/pull/715#pullrequestreview-733405534

> Bad news: somewhere between 3.10.0b1 and 3.10.0rc1, `test_patcher_existing_locks_locked` started failing with `RuntimeError: cannot release un-acquired lock` around
> 
> https://github.com/eventlet/eventlet/blob/c6c350eaa9eb819c6bcabe25113464aed75b9cf5/tests/isolated/patcher_existing_locks_locked.py#L19
> 
> I'm still scratching my head over that one. It looks like `gc.get_referrers(old)` is mostly returning lists and dicts, though? I wonder if it's somehow related to the trouble I was seeing with `__builtins__` showing up as a dict in socket_test.test_error_is_timeout...

I decided to bisect the CPython for the commit which broke this test. I have found this commit https://github.com/python/cpython/commit/e30fe27dabbc6b48736c3c17d57f6fa542376e8f . When running the test suite, it passes on its parent commit, but fails on this commit (the same way it fails for released 3.10.0). Just as information, it was between 3.10.0_beta2 and 3.10.0_beta3.

---

Now, about how to fix it? Sadly I'm quite new to eventlet and I don't understand quite well the code here:
https://github.com/eventlet/eventlet/blob/8904a339f2df3bd7958e73142d640cd3d0a9245c/eventlet/patcher.py#L392-L403
But when I tested the types on multiple targets (py3.8, py3.9, py3.10, pypy7.3.5, pypy7.3.6) it always returned `_thread.RLock` for `type(threading.RLock())`.
So just as a test, I tried to manually set `pyrlock_type = rlock_type` at line 398, and the specific test passed. Here is the diff.
```diff
diff --git a/eventlet/patcher.py b/eventlet/patcher.py
index b249d6f..3f828a7 100644
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -395,7 +395,7 @@ def _green_existing_locks():
         # this happens on CPython3 and PyPy >= 7.0.0: "py3-style" rlocks, they
         # are implemented natively in C and RPython respectively
         py3_style = True
-        pyrlock_type = type(threading._PyRLock())
+        pyrlock_type = rlock_type
     else:
         # this happens on CPython2.7 and PyPy < 7.0.0: "py2-style" rlocks,
         # they are implemented in pure-python

```

---

I will repeat that I have no idea if my change is even mildly correct. I just know it managed to make this test, and added no new regressions when all other tests were run. That is also the reason this isn't an Pull Request, and just an issue report.

My test environment is Gentoo, using `master` version of  eventlet. Tested on CPython 3.8.12 and 3.9.7 (passes with and without the diff) and CPython 3.10.0 (needed the diff, but there were still other failures in other parts, even without the diff), and PyPy3-7.3.5 and PyPy3-7.3.6 (passes with and without the diff).

tipabu:
`gc.get_referrers()` returns nothing but `list`s and `dict`s, so previously we'd always trip a `TypeError` and skip patching.

Closes #730

This also seems to allow `_fix_py3_lock` to actually have an effect on py34-py39, though **finding** the locks that need fixing remains a challenge.


eventlet/patcher.py
```diff
@@ -412,6 +412,13 @@ def _green_existing_locks():
             elif py3_style and not isinstance(obj, pyrlock_type):
                 _fix_py3_rlock(obj)
 
+    if (3, 0) <= sys.version_info <= (3, 10):
+        # Older py3 won't have RLocks show up in gc.get_objects() -- see
+        # https://github.com/eventlet/eventlet/issues/546 -- so green a handful
+        # that we know are significant
+        import logging
+        _fix_py3_rlock(logging._lock)
```
tipabu, line 420: Maybe should add in something like
```
    logging._acquireLock()
    try:
        for ref in logging._handlerList:
            handler = ref()
            if handler:
                _fix_py3_rlock(handler.lock)
            del handler
    finally:
        logging._releaseLock()
```
?

temoto:
@tipabu thanks a lot for fix. I'm going to pull remote verification card here. Did you run anything close to real project with this patch?

tipabu:
> Did you run anything close to real project with this patch?

Not yet, unfortunately. My prod environment is still in the process of transitioning from py27 -> py37, so any significant testing with py310 is still off in the nebulous future. I could probably get it running in my homelab, but haven't had time and I'm not entirely sure how I'd want to validate it.

For a little more context, we're [trying to remove](https://review.opendev.org/c/openstack/swift/+/457110) some `monkey_patch()`-as-import-side-effect so we can be more intentioned about when we're monkey-patched and when we're not. As part of that, we got a little skittish about the impact of delaying monkey-patching -- #546 specifically was identified as a concern, in part because we [currently explicitly green `logging._lock`](https://github.com/openstack/swift/blob/2.29.1/swift/common/utils.py#L103), and it's unclear what impact there may be in taking that out.

thomasgoirand:
Hi,

FYI, this fixed the unit test failure for me with Python 3.10 in Debian unstable. Thanks for the patch!

Cheers,

Thomas Goirand (zigo)

tipabu:
Looks like a testtools bug -- proposed https://github.com/testing-cabal/testtools/pull/336 to address. In the meantime, this patch should be fine to use `is` instead of `==`.

temoto:
@thomasgoirand could you confirm that new version using `is` comparison works?

4383:
@thomasgoirand: o/ please can you confirm that new version using `is` comparison works?

See previous comments

4383 (approved):
LGTM

4383:
@itamarst: Feel free to merge this patch if you think that it is now ok, else do not hesitate to request an other review.

Thanks

Issue #757, opened 2022-06-17
when use eventlet.green.fdopen(）to write a file with 'a' mode raise "io.UnsupportedOperation: File or stream is not writable." error


wuming0:
if you run some code like below in python 3,you will find it will occur a error,"OSError: File or stream is not writable." :
```python
# !/usr/bin/env python
# -*- coding:utf-8 -*-
# create by Zhichao Wu at 2022-06-17

# import os
from os import  O_CREAT, O_APPEND, O_RDWR

from eventlet.green import os


def write_file():
    filepath = './1.txt'
    fd = os.open(filepath, O_RDWR | O_CREAT | O_APPEND, 0o640)
    with os.fdopen(fd, 'a') as fw:
        fw.write('eventlet.greenio.py3.GreenFileIO.writable() bug!')

    print(f'---{filepath} result----')
    with open(filepath, 'r') as fr:
        print(fr.read())


if __name__ == "__main__":
    write_file()
````

if you run in python 3,you will got this error:
![github_evenlet_issue_2](https://user-images.githubusercontent.com/13671455/174300453-8abf2b40-b8d5-4df2-8856-06745ffe919c.PNG)

 What error in this code?
So I use pdb to debug this code,i find the call chain :
eventlet.green.os.fdopen() --> eventlet.greenio.py3.GreenPipe() --> _pyio.open() --> _pyio.BufferedWriter() --> eventlet.greenio.py3.GreenFileIO.writable() 
the code of "BufferedWriter" and “GreenFileIO”  class like this :
```python
class BufferedWriter(_BufferedIOMixin):
    def __init__(self, raw, buffer_size=DEFAULT_BUFFER_SIZE):
        if not raw.writable():
            raise OSError('"raw" argument must be writable.')

        _BufferedIOMixin.__init__(self, raw)
        if buffer_size <= 0:
            raise ValueError("invalid buffer size")
        self.buffer_size = buffer_size
        self._write_buf = bytearray()
        self._write_lock = Lock()

class GreenFileIO(_OriginalIOBase):
    def __init__(self, name, mode='r', closefd=True, opener=None):
         ……

    def writable(self):
        return 'w' in self._mode or '+' in self._mode
```
so GreenFileIO.writable will return false then BufferedWriter.__init__ raise OSError.
I think the function of eventlet.greenio.py3.GreenFileIO.writable() should cause consider the case when open file with mode ‘a’. may like this:
```python
class GreenFileIO(_OriginalIOBase):
[...]

PR #758, opened 2022-06-23
subsystem: fix bug:write a file with 'a' raise error

wuming0:
subsystem: fix bug:use eventlet.green.fdopen() to write a file with 'a' mode but raise "io.UnsupportedOperation: File or stream is not writable." error.

detail: eventlet.greenio.py3.GreenFileIO.writable() add the case——write a file with 'a' mode,so it judge the is or not to write and return excepted result.
issue: https://github.com/eventlet/eventlet/issues/757

### DIFF
--- a/eventlet/greenio/py3.py
+++ b/eventlet/greenio/py3.py
@@ -71,41 +71,6 @@ class GreenFileIO(_OriginalIOBase):
         return 'r' in self._mode or '+' in self._mode
 
     def writable(self):
-        """judge file can write or not
-
-        >>> from eventlet.green import os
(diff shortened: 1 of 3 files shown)


wuming0:
>  Think for replay and suggestion,I will remove those doc tests to 'tests/greenio_test.py'.

wuming0:
The doc tests  is moved to tests/greenio_test.py by me.

github_evenlet_mr_1

temoto:
FYI, this is same code.

```
# no thanks
self.assertEqual(excepted, actual, msg='actual=%s,excepted=%s' % (actual, excepted))

# yes please
assert actual == expected
```

PR #761, opened 2022-06-27
add some tests for issue #757

wuming0:
detail:1.use_assert_replace_assertEqual;2.add some testcase;
issue: https://github.com/eventlet/eventlet/issues/757

### DIFF
--- a/tests/greenio_test.py
+++ b/tests/greenio_test.py
@@ -1061,3 +1061,70 @@ def test_greenpipe_read_overwrite():
 
         actual = tests.read_file(f.name)
         assert actual == new_data
+
+
+class TestGreenFileIO(tests.LimitedTestCase):
+    my_os = eventlet.green.os
+
+    @tests.skip_on_windows
+    def setUp(self):
+        super(self.__class__, self).setUp()
+        self.tempdir = tempfile.mkdtemp('_green_file_io_test')
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
+        super(self.__class__, self).tearDown()
+
+    def test_mode_write(self):
+        filepath = os.path.join(self.tempdir, 'test_file_w.txt')
+        excepted = "Test write to file with mode 'w'."
+        with self.my_os.fdopen(self.my_os.open(filepath, os.O_RDWR | os.O_CREAT | os.O_TRUNC, 0o777), 'w') as fw:
+            fw.write(excepted)
+
+        actual = tests.read_file(filepath, mode='r')
+        assert actual == excepted, 'actual=%s,excepted=%s' % (actual, excepted)
+
+    def test_mode_append(self):
+        filepath = os.path.join(self.tempdir, 'test_file_a.txt')
+        old_data = 'Exist data...\n'
+        with open(filepath, 'w')as fw:
+            fw.write(old_data)
+        new_data = "Test write to file with mode 'a'."
+        with self.my_os.fdopen(self.my_os.open(filepath, os.O_RDWR | os.O_CREAT | os.O_APPEND, 0o777), 'a') as fa:
+            fa.write(new_data)
+
+        excepted = old_data + new_data
+        actual = tests.read_file(filepath, mode='r')
+        assert actual == excepted, 'actual=%s,excepted=%s' % (actual, excepted)
+
+    def test_mode_read_and_plus(self):
+        filepath = os.path.join(self.tempdir, 'test_file_r+.txt')
+        old_data = 'Exist data...\n'
+        with open(filepath, 'w') as fw:
+            fw.write(old_data)
+
+        new_data = "Test write to file with mode 'r+'."
+        with self.my_os.fdopen(self.my_os.open(filepath, os.O_RDWR | os.O_CREAT | os.O_APPEND, 0o777), 'r+') as fw:
+            fw.write(new_data)
+
+        excepted = old_data + new_data
+        actual = tests.read_file(filepath, mode='r')
+        assert actual == excepted, 'actual=%s,excepted=%s' % (actual, excepted)
+
+    def test_mode_write_and_plus(self):
+        filepath = os.path.join(self.tempdir, 'test_file_w+.txt')
+        excepted = "Test write to file with mode 'w+'."
+        with self.my_os.fdopen(self.my_os.open(filepath, os.O_RDWR | os.O_CREAT | os.O_APPEND, 0o777), 'w+') as fw:
+            fw.write(excepted)
+
+        actual = tests.read_file(filepath, mode='r')
+        assert actual == excepted, 'actual=%s,excepted=%s' % (actual, excepted)
+
+    def test_mode_append_and_plus(self):
+        filepath = os.path.join(self.tempdir, 'test_file_a+.txt')
+        excepted = "Test write to file with mode 'a+'."
+        with self.my_os.fdopen(self.my_os.open(filepath, os.O_RDWR | os.O_CREAT | os.O_APPEND, 0o777), 'a+') as fw:
+            fw.write(excepted)
+
+        actual = tests.read_file(filepath, mode='r')
+        assert actual == excepted, 'actual=%s,excepted=%s' % (actual, excepted)


wuming0:
github_evenlet_mr_2

tests/greenio_test.py
```diff
@@ -1061,3 +1061,70 @@ def test_greenpipe_read_overwrite():
 
         actual = tests.read_file(f.name)
         assert actual == new_data
+
+
+class TestGreenFileIO(tests.LimitedTestCase):
```
temoto, line 1066: please no:
- class
- mkdtemp
- `'actual=%s,excepted=%s' % (actual, excepted)`

temoto:
Thanks for extra tests! Did you find another problem?

wuming0:
According to those tests,we can find the bug of #757 is solved. I don't find any other problem.
But I have a idea,we can learn the way of linux how to deal file mode:use a interger to express the file mode.For examle: 
flg  --> mode
0b0000 --> r
0b0001 --> w
0b0011 --> a
0b0101 --> +
0b1000 --> b
......
then 'w+' --> 0b0001 or ob0101  --> 0b0101. So we can know the file can or not write by the last bit more quickly.

temoto:
If I understood correctly, you want to parse mode into integer first and then `return (self._mode_int & WRITABLE)`

```
def f1(s):
    return "w" in s or "a" in s or "+" in s

%timeit f1("r")
71.7 ns ± 0.114 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
%timeit f1("r+")
84.7 ns ± 0.83 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

def f2(s, _x=frozenset("wa+")):
    return bool(frozenset(s) & _x)

183 ns ± 0.821 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
207 ns ± 0.597 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

def f3(s, _r=re.compile(r"[wa+]")):
    return bool(_r.search(s))

153 ns ± 0.688 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
233 ns ± 0.294 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

def f4(i):
    return (i&1)!=0

81.8 ns ± 0.206 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
81.7 ns ± 0.181 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

So checking integer bits works at same speed as current code (f1). But also we'd spend CPU to parse string into integer.

This is fastest version:

```
class GreenFileIO(_OriginalIOBase):
    def __init__(self, name, mode='r', closefd=True, opener=None):
# ...
        self._readable = "r" in self._mode or "+" in self._mode
        self._seekable = None
        self._writable = "w" in self._mode or "+" in self._mode or "a" in self._mode

# ...

    def readable(self):
        return self._readable

    def writable(self):
        return self._writable
```

wuming0:
The result is exceeding my expectations.If the test is exact,"the fastest version" of reply is the best way in this situation.But why f1() is faster than f4()?Let we see the bytecode first.
```
dis.dis("""def f1(s):                                                                                     |dis.dis("""def f4(i):
    return "w" in s or "a" in s or "+" in s""")                                                           |    return (i & 1) != 0""")
  1           0 LOAD_CONST               0 (<code object f1 at 0x00000265D998A710, file "<dis>", line 1>) |  1           0 LOAD_CONST               0 (<code object f4 at 0x00000265D9989660, file "<dis>", line 1>)
              2 LOAD_CONST               1 ('f1')                                                         |              2 LOAD_CONST               1 ('f4')
              4 MAKE_FUNCTION            0                                                                |              4 MAKE_FUNCTION            0
              6 STORE_NAME               0 (f1)                                                           |              6 STORE_NAME               0 (f4)
              8 LOAD_CONST               2 (None)                                                         |              8 LOAD_CONST               2 (None)
             10 RETURN_VALUE                                                                              |             10 RETURN_VALUE
Disassembly of <code object f1 at 0x00000265D998A710, file "<dis>", line 1>:                              |Disassembly of <code object f4 at 0x00000265D9989660, file "<dis>", line 1>:
  2           0 LOAD_CONST               1 ('w')                                                          |  2           0 LOAD_FAST                0 (i)
              2 LOAD_FAST                0 (s)                                                            |              2 LOAD_CONST               1 (1)
              4 CONTAINS_OP              0                                                                |              4 BINARY_AND
              6 JUMP_IF_TRUE_OR_POP     22                                                                |              6 LOAD_CONST               2 (0)
              8 LOAD_CONST               2 ('a')                                                          |              8 COMPARE_OP               3 (!=)
             10 LOAD_FAST                0 (s)                                                            |             10 RETURN_VALUE
             12 CONTAINS_OP              0                                                                |
             14 JUMP_IF_TRUE_OR_POP     22                                                                |
             16 LOAD_CONST               3 ('+')                                                          |
             18 LOAD_FAST                0 (s)                                                            |
             20 CONTAINS_OP              0                                                                |
        >>   22 RETURN_VALUE  
```
According to bytecode,it's seem no reason the f1() will faster than f4() in most sutiation.only except  a few situation like f1('w').So  I do a test use below code:
```python
import timeit
from time import time


def cost_warp(func):
    def inner(*args, **kwargs):
        cost_list = []
        result = None
        for i in range(timeit.default_repeat):
            start_at = time()
            for i in range(timeit.default_number):
                result = func(*args, **kwargs)
            cost_list.append(time() - start_at)
        print(f'avg cost {sum(cost_list) / len(cost_list)} s of {func}{args}')
        return result

    return inner


@cost_warp
def f1(s):
    return "w" in s or "a" in s or "+" in s


@cost_warp
def f4(i):
    return (i & 1) != 0


def measure_by_timeit():
    print(timeit.timeit("""def f1(s):
    return "w" in s or "a" in s or "+" in s
    f1('w')"""), "\t:f1('w')-timeit")
    print(timeit.timeit("""def f1(s):
    return "w" in s or "a" in s or "+" in s
    f1('r')"""), "\t:f1('r')-timeit")
    print(timeit.timeit("""def f1(s):
    return "w" in s or "a" in s or "+" in s
    f1('r+')"""), "\t:f1('r+')-timeit")
    print(timeit.timeit("""def f4(i):
    return (i & 1) != 0
    f4(3)"""), "\t:f4(3)-timeit")


def measure_by_my():
    f1('w')
    f1('r')
    f1('r+')
    f4(3)


def main():
    measure_by_timeit()
    measure_by_my()

if __name__ == "__main__":
    main()
```
I run several  time,the result of  measure_by_timeit() not stable——sometime f1() faster than f4(),but sometime f1() slow than f4(). 
the  result measure_by_my() is very stable——f1() always slow than f4(). This result is fit the except of I after read above bytecode.

I read the timeit.timeit() in python doc with confuse,I observe a note of timeit.timeit() :
```
Note

By default, [timeit()](https://docs.python.org/3/library/timeit.html?highlight=timeit#timeit.Timer.timeit) temporarily turns off [garbage collection](https://docs.python.org/3/glossary.html#term-garbage-collection) during the timing. The advantage of this approach is that it makes independent timings more comparable. The disadvantage is that GC may be an important component of the performance of the function being measured. If so, GC can be re-enabled as the first statement in the setup string. For example:
timeit.Timer('for i in range(10): oct(i)', 'gc.enable()').timeit()
```
so I guess,maybe the gc or other things disturb the accuracy of test.

fig 1. f1() faster than f4()
github_evenlet_reply_1

fig 2. f1() slow than f4()
github_evenlet_reply_2

temoto:
Maybe you come from C or similar low level language background. In Python `(i & 1) != 0` is **very different** from `and ax, 1; cmp ax 0` on CPU, because everything is object, numbers may be infinitely large, there is invisible virtual machine overhead.

- `dis` nicely shows python bytecode, but say **nothing** about speed of execution.
- `time.time` to show current time for humans; `time.perf_counter` for benchmarks, `time.monotonic` for anything else
- `timeit.timeit(statement, setup)` will execute statement N times, setup 1 time. Your code executes `def` N times. Use `timeit.timeit("f1()", "def f1...")` or `def f1()... timeit.timeit("f1()", globals=globals())` or ipython `%timeit`
- IMHO, in microbenchmark like this, everything above best (min) time is hardware/scheduling noise

```python
import statistics, sys, timeit


def f1(s):
    return "w" in s or "a" in s or "+" in s


def f4(i):
    return (i & 1) != 0


code_in_or = '"w" in s or "a" in s or "+" in s'
code_bitmask = '(i & 1) != 0'

N = 1000000
REPEAT = 15


def format_result(label, times):
    speeds = [t * 1e9 / N for t in times]
    best = min(speeds)
    var = statistics.variance(speeds)
#    return f"{label:8s}\t{speeds}"  # to collect data for multiple runs
    return f"{label:8s}\t{best:.1f}\t\t{var:.1f}"


if __name__ == "__main__":
    print("what\t\tbest ns/op\tvariance")
    print("# function")
    for stmt in ("f1('r')", "f1('r+')", "f4(2)", "f4(3)"):
        times = timeit.repeat(stmt, number=N, globals=globals(), repeat=REPEAT)
        print(format_result(stmt, times))
    print("# inline")
    for stmt, setup in [(code_in_or, "s='r'"), (code_in_or, "s='r+'"), (code_bitmask, "i=2"), (code_bitmask, "i=3")]:
        times = timeit.repeat(stmt, setup=setup, number=N, repeat=REPEAT)
        print(format_result(f"{stmt.split()[1]}{setup.split('=')[1]}", times))
```

Run this script 5 times, aggregate best and variance:

```
what		best ns/op	variance
# function
f1('r') 	69.4		2.8
f1('r+')	78.2		1.7
f4(2)   	76.4		0.9
f4(3)   	76.3		1.1
# inline
in'r'   	34.9		0.2
in'r+'  	35.9		0.1
&2      	38.5		0.3
&3      	37.5		0.2
```

IMHO, it says "70..78 ~= 76" or "35..36 ~= 38..37", same, not faster.

tests/greenio_test.py
```diff
@@ -1061,3 +1061,23 @@ def test_greenpipe_read_overwrite():
 
         actual = tests.read_file(f.name)
         assert actual == new_data
+
+
+def test_greenpipe_write_plus():
+    excepted = "write with mode=w+"
```
temoto, line 1067: typo

```suggestion
    expected = "write with mode=w+"
```

exCept close to but/different. exPect close to prediction/wish.

Example dataset: [1,3,5,7,2]
- all numbers are odd except 2; `try/except/finally`; `raise Exception()`
- after 1,3,5,7 we expect 9; 2 is unexpected; expectation is broken

temoto:
"fastest version" is better not because `or` is faster than bitmask, but because you calculate `readable/writable` 1 time. It could be 

```python
def __init__(self, ...):
    self._readable = f1() or regex or bitmask or time.sleep(0.1) or   # anything slow

def readable(self): return self._readable
```

and it would still be faster than anything else in `readable` function.

tests/greenio_test.py
wuming0: I misspelled words of excepted and expected,I had correct it.Forgive my poor english.

wuming0:
Think you very much! Your reply answered my doubt:Why the result of the speed of f1() and f4() use by timeit to evaluate is not stable

I use timeit.timeit() in a wrong way.After I fix it and use "timeit.timeit("f1('r')", globals=globals())" or "timeit.repeat("f1('r')", globals=globals(), number=100_0000, repeat=15)" to evaluate the speed of two function.

I use average of result instead  best of result as evaluation criterion,then i got a more stable result and also match my guess of the bytecode of two function:f1('w') > f4(3) > f1('r+') > f1('r') ,The gap is small.

Think again for your replay,let me learn a lot.

temoto:
Thank you.

PR #766, opened 2022-07-26
Stop using deprecated camelCase threading APIs

tipabu:
`Thread.daemon` was introduced as a replacement for `isDaemon`/`setDaemon` in py26, and py310 started emitting `DeprecationWarning`s about it.


temoto (approved):
thank you

PR #771, opened 2022-11-04
Fix WSGI test

tipabu:
I'm not sure how/why this test stopped passing, but I don't think the `send()`-on-close behavior is really the interesting part.


temoto:
Likely it was caused by Linux kernel upgrade, could you believe it?

https://lore.kernel.org/all/21869cb9-d1af-066a-ba73-b01af60d9d3a@kernel.org/

I hope they went with good stuff despite breaking userspace formality.

temoto (approved):

temoto:
merged, thank you

PR #776, opened 2022-11-24
Fix leak of epoll file descriptors

felixhuettner:
When a python process creates a new thread that uses an eventlet hub (e.g. when using `trampoline`) a new epoll file descriptor is allocated. If the thread now stops after a short amount of work the epoll file descriptor is not closed causing a leak.

If the python process (but not the thread) is long living and regularly spawns a lot of short-lived threads then this can cause an exhaustion of file descriptors.

To solve this the epoll file descriptor is now no longer allocated on hub creation but only when a listener is actually added. We also now close the file descriptor if there are no listeners left.

### DIFF
--- a/eventlet/hubs/epolls.py
+++ b/eventlet/hubs/epolls.py
@@ -29,3 +29,7 @@ class Hub(poll.Hub):
 
     def do_poll(self, seconds):
         return self.poll.poll(seconds)
+
+    def __del__(self):
+        self.poll.close()
+
--- a/tests/hub_test.py
+++ b/tests/hub_test.py
@@ -402,3 +402,7 @@ def test_kqueue_unsupported():
     # https://github.com/eventlet/eventlet/issues/38
     # get_hub on windows broken by kqueue
     tests.run_isolated('hub_kqueue_unsupported.py')
+
+
+def test_epoll_fd_leak():
+    tests.run_isolated('hub_epoll_file_descriptor_leak.py')
--- a/tests/isolated/hub_epoll_file_descriptor_leak.py
+++ b/tests/isolated/hub_epoll_file_descriptor_leak.py
@@ -1,0 +1,62 @@
+__test__ = False
+
+if __name__ == '__main__':
+
+    import eventlet
+    from eventlet import hubs
+    import eventlet.hubs
+    import threading
+    import os
+    import gc
+
+    rfd, wfd = os.pipe()
+    reader = eventlet.greenio.py3.GreenPipe(rfd, "rb", buffering=2)
+
+    def epoll_count():
+        epoll_count = 0
+        for f in os.listdir("/proc/self/fd"):
+            try:
+                p = os.path.join("/proc/self/fd", f)
+                if os.readlink(p) == "anon_inode:[eventpoll]":
+                    epoll_count += 1
+            except:
+                pass
+        return epoll_count
+
+    def fun(fd):
+        # separate trampoline may trigger additional bugs with multiple
+        # listeners
+        hubs.trampoline(fd, write=True)
+
+        with eventlet.greenio.py3.GreenPipe(fd, "wb", buffering=0) as w:
+            w.write(b"12")
+
+
+        # it would be wasteful to drop and recreate hub state every time
+        # all listeners stop
+        epolls = epoll_count()
+        assert epolls == 2, ('epoll file descriptor closed while it might '
+            'still be used. Currently having {}'.format(epolls))
+
+        hubs._threadlocal.hub.abort(wait=True)
+
+        # this makes it work
+        #hubs._threadlocal.hub.__del__()
+        # this one does not help
+        #del hubs._threadlocal.hub
+
+    t = threading.Thread(target=fun, args=(wfd,))
+    t.start()
+
+    assert reader.read(2) == b"12"
+    reader.close()
+
+    t.join()
+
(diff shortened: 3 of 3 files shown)


temoto:
> To solve this the epoll file descriptor is now no longer allocated on hub creation but only when a listener is actually added.

With respect to outstanding job on patch and testing, I think epoll lazy evaluation is completely unrelated to fd leak and should be done via "cached property" approach, so you can't forget to initialize it in some methods. And importantly, please make lazy hub initialization in separate patch, because it will touch all hubs and requires a lot of testing.

tests/isolated/hub_epoll_file_descriptor_leak.py
```diff
@@ -0,0 +1,56 @@
...
+        first_listener = eventlet.spawn(readfd, r)
+        eventlet.sleep()
+
+        os.write(w, b'.')
+        assert first_listener.wait() == b'.'
+
+    def start_thread():
+        threading.Thread(None, test_in_thread).start()
+
+    pool = eventlet.GreenPool()
```
temoto, line 38: This GreenPool does nothing. Thread is not joined. There is no check it was ever run.

```
rfd, wfd = os.pipe()
reader = eventlet.GreenPipe(rfd, "rb", buffering=2)

def fun(fd):
    # separate trampoline may trigger additional bugs with multiple listeners
    hubs.trampoline(fd, write=True)

    with eventlet.GreenPipe(fd, "wb", buffering=0) as w:
        w.write(b"2")

    # it would be wasteful to drop and recreate hub state every time all listeners stop
     assert epoll_count() == 2


t = threading.Thread(target=fun, args=(wfd,))
t.start()

assert reader.read(2) == b"12"
reader.close()

t.join()
eventlet.sleep()

# only now after thread is done we can check it's resources are freed
assert epoll_count() == 1
```

temoto (changes requested):
this is pretty important

eventlet/hubs/epolls.py
```diff
@@ -29,3 +31,11 @@ def add(self, evtype, fileno, cb, tb, mac):
 
     def do_poll(self, seconds):
         return self.poll.poll(seconds)
+
+    def remove(self, listener):
+        if all([len(x) == 0 for x in self.listeners.values()]):
+            return
+        super().remove(listener)
+        if all([len(x) == 0 for x in self.listeners.values()]):
+            self.poll.close()
```
temoto, line 40: Imagine workflow:

- talk some network
- local processing
- at some point there is no more listeners so epoll gets closed
- talk some network again ...

Imagine this goes in hot loop of single use create/close, like DNS client or kind of monitoring agent.

Long story short, last fd closed is not a good signal for resource cleanup.

End of thread should call `Hub.abort()` which is proper way to release all hub resources.

AFAIK, Python memory management should call `Hub.__del__` when it goes out of reach of non-weak references. So fd leak may be due to dangling reference or we need `__del__` of `self.abort()`

temoto (changes requested):

tests/hub_test.py
```diff
@@ -402,3 +402,9 @@ def test_kqueue_unsupported():
     # https://github.com/eventlet/eventlet/issues/38
     # get_hub on windows broken by kqueue
     tests.run_isolated('hub_kqueue_unsupported.py')
+
+
+def test_epoll_fd_leak():
+    # https://github.com/eventlet/eventlet/issues/38
+    # get_hub on windows broken by kqueue
```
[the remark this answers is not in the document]
felixhuettner: fixed

tests/isolated/hub_epoll_file_descriptor_leak.py
felixhuettner: thanks. i now adopted it

eventlet/hubs/epolls.py
felixhuettner: ok, so i have now a working version. The important part in the test is the call to `hubs._threadlocal.hub.abort(wait=True)`.
Without this call (or without the `wait=True`) we do not switch to the `BaseHub.run` method and allow it to exit.
When the thread then exits the `BaseHub.run` method still references `self` (so the Hub) causing it to never be garbage collected.

However i'm still not happy with this solution as it requires an explicit call to `abort(wait=True)` at the end of your thread. This is not easily possible if e.g. eventlet is used in a logging library and therefor called without the main code knowing about eventlet at all

felixhuettner: actually i just realized that we do not leak file descriptors at all if `abort(wait=True)` is called. It seems like the default epoll implementation already closes the filedescriptor for us

felixhuettner: i would therefor close this PR. Thanks very much for all the information you provided

temoto: > However i'm still not happy with this solution as it requires an explicit call to abort(wait=True) at the end of your thread.

Me too! If you want to spend more time, these things should help:

- `sys.getrefcount(hub)` returns internal reference counter. It's +1 to call getrefcount function, so 1 is the lowest possible number for dead object and 2 means something else holds ref to object.
- `gc.get_referrers` basically is what you need, but read description, it will not show some refs.
- `objgraph` library.

felixhuettner: Thanks for the hints. While i will not work on it right away i'll keep it in my mind if we hit this issue again

Issue #692, opened 2021-03-19
Eventlet raises a different exception on SSL timeouts vs standard Python


ShaneHarvey:
With standard Python 3.2+ (not monkey patched) timeouts on ssl connections raise `socket.timeout: The read operation timed out`. `socket.timeout` is a standard class that applications can use to identify network timeouts vs other network errors. For example:
```python
# import eventlet
# eventlet.monkey_patch()

import socket
import ssl

sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, socket.SOL_TCP)
sock.settimeout(2)  # Set timeout to 2 seconds.
sock.connect(("www.python.org", 443))
print('client socket connected')
client_context = ssl.create_default_context()
client_conn = client_context.wrap_socket(sock, server_hostname="www.python.org")
print('client TLS connected')
client_conn.read()  # This line will timeout.
```

The above example produces to following output:
```
$ python3 eventlet_ssl_timeout_bug.py
client socket connected
client TLS connected
Traceback (most recent call last):
  File "eventlet_ssl_timeout_bug.py", line 14, in <module>
    client_conn.read()  # This line will timeout.
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ssl.py", line 1103, in read
    return self._sslobj.read(len)
socket.timeout: The read operation timed out
```

However when we uncomment the monkey patching, eventlet raises the more general `ssl.SSLError` exception:
```
$ python eventlet_ssl_timeout_bug.py
client socket connected
client TLS connected
Traceback (most recent call last):
  File "eventlet_ssl_timeout_bug.py", line 14, in <module>
    client_conn.read()  # This line will timeout.
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/eventlet/green/ssl.py", line 186, in read
    return self._call_trampolining(
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/eventlet/green/ssl.py", line 164, in _call_trampolining
    trampoline(self,
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/eventlet/hubs/__init__.py", line 159, in trampoline
    return hub.switch()
  File "/Users/shane/pymongo-pycharm-3.8/lib/python3.8/site-packages/eventlet/hubs/hub.py", line 313, in switch
    return self.greenlet.switch()
ssl.SSLError: ('timed out',)
```

[...]

PR #778, opened 2022-12-21
SSL socket should raise socket.timeout exception on timeout instead of SSLError

baszoetekouw:
Back in Python 3.2, ssl.SSLError used to be a subclass of socket.error (see https://docs.python.org/3/library/ssl.html#exceptions), so timeouts on monkeypatched ssl sockets would be properly caught by socket.timeout exception handlers in applications.  However, since Python 3.3 ssl.SSLerror is a subclass of OSError, which signifies a different (typically fatal) type of error that is usually not handled gracefully by applications.

By changing the timeout exception back to socket.timeout, libraries such as pymysql and redis will again properly support TLS-connections in monkeypatched applications.

Fixes eventlet/eventlet#692

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -15,7 +15,7 @@ from contextlib import contextmanager
 
 orig_socket = __import__('socket')
 socket = orig_socket.socket
-timeout_exc = SSLError
+timeout_exc = orig_socket.timeout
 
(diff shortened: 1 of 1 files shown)


temoto (approved):
LGTM

temoto:
merged, thank you

Issue #781, opened 2023-01-15
Eventlet fails to patch dnspython 2.3.0 


potiuk:
About 2 hours ago, dnspython 2.3.0 has been released and our "canary" builds in Apache Airflow detected incompatibilities of eventlet and the new dnspython.

This one is triggered by cassandra-driver which tries to import eventlet and during the import eventlet fils with this error during pytest import:

https://github.com/apache/airflow/actions/runs/3924777720/jobs/6709334403

```
_ ERROR collecting tests/system/providers/apache/cassandra/example_cassandra_dag.py _
tests/system/providers/apache/cassandra/example_cassandra_dag.py:28: in <module>
    from airflow.providers.apache.cassandra.sensors.record import CassandraRecordSensor
airflow/providers/apache/cassandra/sensors/record.py:26: in <module>
    from airflow.providers.apache.cassandra.hooks.cassandra import CassandraHook
airflow/providers/apache/cassandra/hooks/cassandra.py:24: in <module>
    from cassandra.cluster import Cluster, Session
cassandra/cluster.py:101: in init cassandra.cluster
    ???
/usr/local/lib/python3.7/site-packages/cassandra/io/eventletreactor.py:18: in <module>
    import eventlet
/usr/local/lib/python3.7/site-packages/eventlet/__init__.py:17: in <module>
    from eventlet import convenience
/usr/local/lib/python3.7/site-packages/eventlet/convenience.py:7: in <module>
    from eventlet.green import socket
/usr/local/lib/python3.7/site-packages/eventlet/green/socket.py:21: in <module>
    from eventlet.support import greendns
/usr/local/lib/python3.7/site-packages/eventlet/support/greendns.py:66: in <module>
    setattr(dns, pkg, import_patched('dns.' + pkg))
/usr/local/lib/python3.7/site-packages/eventlet/support/greendns.py:61: in import_patched
    return patcher.import_patched(module_name, **modules)
/usr/local/lib/python3.7/site-packages/eventlet/patcher.py:135: in import_patched
    *additional_modules + tuple(kw_additional_modules.items()))
/usr/local/lib/python3.7/site-packages/eventlet/patcher.py:109: in inject
    module = __import__(module_name, {}, {}, module_name.split('.')[:-1])
/usr/local/lib/python3.7/site-packages/dns/zone.py:86: in <module>
    class Zone(dns.transaction.TransactionManager):
/usr/local/lib/python3.7/site-packages/dns/zone.py:757: in Zone
[...]

PR #782, opened 2023-01-17
Fix greendns for dnspython 2.3.0+

tipabu:
`dnspython` added more type annotations in 2.3.0 -- because of how we reimport the package piece by piece, though, not all of the types would be available during reimport, leading to `AttributeErrors` like

    module 'dns.rdtypes' has no attribute 'ANY'

Now, do all of our `rdtypes` special-handling *first*, before reimporting any other submodules.

Addresses #781


temoto (approved):
Seems like obviously correct.

PR #783, opened 2023-01-17
Fix tox error

tipabu:
`tox>=4.0.0` changed `passenv` from a space-separated list to a comma-separated list. `tox>=4.0.6` made it a hard error to include spaces, complaining

    pass_env values cannot contain whitespace, use comma to have multiple values in a single line

Switch to using multiple lines for the multiple variables to be compatible with both tox3 and tox4.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,7 +45,9 @@ commands =
     pep8 benchmarks/ eventlet/ tests/
 
 [testenv]
-passenv = CI EVENTLET_DB_TEST_AUTH
+passenv =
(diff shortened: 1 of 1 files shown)


temoto (approved):
thanks, i was trying to fix CI at the same time :-)

temoto:
turns out it was already merged together with dnspython 2.3.0 fix

temoto:
previous statement is false, i made wrong move with branches

this patch is merged in master now

Issue #763, opened 2022-06-28
deprecation warning in eventlet/support/greenlets.py about distutils


tmijieux:
https://github.com/eventlet/eventlet/blob/f0a887b94a86f9567e33037646712b89f02ae441/eventlet/support/greenlets.py#L6-L7

it seems to cause no particular issue at the moment, i'm setting up this issue just in case this may be important ?
```
  C:\src\...\venv\lib\site-packages\eventlet\support\greenlets.py:6: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    preserves_excinfo = (distutils.version.LooseVersion(greenlet.__version__)

  C:\src\...\venv\lib\site-packages\eventlet\support\greenlets.py:7: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    >= distutils.version.LooseVersion('0.3.2'))
```
the python versions where i encountered this warning:
```
$ python -VV
Python 3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)]
```

versions of eventlet that encoutered the issue
eventlet==0.33.1
eventlet @ https://github.com/eventlet/eventlet/archive/f0a887b94a86f9567e33037646712b89f02ae441.zip
(master at 2022-06-28)

PR #786, opened 2023-01-22
dep: greenlet>=1.0 removing unused clear_sys_exc_info stub

temoto:
fixes https://github.com/eventlet/eventlet/issues/763

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ setuptools.setup(
     packages=setuptools.find_packages(exclude=['benchmarks', 'tests', 'tests.*']),
     install_requires=(
         'dnspython >= 1.15.0',
-        'greenlet >= 0.3',
+        'greenlet >= 1.0',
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -19,7 +19,7 @@
     packages=setuptools.find_packages(exclude=['benchmarks', 'tests', 'tests.*']),
     install_requires=(
         'dnspython >= 1.15.0',
-        'greenlet >= 0.3',
+        'greenlet >= 1.0',
```
tipabu, line 22: Oh, 1.0.0 was more recent than I'd realized -- Jan 2021. Still, it's a couple years old; seems like a reasonable up-rev. *I'll* still want to watch out for it, but I'm stuck running old software (like py2) :stuck_out_tongue:

And yeah, once we've got that, `preserves_excinfo` would always be `True` in the old code, so we can rip all the workarounds out.

tipabu (approved):

PR #788, opened 2023-02-10
wsgi: Allow keepalive option to be a timeout

tipabu:
Sometimes the server pool fills up; all `max_size` greenthreads are already in use with other connections. One more connection gets `accept()`ed, but then has to wait for a slot to open up to actually be handled.

This works out fine if your clients tend to make a single request then close the connection upon receiving a response. It works out OK-ish when clients are continually pipelining requests; the new connection still has to wait, but at least there's plenty of work getting processed -- it's defensible. It can work out pretty terribly if clients tend to hold on to connections "just in case" -- we're ignoring fresh work from a new client just so we can be ready-to-go if an existing connection wakes up.

There are a couple existing tunings we can use, but they can each have downsides:

- Increasing `max_size` is nice for dealing with idle connections, but can cause hub contention and high latency variance when all those connections are actually busy.
- `socket_timeout` can be used to limit the idle socket time, but it *also* impacts `send`/`recv` operations while processing a request, which may not be desirable.
- `keepalive` can be set to `False`, disabling request pipelining entirely.

Change the `keepalive_timeout` option to `eventlet.wsgi.server` so it can be the timeout to use while waiting for a new request, separate from `socket_timeout`. By default, `socket_timeout` continues to be used.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -392,7 +392,12 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
             return ''
 
         try:
-            return self.rfile.readline(self.server.url_length_limit)
+            sock = self.rfile._sock if six.PY2 else self.connection
+            if self.server.keepalive and not isinstance(self.server.keepalive, bool):
+                sock.settimeout(self.server.keepalive)
+            line = self.rfile.readline(self.server.url_length_limit)
+            sock.settimeout(self.server.socket_timeout)
+            return line
         except greenio.SSL.ZeroReturnError:
             pass
         except socket.error as e:
@@ -521,6 +526,10 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
                     towrite.append(b'Connection: close\r\n')
                 elif send_keep_alive:
                     towrite.append(b'Connection: keep-alive\r\n')
+                    # Spec says timeout must be an integer, but we allow sub-second
+                    int_timeout = int(self.server.keepalive or 0)
+                    if not isinstance(self.server.keepalive, bool) and int_timeout:
+                        towrite.append(b'Keep-Alive: timeout=%d\r\n' % int_timeout)
                 towrite.append(b'\r\n')
                 # end of header writing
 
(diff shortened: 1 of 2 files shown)


eventlet/wsgi.py
```diff
@@ -799,6 +805,7 @@ def __init__(self,
         self.url_length_limit = url_length_limit
         self.debug = debug
         self.socket_timeout = socket_timeout
+        self.idle_timeout = idle_timeout
```
temoto, line 808:
```suggestion
        self.idle_timeout = idle_timeout or socket_timeout
```

doc says default is to use socket_timeout

eventlet/wsgi.py
```diff
@@ -967,6 +977,7 @@ def server(sock, site,
         url_length_limit=url_length_limit,
         debug=debug,
         socket_timeout=socket_timeout,
+        idle_timeout=idle_timeout,
```
temoto, line 980: Personally I like "idle" term better, but it's widely recognized as "keepalive".

idle_timeout -> keepalive_timeout ?

and then we could gradually surrender the `keepalive=False` option to `..._timeout=0`

eventlet/wsgi.py
tipabu: Sure; then we can also get rid of the `if self.idle_timeout:` guards -- I mainly went this way because I wasn't sure whether there'd be some penalty for calling `settimeout` a bunch, or if so, how big it was.

eventlet/wsgi.py
tipabu: Sure -- I think that opt had me thinking of `SO_KEEPALIVE`, and I forgot that I already [take care of that](https://github.com/openstack/swift/blob/2.31.0/swift/common/wsgi.py#L215) before calling `server` 😃

Makes me realize I might consider sending a [`Keep-Alive` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive), and maybe also adding a `keepalive_max` opt.

tipabu: Renamed, and started the deprecation process. Also went ahead and added the `Keep-Alive` header, provided that we were going to send the `Connection` header anyway.

eventlet/wsgi.py
```diff
@@ -392,7 +394,17 @@ def _read_request_line(self):
             return ''
 
         try:
-            return self.rfile.readline(self.server.url_length_limit)
+            if self.server.keepalive_timeout:
```
tipabu, line 397: Should we skip this for the first request on a connection?

temoto: I'd expect keepalive timeout to work on new connection without requests too.

eventlet/wsgi.py
temoto: Yes, syscall penalty is non-negligible (around 2us best case). Also I think idle timeout is on verge of DoS resilience, and the only faster way I know is `monotonic() > deadline` which gets time via VDSO without syscall in best case.

I think we should document the cost and recommend (as always) external reverse proxy, which downgrades disabled keepalive from DoS to internal config/resource management.

eventlet/wsgi.py
temoto: Seems convenient, can't think of issues with double typed bool/float `keepalive`.

temoto (approved):
LGTM

temoto:
Thank you.

PR #789, opened 2023-02-12
fix truncate size can be None

claudius-kienle:
If I execute a script that contains following code with gunicorn and eventlet, I get the exception that 
```bash
File "/usr/local/lib/python3.8/site-packages/eventlet/greenio/py3.py", line 165, in truncate
--
Sun, Feb 12 2023 8:54:19 pm | rv = _original_os.ftruncate(self._fileno, size)
Sun, Feb 12 2023 8:54:19 pm | TypeError: an integer is required (got type NoneType)
```

```python
with os.fdopen(os.open(path, os.O_WRONLY | os.O_CREAT, 0o600), 'w') as file:
            file.seek(0)
            file.write(json.dumps(token_data, indent=2))
            file.truncate()
```

Consequently, I assumed that `truncate()` must also work if `size=None`.
This pull request checks if `size` is None and in this case replaces it with `-1`.


temoto:
> Consequently, I assumed that truncate() must also work if size=None.

Why?

temoto:
never mind, it's "because CPython does so" https://github.com/python/cpython/blob/main/Modules/_io/bufferedio.c#L1287


thanks for patch

itamarst (approved):

4383 (approved):

PR #794, opened 2023-03-23
Replace distutils with packaging

stephenfin:
distutils is deprecated.

    DeprecationWarning: distutils Version classes are deprecated. Use
    packaging.version instead.
      >= distutils.version.LooseVersion('0.3.2'))

Do as the warning suggests and use packaging.version instead.


stephenfin:
Yup, much better idea.

Issue #795, opened 2023-04-03
[Python 3.12] AttributeError: module 'ssl' has no attribute 'wrap_socket'


hrnciar:
Hello, Fedora is rebuilding all Python packages with upcoming Python 3.12. Eventlet fails with `AttributeError: module 'ssl' has no attribute 'wrap_socket'`. This happens because of a change in Python.

From [What's new](https://docs.python.org/3.12/whatsnew/3.12.html#removed):
Remove the ssl.wrap_socket() function, deprecated in Python 3.7: instead, create a [ssl.SSLContext](https://docs.python.org/3.12/library/ssl.html#ssl.SSLContext) object and call its [ssl.SSLContext.wrap_socket](https://docs.python.org/3.12/library/ssl.html#ssl.SSLContext.wrap_socket) method. Any package that still uses ssl.wrap_socket() is broken and insecure. The function neither sends a SNI TLS extension nor validates server hostname. Code is subject to [CWE-295](https://cwe.mitre.org/data/definitions/295.html): Improper Certificate Validation. (Contributed by Victor Stinner in [gh-94199](https://github.com/python/cpython/issues/94199).)

~~~~
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/sphinx/config.py", line 350, in eval_config_file
    exec(code, namespace)
  File "/builddir/build/BUILD/eventlet-0.33.3/doc/conf.py", line 52, in <module>
    import eventlet
  File "/builddir/build/BUILD/eventlet-0.33.3/eventlet/__init__.py", line 17, in <module>
    from eventlet import convenience
  File "/builddir/build/BUILD/eventlet-0.33.3/eventlet/convenience.py", line 7, in <module>
    from eventlet.green import socket
  File "/builddir/build/BUILD/eventlet-0.33.3/eventlet/green/socket.py", line 21, in <module>
    from eventlet.support import greendns
  File "/builddir/build/BUILD/eventlet-0.33.3/eventlet/support/greendns.py", line 45, in <module>
    from eventlet.green import ssl
  File "/builddir/build/BUILD/eventlet-0.33.3/eventlet/green/ssl.py", line 25, in <module>
    _original_wrap_socket = __ssl.wrap_socket
                            ^^^^^^^^^^^^^^^^^
AttributeError: module 'ssl' has no attribute 'wrap_socket'
~~~~

akshay-joshi:
Hello All,

Any update on this issue? Any ETA?
      OR
[...]

PR #797, opened 2023-05-04
WIP Python 3.12 fixes

hroncok:
This makes the tests pass in Fedora. I have no ide what I am doing, so it may be entirely wrong.

This probably breaks with older Pythons.

Fixes https://github.com/eventlet/eventlet/issues/795

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -22,21 +22,17 @@ __patched__ = [
     'create_default_context', '_create_default_https_context']
 
 _original_sslsocket = __ssl.SSLSocket
-_original_wrap_socket = __ssl.wrap_socket
 _original_sslcontext = getattr(__ssl, 'SSLContext', None)
 _is_under_py_3_7 = sys.version_info < (3, 7)
 
 
 @contextmanager
 def _original_ssl_context(*args, **kwargs):
-    tmp_sslcontext = _original_wrap_socket.__globals__.get('SSLContext', None)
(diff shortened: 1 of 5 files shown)


eventlet/green/ssl.py
```diff
@@ -76,16 +72,21 @@ def __new__(cls, sock=None, keyfile=None, certfile=None,
...
+                    context = _original_sslcontext(protocol=ssl_version)
+                    context.options |= cert_reqs
+                    if certfile or keyfile:
+                        context.load_cert_chain(
+                            certfile=certfile,
+                            keyfile=keyfile,
+                        )
+                    if ca_certs:
+                        context.load_verify_locations(ca_certs)
+                    if ciphers := kw.get('ciphers'):
```
temoto, line 84: sorry, this is 3.8+ construct, not going to happen anytime soon

hroncok: I know. My aim for this draft was to fix this for 3.12. Maintaining compatibility with previous Python versions was out of scope. If we can determine this approach is *correct*, we can add a giant if around the code. If the walrus would cause a problem even inside an if, no problem removing it.

itamarst:
Hi, what needs to be done to get this merged?

temoto:
> Hi, what needs to be done to get this merged?

py3.7 compat (no `:=`) is a good start

itamarst:
Superseded by #817, which builds on this. Thank you for all your work!

PR #809, opened 2023-09-17
greendns: fix getaddrinfo parameter name for Python 3

### LINKED ISSUE #775: Eventlet monkeypatch replacement of socket.getaddrinfo has an incorrectly named parameter, leading to TypeError
* eventlet version: 0.33.2
* Python version: 3.9.5
* Operating System: macOS Monterey 12.6.1
### Description

In Python 3.2, the [socket.getaddrinfo](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) parameter `socktype` was renamed to `type`.  
As a result, code that calls `socket.getaddrinfo`, specifying the type parameter as a keyword argument, raises a TypeError, if  `eventlet.monkey_patch()` was called previously.

### How to reproduce
```python
$ python3
Python 3.9.5 (default, Jun 29 2021, 13:12:16) 
[Clang 12.0.5 (clang-1205.0.22.11)] on darwin
>>> import eventlet; eventlet.monkey_patch()
>>> from smtpd import SMTPServer
>>> SMTPServer(("127.0.0.1", 25), None)
  File "/Users/user/.pyenv/versions/3.9.5/lib/python3.9/smtpd.py", line 645, in __init__
    gai_results = socket.getaddrinfo(*localaddr,
TypeError: getaddrinfo() got an unexpected keyword argument 'type'
```

tuokri:
Fixes #775.

Used this PR from Gevent as reference: https://github.com/gevent/gevent/pull/961.


itamarst:
Thank you for the patch! At this point we're only supporting Python 3, do you think you could update the PR accordingly?

tuokri:
> Thank you for the patch! At this point we're only supporting Python 3, do you think you could update the PR accordingly?

Yes sure, I can update the PR in the next few days.

tipabu:
I went ahead and took out the py2-isms and resolved the merge conflict; I trust no one minds.

Issue #810, opened 2023-09-20
Memory leak in greendns


jiajunsu:
Hi, we've found a memory leak in module `greendns`, when the dns server response `query` with exception `dns.exception.Timeout` or `dns.exception.DNSException`, we got a memory leak with `EAI_EAGAIN_ERROR` or `EAI_NODATA_ERROR`.

The root cause of this memory leak is from `TracebackType`:

1. In [greendns.py](https://github.com/eventlet/eventlet/blob/v0.33.3/eventlet/support/greendns.py#L88), we initialize `EAI_*_ERROR` as instances of class `socket.gaierror`, which has `__traceback__` attribute as [PEP-3134](https://peps.python.org/pep-3134/#traceback-attribute) defined. 
2. The `__traceback__` is a [`TracebackType`](https://docs.python.org/3/reference/datamodel.html#traceback-objects), when the same instance `EAI_NODATA_ERROR` being called multiple times, it inserts `tb_next` to the top of `EAI_NODATA_ERROR.__traceback__.tb_next`.

To simplify and reproduce the problem, we made a script, the main logic is same with function `greendns.resolve`.

```python
import socket
EAI_NONAME_ERROR = socket.gaierror(socket.EAI_NONAME, 'Name or service not known')

def query():
    raise EAI_NONAME_ERROR


def step():
    try:
        query()
    except Exception:
        pass


print("Init")
print(EAI_NONAME_ERROR.__traceback__)  # None

step()
print("After first raise")
print(EAI_NONAME_ERROR.__traceback__)  # <traceback object at 0x7f04dd6a3d90>
print(EAI_NONAME_ERROR.__traceback__.tb_next)  # <traceback object at 0x7f04dd6a1770>
print(EAI_NONAME_ERROR.__traceback__.tb_next.tb_next)  # None

step()
print("After second raise")
print(EAI_NONAME_ERROR.__traceback__)  # <traceback object at 0x7f04dd6a17c0>
print(EAI_NONAME_ERROR.__traceback__.tb_next)  # <traceback object at 0x7f04ddf48aa0>
print(EAI_NONAME_ERROR.__traceback__.tb_next.tb_next)  # <traceback object at 0x7f04dd6a3d90>
print(EAI_NONAME_ERROR.__traceback__.tb_next.tb_next.tb_next)  # <traceback object at 0x7f04dd6a1770>
print(EAI_NONAME_ERROR.__traceback__.tb_next.tb_next.tb_next.tb_next)  # None
[...]

PR #811, opened 2023-09-20
Fix memory leak in module greendns (alternative to #682)

jiajunsu:
Change `EAI_*_ERROR` to `lambda` expression, and initialize them
 when being called, to avoid memory leak with `__traceback__.tb_next`

fixes https://github.com/eventlet/eventlet/issues/810

### DIFF
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -92,11 +92,9 @@ EAI_NONAME_ERROR = socket.gaierror(socket.EAI_NONAME, 'Name or service not known
 # socket.EAI_NODATA is not defined on FreeBSD, probably on some other platforms too.
 # https://lists.freebsd.org/pipermail/freebsd-ports/2003-October/005757.html
 EAI_NODATA_ERROR = EAI_NONAME_ERROR
-EAI_ERRNOS = (socket.EAI_AGAIN, socket.EAI_NONAME)
 if (os.environ.get('EVENTLET_DEPRECATED_EAI_NODATA', '').lower() in ('1', 'y', 'yes')
         and hasattr(socket, 'EAI_NODATA')):
     EAI_NODATA_ERROR = socket.gaierror(socket.EAI_NODATA, 'No address associated with hostname')
-    EAI_ERRNOS = (socket.EAI_AGAIN, socket.EAI_NONAME, socket.EAI_NODATA)
 
 
 def _raise_new_error(error_instance):
@@ -520,7 +518,7 @@ def _getaddrinfo_lookup(host, family, flags):
                 try:
                     answer = resolve(host, qfamily, False, use_network=use_network)
                 except socket.gaierror as e:
-                    if e.errno not in EAI_ERRNOS:
+                    if e.errno not in (socket.EAI_AGAIN, EAI_NONAME_ERROR.errno, EAI_NODATA_ERROR.errno):
                         raise
                     err = e
                 else:
--- a/eventlet/support/greendns.py
+++ b/eventlet/support/greendns.py
@@ -99,7 +99,7 @@ if (os.environ.get('EVENTLET_DEPRECATED_EAI_NODATA', '').lower() in ('1', 'y', '
     EAI_ERRNOS = (socket.EAI_AGAIN, socket.EAI_NONAME, socket.EAI_NODATA)
 
 
-def raise_new_error(error_instance):
+def _raise_new_error(error_instance):
     raise error_instance.__class__(*error_instance.args)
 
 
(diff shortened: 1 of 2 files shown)


temoto (changes requested):
@jiajunsu thank you very much for attention to this problem.

I believe if we're going to break compat with external code by changing public variable type, it's better to do it loud. Would not want somebody to have `isinstance(err, eventlet.support.greendns.EAI_EAGAIN_ERROR)` always False because it's a function now.

Please do either:
- `_new_eagain_error = lambda ...`
- or `EAI_EAGAIN_MESSAGE = "Lookup timed out"; raise socket.gaierror(socket.EAI_AGAIN, EAI_EAGAIN_MESSAGE)`

jiajunsu:
> @jiajunsu thank you very much for attention to this problem.
> 
> I believe if we're going to break compat with external code by changing public variable type, it's better to do it loud. Would not want somebody to have `isinstance(err, eventlet.support.greendns.EAI_EAGAIN_ERROR)` always False because it's a function now.
> 
> Please do either:
> 
> * `_new_eagain_error = lambda ...`
> * or `EAI_EAGAIN_MESSAGE = "Lookup timed out"; raise socket.gaierror(socket.EAI_AGAIN, EAI_EAGAIN_MESSAGE)`

Thanks for the comment. I just tried another way to keep `EAI_*_ERROR` to be instances, adding a function `raise_new_error` for raising error.

jiajunsu:
@temoto I've passed the testcases of `tests/greendns_test.py` with py{37|38|39|311}, please have a review, thank you.

damani42:
Looks good to merge.

eventlet/support/greendns.py
```diff
@@ -85,15 +85,22 @@ def import_patched(module_name):
...
 EAI_NONAME_ERROR = socket.gaierror(socket.EAI_NONAME, 'Name or service not known')
 # EAI_NODATA was removed from RFC3493, it's now replaced with EAI_NONAME
 # socket.EAI_NODATA is not defined on FreeBSD, probably on some other platforms too.
 # https://lists.freebsd.org/pipermail/freebsd-ports/2003-October/005757.html
 EAI_NODATA_ERROR = EAI_NONAME_ERROR
+EAI_ERRNOS = (socket.EAI_AGAIN, socket.EAI_NONAME)
 if (os.environ.get('EVENTLET_DEPRECATED_EAI_NODATA', '').lower() in ('1', 'y', 'yes')
         and hasattr(socket, 'EAI_NODATA')):
     EAI_NODATA_ERROR = socket.gaierror(socket.EAI_NODATA, 'No address associated with hostname')
+    EAI_ERRNOS = (socket.EAI_AGAIN, socket.EAI_NONAME, socket.EAI_NODATA)
```
4383, line 99: Don't we want to store `errno` symbol, like previously made at line 516 (`(socket.EAI_AGAIN, EAI_NONAME_ERROR.errno, EAI_NODATA_ERROR.errno)`) (see diff below)?

tests/greendns_test.py
```diff
@@ -931,6 +931,23 @@ def test_noraise_dns_tcp(self):
             self.assertEqual(list(response.rrset.items)[0].address, expected_ip)
 
 
+class TestRaiseErrors(tests.LimitedTestCase):
+
+    def test_raise_new_error(self):
+        # https://github.com/eventlet/eventlet/issues/810
+        if not hasattr(greendns.EAI_EAGAIN_ERROR, "__traceback__"):
+            self.skipTest("Python version does not implement PEP-3134, skip this testcase.")
```
4383, line 939: We recently removed support of Python version lower than 3.8, so I think we can also start to remove this kind of version check.

Opinion?

eventlet/support/greendns.py
jiajunsu: Thanks, I just remove `EAI_ERRNOS`.

tests/greendns_test.py
jiajunsu: Yeah, if we only support python3.8+, this if expression will always be true, just remove it.

4383: We recently reintroduced support of Python 3.7 (https://github.com/eventlet/eventlet/issues/846), however, I don't think that will impact this code.

4383 (approved):
LGTM

4383:
Codecov is flaky...

4383:
@itamarst any preference between this PR and #682?

itamarst:
This one has a test, so may as well be this one.

itamarst (approved):

PR #817, opened 2023-10-31
Python 3.12 fixes by hroncok

nat-goodspeed:
Credit to @hroncok and @vstinner for logic (#797).
This PR attempts to add compatibility for older Python versions.

We draw the line at Python versions that predate `ssl.SSLContext`, though. The
remaining Python 2.7 documentation doesn't even mention the version at which
that was introduced.


eventlet/green/http/client.py
```diff
@@ -1450,6 +1450,38 @@ def getresponse(self):
 except ImportError:
     pass
 else:
+    try:
+        ssl.wrap_socket
```
temoto, line 1454: if hasattr?

nat-goodspeed: Will change.

eventlet/green/ssl.py
```diff
@@ -95,6 +100,34 @@ def __new__(cls, sock=None, keyfile=None, certfile=None,
             ret.__class__ = GreenSSLSocket
             return ret
 
+    global __ssl
+    try:
+        __ssl.wrap_socket
```
temoto, line 105: same hasattr

eventlet/green/ssl.py
```diff
@@ -436,58 +469,57 @@ def sslwrap_simple(sock, keyfile=None, certfile=None):
...
-            # on to SSLContext instances, but SSLContext.create_default_context
-            # does a bunch of work. Rather than re-implementing it all, just
-            # switch out the __class__ to get our `wrap_socket` implementation
-            context = _original_create_default_context(*a, **kw)
-            context.__class__ = GreenSSLContext
-            return context
-
-        create_default_context = green_create_default_context
-        _create_default_https_context = green_create_default_context
+class GreenSSLContext(_original_sslcontext):
```
temoto, line 472: definition coming out of condition is so cool, great!

tests/tpool_test.py
```diff
@@ -287,11 +287,11 @@ class TpoolLongTests(tests.LimitedTestCase):
     TEST_TIMEOUT = 60
 
     def test_a_buncha_stuff(self):
-        assert_ = self.assert_
+        assertTrue = self.assertTrue
```
temoto, line 290:
```suggestion
```

tests/tpool_test.py
```diff
@@ -287,11 +287,11 @@ class TpoolLongTests(tests.LimitedTestCase):
...
 
         class Dummy(object):
             def foo(self, when, token=None):
-                assert_(token is not None)
+                assertTrue(token is not None)
```
temoto, line 294:
```suggestion
                assert token is not None
```

tests/tpool_test.py
```diff
@@ -330,7 +330,7 @@ def test_leakage_from_tracebacks(self):
         first_created = middle_objs - initial_objs
         gc.collect()
         second_created = len(gc.get_objects()) - middle_objs
-        self.assert_(second_created - first_created < 10,
+        self.assertTrue(second_created - first_created < 10,
```
temoto, line 333:
```suggestion
        assert second_created - first_created < 10, "first loop: {}, second loop: {}".format(first_created, second_created)
```

tests/tpool_test.py
nat-goodspeed: Yes, okay.

vstinner:
FYI Fedora currently uses this patch for Python 3.12: https://src.fedoraproject.org/rpms/python-eventlet/blob/rawhide/f/python3.12.patch

temoto:
@vstinner thanks, i hope few small changes here would not create conflict issue in fedora?

vstinner:
> @vstinner thanks, i hope few small changes here would not create conflict issue in fedora?

If eventlet is ported to Python 3.12, we will just remove our patch in Fedora and update the package to the new eventlet release.

temoto:
> I'm sorry to say I don't know how to resolve the many failures of `verify_hub_empty()` or
> 
> ```
> ssl.SSLError: [SSL: KRB5_S_TKT_NYV] unexpected eof while reading (_ssl.c:2570)
> ```
> 
> I'm willing to work on a fix if I could have diagnostic help.

95% sure that is not related to this changeset.

KRB5_S_TKT_NYV seems like openssl compatibility, non empty hub is likely ref count (mem leak).

nat-goodspeed:
Have you seen similar failures on other branches?

nat-goodspeed:
I ran the tests locally on Python 3.10 against `master` and against this `python3.12` branch, and there was only one difference (other than timestamps, object IDs and PIDs).

    FAIL: test_024_expect_100_continue (tests.wsgi_test.TestHttpd)
      File "/Users/nat/linden/eventlet-nat/tests/wsgi_test.py", line 825, in test_024_expect_100_continue
        with self.assertRaises(socket.error) as caught:
    AssertionError: OSError not raised

Really, the question is: if this unblocks Python 3.12, is it worth merging?

jayofdoom:
Thanks for working on this.

A major downside to the approach taken here is the inability for library callers to configure the `SSLContext`. I wonder if there's a way to factor this where the application using the eventlet ssl module can offer some configuration -- or maybe even a preconfigured SSLContext -- in order to allow the application control over what TLS/SSL versions are supported, and certificates accepted.

jayofdoom:
I know @itamarst was going to take a look at py 3.12 support in eventlet as well. I've linked this PR to him.

itamarst:
I'm happy to leave this part to someone else, but I've been just trying to get tests to run on 3.12 and discovered nose3 is not compatible with 3.12. So I'm going to see if I can fix that. Will open separate issue to get feedback on approaches.

nat-goodspeed:
@jayofdoom 

> A major downside to the approach taken here is the inability for library callers to configure the SSLContext.

It looks as though [you can instantiate `GreenSSLSocket(..., _context=my_context)`](https://github.com/nat-goodspeed/eventlet/blob/python3.12/eventlet/green/ssl.py#L64-L74). Maybe that should be more prominently exposed? I'm open to suggestion on this point.

doko42:
python 3.12 support needs an update for the cpython 3.12 branch apparently, see
https://github.com/python/cpython/issues/112826

onlined:
Quoting from [Python documentation](https://docs.python.org/3.11/library/ssl.html#ssl.SSLContext.wrap_socket):

> Deprecated since version 3.7: Since Python 3.2 and 2.7.9, it is recommended to use the [SSLContext.wrap_socket()](https://docs.python.org/3.11/library/ssl.html#ssl.SSLContext.wrap_socket) instead of [wrap_socket()](https://docs.python.org/3.11/library/ssl.html#ssl.wrap_socket). The top-level function is limited and creates an insecure client socket without server name indication or hostname matching.

I think it's not needed anymore to use `ssl.wrap_socket` and do unnecessary branching because `eventlet` doesn't support versions older than 3.2 and 2.7. You can only use `SSLContext.wrap_socket`. Am I missing something?

4383:
Now that many CI fixes have been merged (thanks @itamarst), I think we can rerun checks. Normally, Py38 to Py311 jobs should be green.

itamarst:
I got everything passing. Further steps... since we're supporting 3.8+, and 3.8 already has `ssl.SSLContext`, possibly the `ssl.wrap_socket` codepath can just be dropped altogether.

itamarst:
Note there were some random irrelevant changes to test modules in the original PR.

eventlet/green/http/client.py
```diff
@@ -1450,35 +1450,21 @@ def getresponse(self):
 except ImportError:
     pass
 else:
-    if ssl._has_ssl_wrap_socket:
```
4383: hmm... I'm not sure to understand your intentions here.
Your commit message said that we only support py38 ("Simplify given we only support 3.8+") but you removed the sections related to py311 and earlier versions and you keep the code related to py312.

Should we not keep this condition to differentiate Py38-Py311 and Py312+.

Please can you give more details about your thinking.

itamarst: The original comment was wrong. It assumed the new API is new in 3.12, but it actually exists in 3.8 too.

4383: ack. thanks for details. Then LGTM

4383 (approved):
LGTM

As this a significant step for eventlet I'd suggest release a new version now, when this patch will be merged. 

@itamarst: Is it ok for you? If yes, do you want to proceed the new tag and the build/upload?

PR #820, opened 2023-11-01
tests: Drop unmaintained and unused stdlib tests

itamarst:
Eventlet had a setup that allowed running stdlib networking tests using eventlet.

However:

1. It is not running in GitHub.
2. It was never updated for Python 3. Random examples: references to `urllib2` (not in Python 3 at all), `test.test_support` (no longer exists in newer Python 3 versions).
3. The previous item means 13 years of stdlib test suite changes, so upgrading it won't be trivial.

Its current location also confuses alternatives to `nose`, so it's an impediment to #638.

The eventlet port to Python 3 was done more than 8 years ago (not sure about exact year), and this hasn't been updated since, and it's not being used, so deleting it seems reasonable. If anyone wants to revive it it will still be in git history.


4383 (approved):
LGTM and totally make sense

4383:
> The eventlet port to Python 3 was done more than 8 years ago (not sure about exact year), and this hasn't been updated since, and it's not being used, so deleting it seems reasonable. If anyone wants to revive it it will still be in git history.

Also, we should notice that [many releases of Python are officially supported in the same time](https://www.python.org/downloads/), so in this case, if we want to continue with this semantic, as each python release owns its own stdlib version, we should have as many stdlib tests folders as Python version supported by eventlet.

damani42:
Looks good to merge.

PR #825, opened 2023-12-12
Add entry points to allow to run examples simply by installing eventlet

4383:
These entry points allow to run some of the examples described in the documentation directly from the command line.

Eventlet users doesn't have to copy these examples to run them.

All entry points are suffixed with "example" to highlight their nature.

They can help to save time to run tests and checks locally.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,5 +49,15 @@ setuptools.setup(
         "Programming Language :: Python",
         "Topic :: Internet",
         "Topic :: Software Development :: Libraries :: Python Modules",
-    ]
+    ],
+    entry_points={
+        'console_scripts': [
+            'eventlet_chat_bridge_example = examples.chat_bridge:__main__',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -49,5 +49,15 @@
         "Programming Language :: Python",
         "Topic :: Internet",
         "Topic :: Software Development :: Libraries :: Python Modules",
-    ]
+    ],
+    entry_points={
+        'console_scripts': [
+            'eventlet_chat_bridge_example = examples.chat_bridge:__main__',
```
tipabu, line 55: Wait, installing eventlet also installs an importable `examples`?? That seems like a bug to me!

4383: Yeah that was more or less the idea, I used it locally to make tests and avoid creating patches based on copy of official examples. 

You are right, that's not the best idea that I had, let's abandon it.

It should have remained local hack :)

PR #826, opened 2023-12-12
Fix security issue related to RFC 9112

4383:
Reject requests which contains headers `content-length` and `transfer-encoding` at the same time.

That's not allowed by RFC 9112 and that could lead to potential security attacks.

https://www.rfc-editor.org/rfc/rfc9112#section-6.1-15

This is an extract of the RFC:

> A server MAY reject a request that contains both Content-Length and
> Transfer-Encoding or process such a request in accordance with the
> Transfer-Encoding alone. Regardless, the server MUST close the
> connection after responding to such a request to avoid the potential
> attacks.

> A server or client that receives an HTTP/1.0 message containing
> a Transfer-Encoding header field MUST treat the message as if the
> framing is faulty, even if a Content-Length is present, and close the
> connection after processing the message. The message sender might have
> retained a portion of the message, in buffer, that could be
> misinterpreted by further use of the connection.

The following request would lead to this scenario:

```
POST / HTTP/1.1
Host: a.com
Transfer-Encoding: chunked
Content-Length: 0
Content-Type: application/x-"##-form-urlencoded
14
id=1'or sleep(1);###
0
```

With these changes, when this kind of request is received the connection is closed and an error 400 is returned.

This scenario can be tested by using the following process:

1. run a wsgi server either: 
- by using the wsgi sample in official examples (http://eventlet.net/doc/examples.html#wsgi-server) 
- or by using my previous patch (https://github.com/eventlet/eventlet/pull/825) that allow you to run wsgi server from the command line (`$ eventlet_wsgi_example`)
2. send the following HTTP request to the running server:
```
curl -d "param1=value1&param2=value2" -X POST -H 'Transfer-Encoding:
chunked' -H 'Content-Length: 0' --http1.1 http://0.0.0.0:8090 -i
```

The previous curl command display returned headers and status code. You can observe that now, with these changes, bad requests are rejected.

 These changes also remove `content-lenght` from the `chunk` tests to avoid reflecting something that's not a bad practice.

This security issue was originally discovered by Keran Mu (mkr22@mails.tsinghua.edu.cn) and Jianjun Chen
(jianjun@tsinghua.edu.cn), from Tsinghua University and Zhongguancun Laboratory

Thanks to them for raising our attention about this security problem.


tests/wsgi_test.py
```diff
@@ -1919,6 +1919,22 @@ def test_close_idle_connections(self):
         except Exception:
             assert False, self.logfile.getvalue()
 
+    def test_rfc9112_reject_bad_request(self):
+        # (hberaud): Transfer-Encoding and Content-Length in the
+        # same headerare not allowed by rfc9112.
+        # Requests containing both headers MUST be rejected to
```
tipabu, line 1925: "MAY"

eventlet/wsgi.py
```diff
@@ -456,6 +457,14 @@ def handle_one_request(self):
                 self.close_connection = 1
                 return
 
+            if transfer_encoding is not None:
+                self.wfile.write(
+                    b"HTTP/1.0 400 Bad Request - rfc9112 Content-Length and "
+                    b"Transfer-Encoding are not allowed together\r\n"
```
tipabu, line 463: That seems like a long status line -- any reason not to do something more like
``` python
msg = b"Content-Length and Transfer-Encoding are not allowed together\n"
self.wfile.write(
    b"HTTP/1.0 400 Bad Request\r\n"
    b"Connection: close\r\n"
    b"Content-Length: %d\r\n"
    b"\r\n%s" % (len(msg), msg))
```
? Maybe even throw in a `Content-Type`?

tipabu (commented):
From the RFC:

> Early implementations of Transfer-Encoding would occasionally send both a chunked transfer coding for message framing and an estimated Content-Length header field for use by progress bars. This is why Transfer-Encoding is defined as overriding Content-Length, as opposed to them being mutually incompatible.

Given that, should we maybe make this in some way configurable, but defaulting to the more-secure behavior? That would allow us an escape hatch for working with old clients that cannot be updated (for example, because their source code may have been lost), particularly if the operator and/or application developer know that the service will not be forwarding any messages.

I'd also like to note that since we're relying on stdlib's HTTP parsing, there are still cases where a client may send both headers but we don't 400. See also: https://github.com/python/cpython/issues/81274 and https://github.com/eventlet/eventlet/pull/574 (which unfortunately I never properly followed up on).

That said, I agree, [the danger is certainly real](https://bugs.launchpad.net/swift/+bug/1840507).

4383:
> From the RFC:
> 
> > Early implementations of Transfer-Encoding would occasionally send both a chunked transfer coding for message framing and an estimated Content-Length header field for use by progress bars. This is why Transfer-Encoding is defined as overriding Content-Length, as opposed to them being mutually incompatible.
> 
> Given that, should we maybe make this in some way configurable, but defaulting to the more-secure behavior? That would allow us an escape hatch for working with old clients that cannot be updated (for example, because their source code may have been lost), particularly if the operator and/or application developer know that the service will not be forwarding any messages.

Totally agree with the "configurable but defaulting to the more secure behavior" path. I'll rework my patch to introduce an option.
 
> 
> I'd also like to note that since we're relying on stdlib's HTTP parsing, there are still cases where a client may send both headers but we don't 400. See also: [python/cpython#81274](https://github.com/python/cpython/issues/81274) and #574 (which unfortunately I never properly followed up on).

Thanks for the heads up.

> 
> That said, I agree, [the danger is certainly real](https://bugs.launchpad.net/swift/+bug/1840507).

eventlet/wsgi.py
4383: Good idea, you are right, the body of the request is more suited to forward error details. Will update.

tests/wsgi_test.py
4383: ack

eventlet/wsgi.py
4383: I have a subsidiary observation, personally I prefer using HTTP/1.0 according to the [`max_http_version` option documentation](https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L936C27-L936C101):

> Set to "HTTP/1.0" to make the server pretend it only supports HTTP 1.0. This can help with applications or clients that don't behave properly using HTTP 1.1.

However, during my development I seen that, in responses, the HTTP protocole is hardcoded and sometime the version is set to `HTTP 1.0` (the majority of the time) and sometime to `HTTP 1.1`:

- https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L746
- https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L28
- https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L433
- https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L439
- https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L454
- and so on...

Still, according to the previous doc sentence, I think that everything should be set to HTTP/1.0, right?

This could be fixed later through a follow up patch.
Just want to confirm the current case related to my patch.

tests/wsgi_test.py
```diff
@@ -2070,7 +2086,7 @@ def test_short_read_with_content_length(self):
     def test_short_read_with_zero_content_length(self):
         body = self.body()
         req = "POST /short-read HTTP/1.1\r\ntransfer-encoding: Chunked\r\n" \
-              "Content-Length:0\r\n\r\n" + body
```
4383, line 2073: Removed to do not reflect and encourage bad practices, I wonder if this test itself could be removed.
Another option, would be to remove the `transfer-encoding` and rename the test and keep it alive.

Any opinion?

tests/wsgi_test.py
```diff
@@ -2058,7 +2074,7 @@ def ping(self, fd):
     def test_short_read_with_content_length(self):
         body = self.body()
         req = "POST /short-read HTTP/1.1\r\ntransfer-encoding: Chunked\r\n" \
-              "Content-Length:1000\r\n\r\n" + body
```
4383, line 2061: Removed to do not reflect and encourage bad practices, I wonder if this test itself could be removed.
Another option, would be to remove the `transfer-encoding` and rename the test and keep it alive.

Any opinion?

tests/wsgi_test.py
```diff
@@ -2122,8 +2138,8 @@ def test_chunked_readline_from_input(self):
 
     def test_chunked_readlines_from_input(self):
         body = self.body()
-        req = "POST /readlines HTTP/1.1\r\nContent-Length: %s\r\n" \
```
4383, line 2125: Removed to do not reflect and encourage bad practices.

tests/wsgi_test.py
```diff
@@ -2112,8 +2128,8 @@ def test_chunked_readline(self):
 
     def test_chunked_readline_from_input(self):
         body = self.body()
-        req = "POST /readline HTTP/1.1\r\nContent-Length: %s\r\n" \
```
4383, line 2115: Removed to do not reflect and encourage bad practices.

tests/wsgi_test.py
```diff
@@ -2102,8 +2118,8 @@ def test_dirt(self):
 
     def test_chunked_readline(self):
         body = self.body()
-        req = "POST /lines HTTP/1.1\r\nContent-Length: %s\r\n" \
```
4383, line 2105: Removed to do not reflect and encourage bad practices.

4383:
codecov is a bit flaky...

eventlet/wsgi.py
```diff
@@ -340,11 +356,12 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
     # so before going back to unbuffered, remove any usage of `writelines`.
     wbufsize = 16 << 10
 
-    def __init__(self, conn_state, server):
+    def __init__(self, conn_state, server, reject_bad_requests=True):
```
4383, line 359: Is this level of configuration is enough for you?

itamarst (approved):

4383:
@tipabu is the latest version is ok for you?

itamarst:
If no answer in a few days should probably just proceed, given it's a security issue.

4383:
> If no answer in a few days should probably just proceed, given it's a security issue.

yes

PR #832, opened 2023-12-14
Green CI, part 2: pass tests on 3.10 and 3.11

### LINKED ISSUE #830: Green CI, part 2: Pass tests on 3.10 and 3.11
* Add 3.11 to GitHub Actions.
* Make sure tests pass.

itamarst:
Fixes #830 

I came up with a simpler fix for 3.10/3.11 support that continues to use the same RLock code in Python stdlib.

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -411,7 +411,7 @@ def _green_existing_locks():
             if not py3_style and isinstance(obj._RLock__block, lock_type):
                 _fix_py2_rlock(obj, tid)
             elif py3_style and not isinstance(obj, pyrlock_type):
-                _fix_py3_rlock(obj)
+                _fix_py3_rlock(obj, tid)
 
 
 def _fix_py2_rlock(rlock, tid):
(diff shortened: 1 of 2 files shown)


eventlet/patcher.py
```diff
@@ -424,17 +424,36 @@ def _fix_py2_rlock(rlock, tid):
         rlock._RLock__owner = tid
 
 
-def _fix_py3_rlock(old):
+def _fix_py3_rlock(old, tid):
     import gc
     import threading
+    from eventlet.green.thread import allocate_lock
     new = threading._PyRLock()
+    assert hasattr(new, "_block")
```
4383, line 432: We should not use `assert` outside of unit tests.

Indeed if someone his python interpreter with optimisation enabled (`python3.10 -o `), these statement would be removed.

Please consider raising an exception manually.

See the CPython doc:

https://github.com/python/cpython/blob/d9e1b5794a8fade21773d18f91a07f80b55c839c/Doc/using/cmdline.rst?plain=1#L315

4383 (changes requested):

eventlet/patcher.py
4383: `patcher.py` is module loaded at runtime.

jstasiak: > Indeed if someone use his python interpreter with optimisation enabled (python3.10 -o ), these statement would be removed.

I continue to be severely annoyed by that (C)Python design decision.

4383 (approved):

PR #836, opened 2023-12-14
Replace nose with pytest

### LINKED ISSUE #638: WIP: Remove dependency on nose

itamarst:
Fixes #638 

Unless nose3 starts being maintained again, this is a prerequisite for Python 3.12 support. nose3 is, in any case, essentially a dead project.

* [x] Everything passes in CI
* [x] Everything _runs_ in CI. nose runs 722 tests, pytest now runs 722 tests too.


4383:
I'm ok to move from nose to pytest, motivations (#638) looks legit.

itamarst:
This is fun. nose when running tests does 722 tests, when you run `nosetests --collect-only` it finds 723 :sob:

setup.py
```diff
@@ -31,7 +31,6 @@
             'README.rst'
         )
     ).read(),
-    test_suite='nose.collector',
```
itamarst, line 34: This was for `setup.py test`, which is now deprecated.

4383:
> This is fun. nose when running tests does 722 tests, when you run `nosetests --collect-only` it finds 723 

I'm just guessing... Maybe 'SkipTest' is just skipped at runtime or something like that, although it seems unlikely that we'll skip only a single test...

setup.py
4383: yes, no problem

4383 (approved):
God job! Thanks @itamarst

PR #827, opened 2023-12-13
Drop older Python versions

itamarst:
Fixes #819
Fixes #822 

Tests won't pass for now, but should at least run as before, just fewer builders.

PR #841, opened 2023-12-15
Stop claiming to create universal wheels

tipabu:
From https://wheel.readthedocs.io/en/stable/user_guide.html :

> If your project contains no C extensions and is expected to work on both Python 2 and 3, you will want to tell wheel to produce universal wheels by adding this to your `setup.cfg` file:
>
>     [bdist_wheel]
>     universal = 1

Now that we [no longer support Python 2](https://github.com/eventlet/eventlet/pull/827), it's inappropriate for us to claim that our wheels are universal. Note that claiming that a wheel is universal when it's not has caused trouble for other projects in the past; see https://github.com/PyCQA/bandit/issues/663

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,2 @@
 [metadata]
 description_file = README.rst
-
-[wheel]
-universal = True


tipabu:
Before:
```
(py3) $ python -m build .
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (setuptools >= 40.8.0, wheel)
* Getting build dependencies for sdist...
...
Successfully built eventlet-0.33.3.tar.gz and eventlet-0.33.3-py2.py3-none-any.whl
```
Note the `py2.py3` tag in the wheel name; this is despite the metadata properly reflecting the requirement on py38+:
```
(py3) $ unzip -p dist/eventlet-0.33.3-py2.py3-none-any.whl eventlet-0.33.3.dist-info/METADATA | grep Requires-Python
Requires-Python: >=3.8.0
```
After:
```
(py3) $ python -m build .
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (setuptools >= 40.8.0, wheel)
* Getting build dependencies for sdist...
...
Successfully built eventlet-0.33.3.tar.gz and eventlet-0.33.3-py3-none-any.whl
```
Also note that it looks like there have previously been issues around upload order to PyPI impacting the metadata recorded for sdists, so we'll probably want to keep an eye out for issues during/following the next release. See also: https://github.com/pypa/twine/issues/99

jstasiak (approved):

4383 (approved):
LGTM

Issue #843, opened 2023-12-15
[Process] Release management


4383:
Hey folks,

I think we need to setup a bit of process to decide how we want to manage releases.
Here is my proposal.

I think we need to designate an official release manager and his backup. Two people would be necessary. 

Everyone could keep access on pypi, but the goal here, is more to designate an identified person and avoid asking or requesting to everybody the same questions at each new release.

When the release manager think that a new release is needed, then he could create a new version by announcing it via a dedicated issue. Then the release could be created the next day to let's time to everyone to react.

If the release manager plan to be absent, then he could inform his backup and the release will continue to be managed during his absence.

Release manager and his backup would serve as contact for release topics with the outside world, more or less like the security contact role and security topics.

I'd suggest to avoid releases Friday afternoon, and in this case to wait early the next Monday in case If a patch is missing, or a last minute bug found.

I'd suggest to follow [semver](https://semver.org/) to choose the version number.

Release could be automatically created by pushing tags, and by using github actions coupled with OIDC on Pypi. With that anybody who push a tag will generate a new release. However, we need a bit of coordination to manage this tag creation, that's the goal of this proposal.

I don't want to create a bureaucracy, but I think we need a bit of official process if we want to avoid anoying situation and to know how to work together.

I'm used to manage releases, so, if no one else volunteer, I'd be happy to play the release manager role, else I'd be also happy to leave the role to someone else. 

Thoughts?

4383:
Our machinery is now a bit more automatized and modernized.

However, I keep this ticket opened to allow time to react and discuss this topic with other maintainers.

4383:
Without too much reactions, I think we close this issue now.

Issue #842, opened 2023-12-15
New eventlet release 0.34.1


4383:
As significant steps have been done I'd suggest releasing a new version of eventlet, 0.34.0.

- #822
- #817 
- #820 
- #827 
- #831 
- #832 
- #841

Why a minor version and not a bug fix version or a major version?

Because support of Python 2.7 have been removed and support of Python 3.12 have been added.

We can consider that backward compatibility is now broken, due to the removing of the support of Python 2.7, usually it would require a major release, but as eventlet is still not considered as stable, higher than version 1.0.0, I won't recommend a major version here.

Also a bug fix version won't reflect the support removal.

So a minor version remain our better option.

Are you ok with that?

If yes, does someone volunteer to create the related tag and run the mechanisms to upload to pypi?

itamarst:
Probably doing https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/ is better than manual upload, especially if long term goal is to focus on stable maintenance only, so the less work the better.

itamarst:
I can try to set that infrastructure up next week, I've done it a couple of times.

4383:
> Probably doing https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/ is better than manual upload, especially if long term goal is to focus on stable maintenance only, so the less work the better.

Yeah, that's I was thinking (OIDC), that's what I made on all my recent github project, including on aiohub:

https://github.com/4383/aiohub/blob/main/.github/workflows/main.yml#L52

We don't want to publish manually, but we want a bit of coordination for tag creation, that's the goal of my proposal here https://github.com/eventlet/eventlet/issues/843

4383:
> I can try to set that infrastructure up next week, I've done it a couple of times.

As you want, I can help you if you want. Let me know :)

4383:
I configured Pypi to use OIDC and I just proposed a patch to publish from github actions on tag creation

itamarst:
Also blocked on #846.

itamarst:
Having read the discussion in #754, at least part of that needs to make it in before release, specifically the change from `==` to `is`.

4383:
We also to do another round of updates against the changelog (NEWS) to reflect all the recent changes.

4383:
@itamarst: I think we are ready to go, please can you double check and confirm?

4383:
[...]

PR #845, opened 2023-12-18
Automatize and modernize deployments

4383:
These changes aim to avoid unnecessary workload by automatizing major parts of the release process.
    
With these changes, deployments would be automatic.
Pypi will accept new releases coming from eventlet/eventlet by adding a trusted publisher based on OIDC:
 - https://docs.pypi.org/trusted-publishers/
 - https://en.wikipedia.org/wiki/OpenID#OpenID_Connect_(OIDC)
    
These changes propose:
- Manage version number by using git tags rather than manual versioning (in python code)
- Modernize packaging by using pyproject.toml rather than setup.py
- Remove old artifacts related to publishing (`bin/release`)
- Adding a github workflow dedicated to pypi publish
- Added OIDC trusted publisher management https://pypi.org/manage/project/eventlet/settings/publishing/

These changes updated the list of authors to reflect current active core maintainers and allow community members    to contact the team.

    
Related to recent discussions:
- https://github.com/eventlet/eventlet/issues/843
- https://github.com/eventlet/eventlet/issues/842


damani42:
Looks good to merge.

4383:
Please, let's wait a bit before merging this patch.

I'd appreciate to get @itamarst's feedback too, as we discussed this topic together last Friday.

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+dependencies = []
+
+[project.urls]
+Homepage = "https://github.com/eventlet/eventlet"
+History = "https://github.com/eventlet/aiohub/blob/main/NEWS"
+Tracker = "https://github.com/eventlet/eventlet/issues"
+Source = "https://github.com/eventlet/aiohub"
+
+[project.optional-dependencies]
+dev = ["black", "isort", "pip-tools", "build", "twine", "pre-commit", "commitizen"]
```
4383, line 44: This is optional development dependencies.

If you are ok we that, once major issues will be solved, we could start using `pre-commit` and migrate to `black` formatting. Using `black` would avoid to loose time with formatting issues and pep8...

pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+requires = [
+  "hatch-vcs>=0.3",
+  "hatchling>=1.12.2",
+]
+build-backend = "hatchling.build"
+
+[project]
+name = "eventlet"
+authors = [
+    {name = "Liden Lab", email = "eventletdev@lists.secondlife.com"},
```
4383, line 11: Wonder if we want to update the author, I'm not sure this email address still work and I don't know if someone among us still own access to this inbox, any idea?

Maybe we could create a mailing list to allow user to contact us and then use the address of this mailing list there.
Also I'd suggest replacing the author name by the eventlet orga.

I don't want to hide the previous author, so, we could keep this line in the authors, and, then, just add another line before, to reflect the current maintainers information.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,38 @@
...
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+    - name: Build package
+      run: python -m build
+    - name: Publish package
+      # deploy only when a new tag is pushed to github
+      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
4383, line 37: publishing on pypi would be allowed only when we push a new tag.

4383:
the pypy jobs seems a bit flaky... the error doesn't seems legit, but, I'm not able to trigger an other run.

pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+    'greenlet >= 1.0',
+    'monotonic >= 1.4;python_version<"3.5"',
+    'six >= 1.10.0',
+]
+
+[project.urls]
+Homepage = "https://github.com/eventlet/eventlet"
+History = "https://github.com/eventlet/aiohub/blob/main/NEWS"
+Tracker = "https://github.com/eventlet/eventlet/issues"
+Source = "https://github.com/eventlet/aiohub"
```
itamarst, line 46: Wrong URL

pyproject.toml
```diff
@@ -0,0 +1,56 @@
...
+dependencies = [
+    'dnspython >= 1.15.0',
+    'greenlet >= 1.0',
+    'monotonic >= 1.4;python_version<"3.5"',
+    'six >= 1.10.0',
+]
+
+[project.urls]
+Homepage = "https://github.com/eventlet/eventlet"
+History = "https://github.com/eventlet/aiohub/blob/main/NEWS"
```
itamarst, line 44: Wrong URL

itamarst:
This all seems fine, other than:

* A few minor comments, see above.
* Deciding #846 maybe.

itamarst:
I guess merging this doesn't actually mean doing a release, and #836 is a blocker for a release, not release infrastructure. So please address and merge, thank you!

itamarst (approved):

4383:
done

4383:
> I guess merging this doesn't actually mean doing a release, and #836 is a blocker for a release, not release infrastructure. So please address and merge, thank you!

These changes won't release a new version, they are just the machinery.
We can merge this patch, manage #836, and then release something.

4383:
We are now ready to bring back Python 3.7 at life

eventlet/__init__.py
```diff
@@ -8,71 +8,65 @@
         DeprecationWarning,
     )
 
-version_info = (0, 33, 3)
-__version__ = '.'.join(map(str, version_info))
```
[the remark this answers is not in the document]
4383: Yes I removed it intentionally.
Now versions are managed by using code version control (git tags).

Version can be retrieved by using pip.

Someone else reported a similar issue yesterday evening, see the discussion there: https://github.com/eventlet/eventlet/issues/842#issuecomment-1864896273

I proposed to reintroduced something like:

```python
from importlib.metadata import version as getversion

__version__ = getversion("eventlet")
```

However, I'd love to understand why gunicorn is checking versions like that and not by using requirements file or dependencies mechanisms. Any idea?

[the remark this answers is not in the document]
jstasiak: I'm not gunicorn but speaking from my experience:

Sometimes you need to support multiple versions and modify your behavior based on the version you deal with.

Sometimes you want to import a package from a development repository tarball you just downloaded (where there's no metadata or Git tags present).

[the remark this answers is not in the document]
4383: Fix proposed https://github.com/eventlet/eventlet/pull/859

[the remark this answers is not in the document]
4383: > Sometimes you want to import a package from a development repository tarball you just downloaded (where there's no metadata or Git tags present).

That confirms my choice to not use `setuptools_scm `or `importlib.metadata.version`

PR #847, opened 2023-12-18
Fix Python 3.13 compat by adding missing attibute '_is_main_interpreter'

### LINKED ISSUE #604: missing attibute '_is_main_interpreter' (python 3.9 support)
Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1794283

Currently building with python3.9 prereleases works, but tests fail. 

```tests.greendns_test.test_reverse_name ... Program /builddir/build/BUILD/eventlet-0.25.1/tests/isolated/greendns_from_address_203.py output:
---
  File "/builddir/build/BUILD/eventlet-0.25.1/tests/isolated/greendns_from_address_203.py", line 6, in <module>
    eventlet.monkey_patch(all=True)
  File "/builddir/build/BUILD/eventlet-0.25.1/eventlet/patcher.py", line 285, in monkey_patch
    modules_to_patch += modules_function()
  File "/builddir/build/BUILD/eventlet-0.25.1/eventlet/patcher.py", line 435, in _green_thread_modules
    from eventlet.green import threading
  File "/builddir/build/BUILD/eventlet-0.25.1/eventlet/green/threading.py", line 22, in <module>
    eventlet.patcher.inject(
  File "/builddir/build/BUILD/eventlet-0.25.1/eventlet/patcher.py", line 100, in inject
    module = __import__(module_name, {}, {}, module_name.split('.')[:-1])
  File "/usr/lib64/python3.9/threading.py", line 37, in <module>
    _is_main_interpreter = _thread._is_main_interpreter
AttributeError: module 'eventlet.green.thread' has no attribute '_is_main_interpreter'
---
FAIL
```

This was apparently added in https://bugs.python.org/issue37266
(A bunch of the other tests also fail the same way, full list can be seen in: https://download.copr.fedorainfracloud.org/results/@python/python3.9/fedora-rawhide-x86_64/01309177-python-eventlet/build.log.gz

### LINKED ISSUE #838: missing attibute '_is_main_interpreter'
This is essentially a recurrence of #604, but affecting a recently released Python version: 3.12.1.  See https://github.com/python/cpython/issues/112826.  I merged a workaround in CPython for 3.12 (to preserve compatibility with 3.12.0), but for 3.13+ this will need to be fixed in eventlet.

Basically, eventlet/eventlet/green/thread.py needs something like the following added:

```py
_is_main_interpreter = __thread._is_main_interpreter
```

I'm happy to open a PR if that would help.

4383:
https://github.com/python/cpython/issues/112826

Python 3.13 doesn't have workaround and so eventlet is broken with Python versions higher than 3.12.

Fix #838
Fix #604

### DIFF
--- a/eventlet/green/thread.py
+++ b/eventlet/green/thread.py
@@ -15,6 +15,8 @@ error = __thread.error
 LockType = Lock
 __threadcount = 0
 
+if hasattr(__thread, "_is_main_interpreter"):
+    _is_main_interpreter = __thread._is_main_interpreter
 
 if six.PY3:
     def _set_sentinel():
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@ Unreleased
 ==========
 
 * Dropped support for Python 3.7 and earlier.
+* Fix Python 3.13 compat by adding missing attibute '_is_main_interpreter' https://github.com/eventlet/eventlet/pull/847
 
 0.33.3
 ======


eventlet/green/thread.py
```diff
@@ -15,6 +15,8 @@
 LockType = Lock
 __threadcount = 0
 
+if sys.version_info > (3, 12):
```
itamarst, line 18: Since some versions of 3.12 have this, maybe use a `hasattr()`? Alternatively, maybe catchall logic that just re-exports everything in the `__thread` module, just in case?

itamarst (approved):
Address comments as you see fit, thank you.

eventlet/green/thread.py
4383: Done.

You are right the `hasattr` approach is better.

The catchall logic could be implemented in a follow patch and then we would be able to remove this condition.
However, I won't spend time on it for now, so the current fix could be considered as a short term solution to be landed with the next eventlet release. We will have time early in 2024 to implement a more robust logic.

Thanks for your feedback.

PR #848, opened 2023-12-18
Add back 3.7 support

### LINKED ISSUE #846: Reconsider 3.7 support
@tipabu suggests at https://github.com/eventlet/eventlet/pull/827#issuecomment-1858614954 that maybe dropping 3.7 was a mistake; it's still about an eighth of downloads.

* Debian Buster is still getting security updates, and includes 3.7.
* Ubuntu 18.04 had 3.7 as an option. 18.04 was end of life in April, unless you pay for extended support, but I imagine some people are still using it.
* Probably there's a bunch of Docker images out there that hardcoded buster, e.g. `python:3.7-buster-slim` or similar.
* I think RHEL never had 3.7? RHEL 8 skipped it, if memory serves.

I automatically suggested dropping 3.7 on my default attitude as open source maintainer that it's been 4 years since Python 3.8 came out, and every relevant LTS distro (that I pay attention to) with 3.7 has had two major updates since.

But, that might have been the wrong perspective. From a harm reduction perspective it's possible that given the scope of eventlet usage, supporting 3.7 for a while yet is still worth it. `cryptography` is still supporting 3.7, for example.

If we do add back support, I would suggest any usage on 3.7 come with a very clear warning (or log message?) on startup.

itamarst:
Fixes #846 

1. Add back 3.7 support
2. Make 3.7 tests pass. I'm ... a little worried about this part but I did try to limit it so it mostly just impacts 3.7, except in one case where it was eventlet being incompatible with documented stdlib API.

### DIFF
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -24,6 +24,7 @@ __patched__ = [
 _original_sslsocket = __ssl.SSLSocket
 _original_sslcontext = __ssl.SSLContext
 _is_under_py_3_7 = sys.version_info < (3, 7)
+_is_py_3_7 = sys.version_info[:2] == (3, 7)
 _original_wrap_socket = __ssl.SSLContext.wrap_socket
 
 
@@ -191,6 +192,11 @@ class GreenSSLSocket(_original_sslsocket):
                                    write=True,
                                    timeout=self.gettimeout(),
                                    timeout_exc=timeout_exc('timed out'))
+                    elif _is_py_3_7 and "unexpected eof" in exc.args[1]:
+                        # For reasons I don't understand on 3.7 we get [ssl:
+                        # KRB5_S_TKT_NYV] unexpected eof while reading]
+                        # errors...
+                        raise IOClosed
                     else:
                         raise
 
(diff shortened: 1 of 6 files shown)


eventlet/green/ssl.py
```diff
@@ -191,6 +192,11 @@ def _call_trampolining(self, func, *a, **kw):
                                    write=True,
                                    timeout=self.gettimeout(),
                                    timeout_exc=timeout_exc('timed out'))
+                    elif _is_py_3_7 and "unexpected eof" in exc.args[1]:
```
itamarst, line 195: The error code is `1` in this case, which is Not Helpful.

tox.ini
```diff
@@ -30,7 +30,7 @@ setenv =
     {[testenv]setenv}
     eventlet_test_ipv6 = 1
 deps =
-    coverage==7.3.3
+    coverage
```
itamarst, line 33: Just use whichever version works with current Python version.

4383 (changes requested):
Please also remove https://github.com/eventlet/eventlet/blob/master/NEWS#L4

4383 (approved):
Changelog updated here https://github.com/eventlet/eventlet/pull/849

PR #852, opened 2023-12-19
rename github workflow to fix ignored action

4383:


.github/workflows/publish.yaml
```diff
@@ -6,17 +6,15 @@
 # separate terms of service, privacy policy, and support
 # documentation.
 
-name: Upload Python Package
+name: Build & Optional Deploy
 
-on:
-  release:
```
4383, line 12: We won't produce github releases, and we want to check if package building works during pull requests.

Also this is why this job was ignored during previous pull-requests and previous push.
Now you can observe that this job is triggered and you can observe that the deployment steps are ignored on pull requests.

.github/workflows/publish.yaml
```diff
@@ -6,17 +6,15 @@
...
-    types: [published]
+on: [push, pull_request]
 
 permissions:
   contents: read
 
 jobs:
-  deploy:
+  build-and-deploy:
```
4383, line 17: Better reflect the nature of this job.

.github/workflows/publish.yaml
```diff
@@ -6,17 +6,15 @@
 # separate terms of service, privacy policy, and support
 # documentation.
 
-name: Upload Python Package
+name: Build & Optional Deploy
```
4383, line 9: Better reflect the nature of this job

itamarst (approved):

PR #853, opened 2023-12-19
Drop old code based on python < 3.7, part 1: eventlet directory

eli-schwartz:
@jayofdoom successfully nerdsniped me, so I nerdsniped him back by teaching him about pyupgrade and ruff for automated migration. Here's a nice chunk of changes, which has the advantage of dropping a dependency on the "six" module.

Automatically upgrade code to drop python2 specific logic. Done in three passes:
- automatically run `pyupgrade --py3-only --keep-percent-format` on all files, to update to python 3.0 without changing percent formatting to `.format()` (this would be fairly noisy and isn't even 100% desired by everyone)
- manually fix up a bunch of six specific stuff, mainly dropping unused "import six" left behind by pyupgrade, and drop six from install requires.
- switch over to `ruff check` with its `--select UP*` class of rules targeting py37, and apply autofixes.

### DIFF
--- a/eventlet/debug.py
+++ b/eventlet/debug.py
@@ -36,7 +36,7 @@ class Spew:
                 try:
                     src = inspect.getsourcelines(frame)
                     line = src[lineno]
-                except IOError:
+                except OSError:
                     line = 'Unknown code named [%s].  VM instruction #%d' % (
                         frame.f_code.co_name, frame.f_lasti)
             if self.trace_names is None or name in self.trace_names:
--- a/eventlet/debug.py
+++ b/eventlet/debug.py
@@ -12,7 +12,7 @@ __all__ = ['spew', 'unspew', 'format_hub_listeners', 'format_hub_timers',
            'hub_prevent_multiple_readers', 'hub_timer_stacks',
            'hub_blocking_detection']
 
-_token_splitter = re.compile('\W+')
+_token_splitter = re.compile(r'\W+')
 
 
 class Spew:
--- a/eventlet/debug.py
+++ b/eventlet/debug.py
@@ -1,6 +1,5 @@
 """The debug module contains utilities and functions for better
 debugging Eventlet-powered applications."""
-from __future__ import print_function
 
 import os
 import sys
--- a/eventlet/debug.py
+++ b/eventlet/debug.py
@@ -16,7 +16,7 @@ __all__ = ['spew', 'unspew', 'format_hub_listeners', 'format_hub_timers',
 _token_splitter = re.compile('\W+')
 
 
-class Spew(object):
+class Spew:
 
     def __init__(self, trace_names=None, show_values=True):
         self.trace_names = trace_names
--- a/eventlet/green/http/__init__.py
+++ b/eventlet/green/http/__init__.py
@@ -52,8 +52,6 @@
 # 8. By copying, installing or otherwise using Python, Licensee
 # agrees to be bound by the terms and conditions of this License
 # Agreement.
-import six
-assert six.PY3, 'This is a Python 3 module'
 
 from enum import IntEnum
 
--- a/eventlet/green/http/client.py
+++ b/eventlet/green/http/client.py
@@ -1477,7 +1477,7 @@ else:
                      timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
                      source_address=None, *, context=None,
                      check_hostname=None):
-            super(HTTPSConnection, self).__init__(host, port, timeout,
+            super().__init__(host, port, timeout,
                                                   source_address)
             self.key_file = key_file
             self.cert_file = cert_file
--- a/eventlet/green/http/client.py
+++ b/eventlet/green/http/client.py
@@ -126,10 +126,7 @@ import email.parser
 import email.message
 import io
 import re
-try:
-    from collections.abc import Iterable
-except ImportError:
-    from collections import Iterable
+from collections.abc import Iterable
 from urllib.parse import urlsplit
 
 from eventlet.green import http, os, socket
--- a/eventlet/green/http/client.py
+++ b/eventlet/green/http/client.py
@@ -1126,7 +1126,7 @@ class HTTPConnection:
 
                 if encode_chunked and self._http_vsn == 11:
                     # chunked encoding
-                    chunk = '{0:X}\r\n'.format(len(chunk)).encode('ascii') + chunk + b'\r\n'
+                    chunk = '{:X}\r\n'.format(len(chunk)).encode('ascii') + chunk + b'\r\n'
                 self.send(chunk)
 
             if encode_chunked and self._http_vsn == 11:
@@ -1294,7 +1294,7 @@ class HTTPConnection:
         encode_chunked = kwds.pop('encode_chunked', False)
         if kwds:
             # mimic interpreter error for unrecognized keyword
-            raise TypeError("endheaders() got an unexpected keyword argument '{0}'"
+            raise TypeError("endheaders() got an unexpected keyword argument '{}'"
                             .format(kwds.popitem()[0]))
 
         if self.__state == _CS_REQ_STARTED:
@@ -1308,7 +1308,7 @@ class HTTPConnection:
         encode_chunked = kwds.pop('encode_chunked', False)
         if kwds:
             # mimic interpreter error for unrecognized keyword
-            raise TypeError("request() got an unexpected keyword argument '{0}'"
+            raise TypeError("request() got an unexpected keyword argument '{}'"
                             .format(kwds.popitem()[0]))
         self._send_request(method, url, body, headers, encode_chunked)
 
--- a/eventlet/green/ssl.py
+++ b/eventlet/green/ssl.py
@@ -59,7 +59,7 @@ class GreenSSLSocket(_original_sslsocket):
                 ssl_version=PROTOCOL_SSLv23, ca_certs=None,
                 do_handshake_on_connect=True, *args, **kw):
         if sys.version_info < (3, 7):
-            return super(GreenSSLSocket, cls).__new__(cls)
+            return super().__new__(cls)
         else:
             if not isinstance(sock, GreenSocket):
                 sock = GreenSocket(sock)
@@ -133,7 +133,7 @@ class GreenSSLSocket(_original_sslsocket):
         if sys.version_info < (3, 7):
             # nonblocking socket handshaking on connect got disabled so let's pretend it's disabled
             # even when it's on
-            super(GreenSSLSocket, self).__init__(
+            super().__init__(
                 sock.fd, keyfile, certfile, server_side, cert_reqs, ssl_version,
                 ca_certs, do_handshake_on_connect and six.PY2, *args, **kw)
         # the superclass initializer trashes the methods so we remove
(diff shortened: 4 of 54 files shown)


jstasiak:
If this is at least partially a result of some automated migration – can you share the exact commands you ran (and any other relevant context)?

It'll help in reviewing this and if/when someone needs to do something like this in the future (knowledge sharing).

I'd put it in the commit message when merging.

eli-schwartz:
> If this is at least partially a result of some automated migration – can you share the exact commands you ran (and any other relevant context)?



> I'd put it in the commit message when merging.

It's in the commit message :) that's one of my rules when I do this sort of thing.

jstasiak:
Oops, sorry – very nice (although I'm not sure about Markdown in there, outside of GH it's noise).

Let me swap "commit message" for "commit message and the PR description" then :)

eli-schwartz:
And to also share a bit of knowledge I told Jay in IRC...

The main advantage of pyupgrade over ruff check --fix is that ruff didn't implement fixers for migrating off of six (and pyupgrade's aren't complete). Other than that, and particularly for gating on versions starting with python 3.7, both tools are more or less equal in value in my experience other than ruff being *much* less opinionated because ruff allows you to pick and choose which fixer rules you want. Pyupgrade is kind of a pain for that (and also has a weird insistence on passing each filename by hand, because the author thinks that you should only be allowed to run tools via his "pre-commit.ci" tool).

4383:
Nice changes thanks.
I still have to finish to review them.

eventlet/debug.py
```diff
@@ -13,10 +12,10 @@
            'hub_prevent_multiple_readers', 'hub_timer_stacks',
            'hub_blocking_detection']
 
-_token_splitter = re.compile('\W+')
+_token_splitter = re.compile(r'\W+')
```
4383, line 15: raw string looks correct

eventlet/green/http/__init__.py
```diff
@@ -52,8 +52,6 @@
 # 8. By copying, installing or otherwise using Python, Licensee
 # agrees to be bound by the terms and conditions of this License
 # Agreement.
-import six
-assert six.PY3, 'This is a Python 3 module'
```
4383, line 56: `assert` must not be used in code loaded at runtime, if user call python with the optimization flag, all asserts would be removed, and, so, this kind of code would be removed at runtime.

That's one less to transform.

eventlet/green/http/client.py
```diff
@@ -1126,7 +1126,7 @@ def _send_output(self, message_body=None, encode_chunked=False):
 
                 if encode_chunked and self._http_vsn == 11:
                     # chunked encoding
-                    chunk = '{0:X}\r\n'.format(len(chunk)).encode('ascii') + chunk + b'\r\n'
+                    chunk = '{:X}\r\n'.format(len(chunk)).encode('ascii') + chunk + b'\r\n'
```
4383, line 1129: The current version LGTM, but, since we support versions of python higher than 3.7,  f-string could be used for this kind code.

eventlet/green/ssl.py
```diff
@@ -126,17 +125,12 @@ def __init__(self, sock, keyfile=None, certfile=None,
             sock = GreenSocket(sock)
         self.act_non_blocking = sock.act_non_blocking
 
-        if six.PY2:
-            # On Python 2 SSLSocket constructor queries the timeout, it'd break without
-            # this assignment
-            self._timeout = sock.gettimeout()
-
         if _is_under_py_3_7:
```
4383, line 128: This section (including the following lines), should be removed.

eventlet/green/ssl.py
```diff
@@ -126,17 +125,12 @@ def __init__(self, sock, keyfile=None, certfile=None,
...
             # nonblocking socket handshaking on connect got disabled so let's pretend it's disabled
             # even when it's on
-            super(GreenSSLSocket, self).__init__(
+            super().__init__(
                 sock.fd, keyfile, certfile, server_side, cert_reqs, ssl_version,
-                ca_certs, do_handshake_on_connect and six.PY2, *args, **kw)
+                ca_certs, False, *args, **kw)
```
4383, line 133: That's specific to Python versions lower than 3.7, let's drop that code.

4383 (changes requested):
I still have files to reviews, but some part of the current code should be updated, see my inline comments.

eventlet/green/subprocess.py
```diff
@@ -103,17 +101,14 @@ def wait(self, timeout=None, check_interval=0.01):
         # just want a version that uses eventlet.green.select.select()
         # instead of select.select().
         _communicate = FunctionType(
-            six.get_function_code(six.get_unbound_function(
-                subprocess_orig.Popen._communicate)),
+            subprocess_orig.Popen._communicate.__code__,
             globals())
         try:
             _communicate_with_select = FunctionType(
-                six.get_function_code(six.get_unbound_function(
```
4383, line 111: Not sure to see why these calls were duplicated... doesn't matter now that they are removed

eventlet/websocket.py
```diff
@@ -427,7 +426,7 @@ def _parse_messages(self):
                 buf = buf[end_idx + 1:]
             elif frame_type == 255:
                 # Closing handshake.
-                assert six.indexbytes(buf, 1) == 0, "Unexpected closing handshake: %r" % buf
+                assert buf[1] == 0, "Unexpected closing handshake: %r" % buf
```
4383, line 429: I would suggest taking advantage of the situation to replace this “assertion” with a condition. This is a module loaded at runtime, so if optimization is used this code will be removed.

Else it should be rework with a follow up patch.

eventlet/zipkin/http.py
```diff
@@ -12,45 +11,14 @@
...
-
-    def _patched_begin(self):
-        __org_begin__(self)
-
-        if api.is_tracing():
-            api.put_annotation('Client Recv (%s)' % self.status)
-
-
 def patch():
-    if six.PY2:
```
4383, line 39: I wonder if it would be more convenient to remove this entire module. I'm not sure, but the warning message below seems to suggest that zipkin is not supported by Python 3+.

Thoughts?

tests/green_http_test.py
```diff
@@ -1,8 +1,7 @@
 import eventlet
-import six
 import tests
 
-__test__ = six.PY3
+__test__ = True
```
4383, line 4: I think we could simply remove that line. Pytest would ignore test if `__test__ = False`, so I don't think this line is necessary.

It reflect a time where tests would had to be ignored with python 2.7

tests/isolated/patcher_open_kwargs.py
```diff
@@ -4,7 +4,7 @@
     import eventlet
     eventlet.monkey_patch(builtins=True, os=True)
 
-    with open(__file__, mode="rt", buffering=16):
```
4383, line 7: text mode is the default, LGTM

4383:
I now reviewed all the files, I've no other comments than those I already made, please have a look.
Thanks

4383:
Also, considering the heavy size of this patch, I'd love another double check from maintainers, I may missed some issues in my review.

itamarst:
I am pretty worried about this PR so please do NOT merge it until I have time (later today) to validate my concerns and write them up here.

eventlet/websocket.py
eli-schwartz: From a hasty look at the context, what you're saying makes sense.

Asserts are for cases where the matched condition indicates your own code is incorrectly written. They "cannot happen" because as everyone knows, we ourselves write bug-free code, so why check it.

This code  *appears* to be parsing external inputs, which means other people's code is buggy and we have to check that :D or well, innumerable things including fuzzer attacks.

I can fix it with a follow-up patch.

eventlet/zipkin/http.py
eli-schwartz: I wondered about that but lacked confidence to delete it outright since it might for example have a TODo intention to port it.

4383: I've no idea if someone else wanted to port it.
I'd personally suggest to remove it, but, maybe we want some deprecation period first, though, if I'm right, it won't work with python 3...

eventlet/websocket.py
4383: thanks

itamarst:
Here are my concerns, after reviewing the PR a bit more I am not sure that they are too big of an issue, but still worth considering.

### Problem: Eventlet is special

Eventlet is a weird library that _emulates_ the Python standard library, currently for 2.7 through 3.12. The auto-upgrade tools don't know about this scenario at all, they're assuming you're just a user of APIs.

Consider this code snippet, that was part of @jayofdoom's motivation, since it apparently caused breakage for someone: https://github.com/eventlet/eventlet/blob/4da815f51ba7afb1b476b8c0f539da1c8ba3e15a/eventlet/green/ssl.py#L140-L148

The auto-upgrade stuff doesn't touch it. There are likely other unconditional pieces of code only there to emulate Python 2 stdlib, or Python 3.5 stdlib, etc..

And by merging this PR, we take away the context that lets us figure out _why_ that code is doing what it does. In this case there's a comment saying it's for Python 2, but my worry was that there wouldn't always be.

In practice, I skimmed the PR and didn't see many other cases like this, but it's very long and I may have missed those...

### Problem: Code and tests in a single pass

Given how dynamic Python is, we need to rely heavily on tests for API stability. However, this PR modifies both code and tests in one go.

A better procedure would be to upgrade all the code, run the tests and make sure they pass, and _then_ upgrade the tests.

4383:
> Given how dynamic Python is, we need to rely heavily on tests for API stability. However, this PR modifies both code and tests in one go.
> 
> A better procedure would be to upgrade all the code, run the tests and make sure they pass, and _then_ upgrade the tests.

+1, also, that would reduce the number of files to review, and, so decrease the chances to miss something. Could be done by  simply unstagging changes temporarly from git and by committing both paths (`eventlet/` and `tests/`) independently.

eli-schwartz:
I agree it makes sense to split up the affected changes into two commits, one just for the tests. Small caveat: GitHub Actions won't test them, it only runs on the branch head. :P Would you prefer it in two PRs? Something else?

jayofdoom:
Two PRs is the easiest way to demonstrate it, for sure.

itamarst:
> I agree it makes sense to split up the affected changes into two commits, one just for the tests. Small caveat: GitHub Actions won't test them, it only runs on the branch head. :P Would you prefer it in two PRs? Something else?

The tests seem to run on PRs?

eli-schwartz:
> The tests seem to run on PRs?

It does not bisect a PR and check that tests pass at each step of the way. This isn't Gerrit, PRs cannot be treated as a series of independent patches in-series.

itamarst:
Oh, I see. Yeah, two PRs is the way to go.

eli-schwartz:
Okay, I separated them out and also spent some time breaking it up even more. Now I ran ruff per-fixer upgrades one at a time, and  used pyupgrade at the end -- mostly six-based stuff.

4383 (approved):
Latest patch set LGTM, waiting for @itamarst's feedback.

4383:
A couple of PEP8 issue need to be addressed.

itamarst (approved):

itamarst:
Just the style changes left, yeah.

eli-schwartz:
Style changes fixed up and absorbed into the relevant commits (which have been updated to note that they aren't a straight commit of the results anymore).

4383 (approved):
Latest patch set LGTM

eli-schwartz:
Thanks for merging.

Aside: in general I'd appreciate knowing in advance if a PR is going to be merged via squashing.

itamarst:
> Thanks for merging.
> 
> Aside: in general I'd appreciate knowing in advance if a PR is going to be merged via squashing.

Thank you for contributing, and apologies if lack of that knowledge caused you extra work.

eli-schwartz:
It's not about the extra work as much as it is, that if I'm aware in advance then there are some things I can do to partially mitigate the extremity level of the resulting damage.

Once the squash merge has happened, it is too late -- the damage is done, there is no going back.

PR #858, opened 2023-12-21
Drop old code based on python < 3.7, part 2: everything other than eventlet/

eli-schwartz:
Part 2 for #853, includes all commits from there.

Do not merge until that PR is merged and this one is rebased.


eli-schwartz:
It *should* just be a result of the GitHub workflow lacking facilities to have one pull request depend on a different pull request, yes.

tests/mock.py
```diff
@@ -2201,7 +2190,7 @@ def create_autospec(spec, spec_set=False, instance=False, _parent=None,
         else:
             parent = mock
             if isinstance(spec, FunctionTypes):
-                parent = mock.mock
```
itamarst, line 2204: I'm... curious why this change happened.

itamarst: I looked at stdlib, it matches old version, so I reverted this change.

itamarst (approved):

tests/mock.py
eli-schwartz: Pretty sure that `mock.mock` is just flat out incorrect in all cases. The expected use of the mock module is that `mock/__init__.py` imports all objects from `mock/mock.py` and you use them at the top level.

PR #859, opened 2023-12-21
Generate module version file at build

4383:
Gunicore rely on `eventlet.__version__` [1], however this data have been removed during our modernization of the continuous deployment mechanisms [2].

People reported problem with gunicore after 0.34.1 [3][4], so, it could be worth to reintroduce similar version info, to avoid side effects.

This patch propose to use a `hatch-vcs` hook [5] to generate dynamically, at build, the missing data. Other solutions exists but each of them have their own problems [6].

Indeed, considering "footgun" described in [6] I choose the hatch-vcs approach, because, retrieving a wrong version number during development when the lib is installed in editable mode, is not, I think, something horrible. I prefer this side effect rather than relying on another additional underlying library just to print a version number when eventlet is installed in editable mode. A new additional requirement which would be installed anytime at runtime and production.

Moreover, sometimes you want to import a package from a development repository tarball you just downloaded (where
there's no metadata or Git tags present). So, Using `setuptools_scm` or `importlib.metadata.version` won't works in that context.

Fix https://github.com/benoitc/gunicorn/issues/3120

[1] https://github.com/benoitc/gunicorn/issues/3120
[2] https://github.com/eventlet/eventlet/pull/845
[3] https://github.com/eventlet/eventlet/pull/845#discussion_r1433985555
[4] https://github.com/eventlet/eventlet/issues/842#issuecomment-1864896273
[5] https://github.com/ofek/hatch-vcs#build-hook
[6] https://github.com/maresb/hatch-vcs-footgun-example


eventlet/__init__.py
```diff
@@ -17,7 +11,19 @@
...
+# hatch have a build hook who generate version file, however,
+# if the project is installed in editable mode then the _version.py file
+# will not be updated unless the package is reinstalled (or locally rebuilt).
+# for further details, please read:
+# https://github.com/ofek/hatch-vcs#build-hook
+# https://github.com/maresb/hatch-vcs-footgun-example
+try:
+    from eventlet._version import __version__
+except ImportError:
+    pass
```
itamarst, line 24: Maybe set it to some default value in this case? e.g. "0.0" or "unknown".

eventlet/__init__.py
```diff
@@ -2,12 +2,6 @@
 import sys
 import warnings
 
-if sys.version_info < (3, 5):
```
itamarst, line 5: Maybe change this to (3, 8), instead of deleting?

pyproject.toml
```diff
@@ -59,3 +59,6 @@ exclude = ["tests*", "benchmarks", "examples"]
 
 [tool.hatch]
 version.source = "vcs"
+
+[tool.hatch.build.hooks.vcs]
```
itamarst, line 63: This is the wrong path, it should be "eventlet/_version.py".

itamarst (changes requested):
I tested `python -m build` and it doesn't actually include a `_version.py` file, probably because of the config issue noted in the review comments.

eventlet/__init__.py
4383: 3.7

pyproject.toml
4383: Updated and successfully tested locally. Good catch, thanks.

eventlet/__init__.py
4383: Done, upgraded to version 3.7. Thanks

eventlet/__init__.py
4383: Default to 0.0.0 but now that the path is fixed I don't think we will run the `except` part of this statement.

eventlet/__init__.py
```diff
@@ -2,12 +2,12 @@
...
+if sys.version_info < (3, 7):
     warnings.warn(
-        "Support for your Python version is deprecated and will be removed in the future",
+        "Your Python version is no longer supported by eventlet",
+        "Please consider upgrading to minimal supported Python version",
```
osfrickler, line 8: this raises an error for me, you'll have to make the warning a single str arg

eventlet/__init__.py
```diff
@@ -17,7 +17,19 @@
 from eventlet import semaphore
 from eventlet import support
 from eventlet import timeout
+# NOTE(hberaud): Version are now managed by hatch and control version
+# hatch have a build hook who generate version file, however,
+# if the project is installed in editable mode then the _version.py file
+# will not be updated unless the package is reinstalled (or locally rebuilt).
+# for further details, please read:
```
osfrickler, line 24: Suggestion to fix grammar and spelling:

```
# NOTE(hberaud): Versions are now managed by hatch and control version.
# hatch has a build hook which generates the version file, however,
# if the project is installed in editable mode then the _version.py file
# will not be updated unless the package is reinstalled (or locally rebuilt).
# For further details, please read:
```

eventlet/__init__.py
4383: Fixed.

```
>>> import eventlet
>>> eventlet.__version__
'0.34.2.dev1+gc2236b8.d20231222'
```

4383: Let me know if you still face the error.

eventlet/__init__.py
4383: Changed to your proposal.

eventlet/__init__.py
osfrickler: I must admit that originally I didn't actually test with python3.6, but just ran the warn() statement in py3.10, because I wanted to see how the multiline message would look like. With your original code, the issue would only be triggered when running with py3.6 or older. Now actually testing with py3.6, the error is gone, but also the deprecation warning actually isn't actually shown by default, since https://peps.python.org/pep-0565/ was only implemented for py3.7. So one would have to set something like `PYTHONWARNINGS=error` for the warning to actually take effect. This may just be the expected behavior and maybe I'm worrying too much about historic python versions, but maybe the alternative option of changing the warning to be a real error immediately might be a better solution?

4383: Thanks for your verification, and for your advices, they are really relevant.

Let's remove, from this pull request, the changes related to this warning. We can handle it in a separated pull request to more isolate the topics. I prefer to keep this pull request more focused on the sole topic of python version. I should have treated them separately.

4383: And anyway, this warning will be removed by https://github.com/eventlet/eventlet/pull/853/files
So, I suggest we do not waste time on it anywhere else.

itamarst (approved):
Thank you!

itamarst:
I added a `.gitignore` entry for `_version.py`. Looks good to merge.

Issue #432, opened 2017-08-18
Can't log from within tpool.execute


smerritt:
If you try to log from within a function called by `tpool.execute`, there is a chance that the tpool thread never returns, and you see a stack trace like this one:

```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/hub.py", line 458, in fire_timers
    timer()
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/timer.py", line 58, in __call__
    cb(*args, **kw)
  File "/usr/local/lib/python2.7/dist-packages/eventlet/semaphore.py", line 147, in _do_acquire
    waiter.switch()
error: cannot switch to a different thread
```

This is because many logging handlers have mutexes that are `threading._RLock` objects, and Eventlet's replacement for `thread.allocate_lock` returns an `eventlet.semaphore.Semaphore` object, which does not work across different hubs in different pthreads.

Here's a small script to reproduce the issue:
```
#!/usr/bin/env python
#
# Demonstrates the crash with logging across pthreads
import eventlet.patcher
import eventlet.tpool
import logging
import random
import sys
import time

eventlet.patcher.monkey_patch()

logger = logging.getLogger("logger-test")
# This handler's .lock is a threading._RLock
logger.addHandler(logging.StreamHandler(sys.stdout))
logger.setLevel(logging.DEBUG)


def log_n_times(me, n):
    for x in range(n):
        logger.info("%s %d", me, x)
        time.sleep(random.random() * 0.01)


logger.info("starting")

greenthread = eventlet.spawn(log_n_times, 'greenthread', 50)
eventlet.tpool.execute(log_n_times, 'pthread', 50)
greenthread.wait()

logger.info("done")
```

The bug was originally found in Openstack Swift, and there's a better writeup at https://bugs.launchpad.net/swift/+bug/1710328 and a commit at https://github.com/openstack/swift/commit/6d160797fc3257942618a7914d526911ebbda328 that fixes the problem, but only for Swift, and not in a general way.

I'd like to figure out how to fix this in general, but I'm not sure how to proceed. Using a pipe-based mutex for all `_RLock` objects would work, but would be a very expensive fix. Perhaps just the locks in logging handlers, because those are global?

temoto:
@smerritt thank you for this sad information.

[...]

PR #870, opened 2024-01-03
An asyncio hub for eventlet

### LINKED ISSUE #869: An asyncio hub for eventlet
See #868 for motivation.

A minimal implementation would just have this as an optional hub. 

My guess is that having the asyncio hub as the default one, at least on Unix-y platforms, would be useful, will see once I have it working. Probably best done as follow-up issue after some amount of real-world usage of the hub has found any initial bugs.

itamarst:
Fixes #869 

I gave up on multi-reader support, I can't even figure out how it works and it seems like a terrible idea (though people are using it, based on quick GitHub search).

### DIFF
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -4,6 +4,7 @@ Asyncio-based hub, originally implemented by Miguel Grinberg.
 
 import asyncio
 import os
+import sys
 
 from eventlet.hubs import hub
 
@@ -38,7 +39,12 @@ class Hub(hub.BaseHub):
         else:
             self.loop.remove_writer(fileno)
 
-        cb(fileno)
+        try:
+            cb(fileno)
+        except self.SYSTEM_EXCEPTIONS:
+            raise
+        except:
+            self.squelch_exception(fileno, sys.exc_info())
         self.sleep_event.set()
 
     def add(self, evtype, fileno, cb, tb, mark_as_closed):
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -17,10 +17,6 @@ def is_available():
 class Hub(hub.BaseHub):
     """An Eventlet hub implementation on top of an asyncio event loop."""
 
-    @staticmethod
-    def is_available():
-        return True
-
     def __init__(self):
         super().__init__()
         self.sleep_event = asyncio.Event()
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -33,12 +33,7 @@ class Hub(hub.BaseHub):
         super().add_timer(timer)
         self.sleep_event.set()
 
-    def _file_cb(self, evtype, cb, fileno):
-        if evtype == hub.READ:
-            self.loop.remove_reader(fileno)
-        else:
-            self.loop.remove_writer(fileno)
-
+    def _file_cb(self, cb, fileno):
         try:
             cb(fileno)
         except self.SYSTEM_EXCEPTIONS:
@@ -56,9 +51,9 @@ class Hub(hub.BaseHub):
         listener = super().add(evtype, fileno, cb, tb, mark_as_closed)
         if not already_listening:
             if evtype == hub.READ:
-                self.loop.add_reader(fileno, self._file_cb, evtype, cb, fileno)
+                self.loop.add_reader(fileno, self._file_cb, cb, fileno)
             else:
-                self.loop.add_writer(fileno, self._file_cb, evtype, cb, fileno)
+                self.loop.add_writer(fileno, self._file_cb, cb, fileno)
         return listener
 
     def remove(self, listener):
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -10,7 +10,12 @@ from eventlet.hubs import hub
 
 
 def is_available():
-    """Python 3 always has asyncio."""
+    """
+    Indicate whether this hub is available, since some hubs are
+    platform-specific.
+
+    Python always has asyncio, so this is always ``True``.
+    """
     return True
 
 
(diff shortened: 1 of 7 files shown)


tipabu:
FWIW, I've worked with code that needed to allow multiple readers. The long and short of it was needing to have a lock that would play nice with both greenthreads and real threads to work around https://github.com/eventlet/eventlet/issues/432; more details are available in https://github.com/openstack/swift/commit/6d160797fc3257942618a7914d526911ebbda328 / [LP #1710328](https://bugs.launchpad.net/swift/+bug/1710328)

Unfortunately, it *also* worked across forks, which seemed unnecessary and likely to impact performance. Eventually I managed to convince myself it wasn't necessary since we were always using UDP/Unix domain sockets for logging, so we wound up dropping the logging lock in https://github.com/openstack/swift/commit/e09d4bcf4250602b78eef8fee8cdeeef8894c931

I don't really have anything to add beyond that; I'll try to give this a spin in the near future, though. I just wanted to add some color to

> I gave up on multi-reader support, I can't even figure out how it works and it seems like a terrible idea (though people are using it, based on quick GitHub search).

itamarst:
If someone can make it work, great, I just gave up because how it works at all is an utter mystery to me (notice `eventlet.hubs.hub.Hub.secondaries` is never accessed by anything!). Given the goal here is for people to port to asyncio, and I suspect this isn't an asyncio feature anyway, so porting would have to drop such usages anyway. And it sounds like in practice Swift doesn't need it anymore?

eventlet/hubs/asyncio.py
```diff
@@ -0,0 +1,120 @@
...
+def is_available():
+    """Python 3 always has asyncio."""
+    return True
+
+
+class Hub(hub.BaseHub):
+    """An Eventlet hub implementation on top of an asyncio event loop."""
+
+    @staticmethod
+    def is_available():
```
4383, line 21: Same thing here. I'm not sure what kind of availability is expected here, but I think this is related to python releases, so if I'm right, this part could be also removed.

tests/greenio_test.py
```diff
@@ -856,6 +857,10 @@ class TestGreenIoLong(tests.LimitedTestCase):
     TEST_TIMEOUT = 10  # the test here might take a while depending on the OS
 
     def test_multiple_readers(self):
+        from eventlet.hubs.asyncio import Hub
+        if isinstance(get_hub(), Hub):
```
4383, line 861: Can't we modify this condition to catch the `RuntimeError` raised at line 119 in `eventlet.hubs/asyncio.py` and then act accordingly rather than checking the type of the instance?

That would allow to assert that the mechanism implemented into the hub are well triggered in this kind of condition.

4383:
Docstrings should be documented. If we want to encourage migration to eventlet and, hence, hope to see this feature adopted, then we need to provide a well documented hub.

Do you already plan to add docstring or should we add it to our todo list, and, add doc through a follow up patch?

Concerning gates, they seems happy, even when the asyncio hub is enabled. That's encouraging!

eventlet/hubs/asyncio.py
```diff
@@ -0,0 +1,120 @@
...
+import asyncio
+import os
+import sys
+
+from eventlet.hubs import hub
+
+
+def is_available():
+    """Python 3 always has asyncio."""
+    return True
```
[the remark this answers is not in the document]
itamarst: No, the Hub API requires this, if you don't have this it emits an warning. See https://github.com/eventlet/eventlet/blob/master/eventlet/hubs/__init__.py#L81-L88

eventlet/hubs/asyncio.py
itamarst: This can probably be removed.

tests/greenio_test.py
itamarst: Mm, maybe. Ideally we'd figure out how to do multiple readers in asyncio.

itamarst:
(I have implemented the start of next step, running asyncio in combination with eventlet, just to see if it actually works, and ... it does! I have a test with aiohttp talking to eventlet's WSGI server. So feeling good about this approach.)

tests/greenio_test.py
itamarst: Oh also the hub is already started at this point I think? So it wouldn't actually raise a `RuntimeError` in a place that the test can actually deal with.

This also raises the issue though that one can set the flag after the hub is running, so the `RuntimeError` check is not sufficient...

4383:
> > Docstrings should be documented. If we want to encourage migration to eventlet and, hence, hope to see this feature adopted, then we need to provide a well documented hub.
> 
> The public APIs are the `eventlet.hubs.Hub` interface, so in general that documentation is the relevant one. Users won't be using this class directly much, possibly not at all.

Indeed.
> 
> Will see if there's anything that does need it though.

WFM.

> 
> > Concerning gates, they seems happy, even when the asyncio hub is enabled. That's encouraging!
> 
> Gates?

jobs/tests triggered by github actions

tests/greenio_test.py
itamarst: OK came up with a better approach.

itamarst:
Filed #874 for multiple reader support.

4383 (approved):
As far as I can see, these changes LGTM.

That could be useful, during the squash, to indicate in the commit message, that this patch is inspired from https://gist.github.com/miguelgrinberg/829a20792d7283ae27b1f6a390c378b9 . We shouldn't skimp on the history.

4383:
> That could be useful, during the squash, to indicate in the commit message, that this patch is inspired from https://gist.github.com/miguelgrinberg/829a20792d7283ae27b1f6a390c378b9 . We shouldn't skimp on the history.

Oh just seen you already put a comment, awesome!

4383:
I let you rewrite the commit message and drop useless info.
I let you push the merge button.

PR #877, opened 2024-01-11
Support using asyncio coroutines from inside greenlets

### LINKED ISSUE #873: Support calling asyncio-using `async def` functions from inside greenlets
This will allow eventlet-based code to use asyncio-based libraries, which gives a migration path off of eventlet for eventlet-only libraries.

This will only work if the asyncio hub is being used.

itamarst:
Fixes #873

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -72,6 +72,7 @@ deps =
     py{310,311}: psycopg2-binary==2.9.5
     py{310,311}: pyzmq==25.0.0
     dnspython1: dnspython<2
+    asyncio: aiohttp
 usedevelop = True
 commands =
     pip install -e .
--- a/eventlet/asyncio.py
+++ b/eventlet/asyncio.py
@@ -2,7 +2,10 @@
 Asyncio compatibility functions.
 """
 import asyncio
-from .greenthread import spawn
+
+from greenlet import GreenletExit
+
+from .greenthread import spawn, getcurrent
 from .event import Event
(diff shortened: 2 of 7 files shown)


tox.ini
```diff
@@ -72,6 +72,7 @@ deps =
     py{310,311}: psycopg2-binary==2.9.5
     py{310,311}: pyzmq==25.0.0
     dnspython1: dnspython<2
+    asyncio: aiohttp
```
4383, line 75: is aiohttp is only required by tests? I suppose that yes.

4383 (commented):
@itamarst: Amazing work!

I just have 2 inline comments, but globally everything LGTM.

itamarst:
Need to check if this affected by #408.

tox.ini
itamarst: Yes, it's just a testing thing.

itamarst:
Looks like #408 is only an issue in Python 3.10 or earlier, so not going to worry about it.

itamarst:
@4383 if you're happy can you formally do the approval so I can merge? Thanks!

4383 (approved):
LGTM

4383:
I let you push the button and rewrite the commit message if you think that's necessary.

itamarst:
I made one final change, renaming the function to `spawn_for_awaitable()` since that's a more accurate description of the API.

Issue #879, opened 2024-01-12
Fix doc generation and automatize publishing process


4383:
I recently generated a new doc version (https://github.com/eventlet/eventlet/commit/c10bd90d7250b4cee213346d0672b3354eb767a2), but the static website seems still stuck to version 0.33.3 (https://eventlet.net/doc/index.html).

We should also notice that actually the docs is manually generated and manually published to github pages, it could be worth automatizing this process each time a new tag/release is made.

4383:
Generating the docs failed due to many errors in docstrings, here are some examples:

```
/home/hberaud/dev/perso/eventlet/eventlet/hubs/__init__.py:docstring of eventlet.hubs.get_default_hub:10: ERROR: Undefined substitution referenced: "internal".                                                                                
/home/hberaud/dev/perso/eventlet/eventlet/hubs/__init__.py:docstring of eventlet.hubs.get_hub:3: ERROR: Undefined substitution referenced: "internal".
/home/hberaud/dev/perso/eventlet/eventlet/hubs/__init__.py:docstring of eventlet.hubs.trampoline:13: ERROR: Undefined substitution referenced: "internal".
/home/hberaud/dev/perso/eventlet/eventlet/support/greendns.py:docstring of eventlet.support.greendns.ResolverProxy.query:7: ERROR: Unexpected indentation.
```

I'll fix these errors

4383:
My previous comment is not anymore relevant.
Github pages are not designed to host restructred text and manually compiled files.
Github pages are poorly integrated with github actions and automatize publication on gh-pages require too many works compared to the native integration provided by readthedocs.

I prefer focus on using readthedocs to see our documentation automatically updated.
Readthedocs allow to host docs in parallel for each new tags that we produce and also for the master branch. That's not with github pages.

readthedocs generate a new version each time a push is made on github (commits, tags). Workflow will be better automatized by using this solution.

4383:
I think this topic is almost done, let's close that issue.

PR #880, opened 2024-01-12
Modernize doc generation

4383:
Related to #879 

1. Use tox to build docs:
- Move files into doc/source
- Create a new testenv related to docs.
- Manage docs requirements into a dedicated file
- Remove obsolete makefile and build artifacts

2. Run docs generation on pull requests, too many docstring errors have been accepted, leading us to broken docs generation.

3. Fix a couple of error related to docs.

4. Setup the eventlet repo to host the doc on readthedocs rather than on github pages. Docs are automatically generated each time we will push changes (commits, tags), no needs to run doc generation manually and to push it to github-pages.


.readthedocs.yaml
```diff
@@ -0,0 +1,32 @@
+# .readthedocs.yaml
```
itamarst, line 1: We should link to the new URL for docs in the README.

.readthedocs.yaml
```diff
@@ -0,0 +1,32 @@
...
+# Read the Docs configuration file
```
itamarst, line 2: Do we have a way to update the old docs to point to the new URL? Otherwise Google will point people at old docs and they won't know new ones exist.

itamarst (approved):
Beyond previously mentioned stuff, there's:

* A broken link in the README, "open a new issue".
* I can't actually see that ReadTheDocs publishing works since I don't know what the URL is (or maybe it's just not public at the obvious URL?)

4383:
> Beyond previously mentioned stuff, there's:
> 
>     * A broken link in the README, "open a new issue".
> 
>     * I can't actually see that ReadTheDocs publishing works since I don't know what the URL is (or maybe it's just not public at the obvious URL?)

The docs is not yet online on readthedocs as the readthedocs config file is mandatory to allow building. The snake bites its tail. The url will be https://eventlet.readthedocs.io/ and http://eventlet.rtfd.io/

.readthedocs.yaml
4383: I'll do another patch to update the links (in the pyproject.toml file too) once the docs will online on readthedocs.

.readthedocs.yaml
4383: readthedocs allow custom domains. We could simply redirect the `eventlet.org` domain to the new readthedocs hosted instance. Every users would be automatically redirected.

Also, if necessary, we could publish an empty html index on the github page containing the following javascript instruction to redirect to the new docs hosted on readthedocs:

`window.location.href = ‘https://eventlet.readthedocs.io/’;`

4383:
>     * A broken link in the README, "open a new issue".

Fixed by https://github.com/eventlet/eventlet/commit/131651af900376256740bb4a332789e25a522551

PR #881, opened 2024-01-13
Bugfix/837

### LINKED ISSUE #837: Revive doctests (or kill them permanently)
At some point these existed, but they're not run in CI, so will need some more work to get them working again.

Alternatively, decide not to support them and remove mention from the documentation.

Do this after #638 is fixed because pytest adds more complications.

radeksm:
Hi! 
   While started looking on the doctest made these two changes 

1.  4f56a9be78b46acd981f4533529de7391894f5bd fixes probably the only doctest problem
2.  19d1b55b0503b6c9986e28b44053571310ad3b76 is just a tip of the iceberg :) 

Partially fixes #837


4383 (approved):
Cool, thank you

PR #883, opened 2024-01-16
Fix bad exceptions handlings

4383:
This patch fix 2 conditions that are badly designed. They can't work. In all case the exception will
be catched by the first `except` and then the second one will be all the time ignored.

Here is a tiny PoC to demonstrate the bad design:

```python
import errno

def trigger(err, message):
    print(f"try to simulate {message}: {err}")
    raise OSError(err, message)


def catcher(err, message):
    try:
        trigger(err, message)
    except OSError as e:
        print("I'm here")
        if e.errno != errno.EAGAIN:
            raise
    except OSError as e:
        print("I'm there")
        if e.errno == errno.EPIPE:
            return ''
        raise


if __name__ == "__main__":
    try:
        catcher(errno.EPIPE, "pipe!")
    except Exception as e:
        print(e)
    try:
        catcher(errno.EAGAIN, "again!")
    except Exception as e:
        print(e)
```

The previous code will output:

```
try to simulate pipe!: 32                                                      
I'm here                                                                       
[Errno 32] pipe!                                                                                                                                               
try to simulate again!: 11     
I'm here   
```

### DIFF
--- a/eventlet/green/os.py
+++ b/eventlet/green/os.py
@@ -41,12 +41,10 @@ def read(fd, n):
         try:
             return __original_read__(fd, n)
         except OSError as e:
-            if get_errno(e) != errno.EAGAIN:
-                raise
-        except OSError as e:
             if get_errno(e) == errno.EPIPE:
                 return ''
-            raise
+            if get_errno(e) != errno.EAGAIN:
+                raise
         try:
             hubs.trampoline(fd, read=True)
         except hubs.IOClosed:
@@ -65,10 +63,7 @@ def write(fd, st):
         try:
             return __original_write__(fd, st)
         except OSError as e:
-            if get_errno(e) != errno.EAGAIN:
-                raise
-        except OSError as e:
-            if get_errno(e) != errno.EPIPE:
+            if get_errno(e) not in [errno.EAGAIN, errno.EPIPE]:
                 raise
         hubs.trampoline(fd, write=True)
 


itamarst (approved):
Good catch! I wonder if this is the reason for the some of the weird TLS shutdown edge cases I've seen.

4383:
> Good catch! I wonder if this is the reason for the some of the weird TLS shutdown edge cases I've seen.

No idea, my discover is pure serendipity...

JerryFuGuan:
This commit changes the previous behavior when an EPIPE exception occurs.

4383:
> This commit changes the previous behavior when an EPIPE exception occurs.

Well, the way these exceptions were designed originally (before this patch), made them not able to work as expected, so the previous behavior was simply to do nothing when EPIPE exception occurs...

JerryFuGuan:
> > This commit changes the previous behavior when an EPIPE exception occurs.
> 
> Well, the way these exceptions were designed originally (before this patch), made them not able to work as expected, so the previous behavior was simply to do nothing when EPIPE exception occurs...

When subprocess.Popen.compunicate is used in the service code, the code enters an infinite loop in the os.write function and the program is suspended. Previously, we would go to the first exception handling, raise, and then exit the function, because get_errno(e)==errno.EPIPE != errno.EAGAIN. Or return an empty string like the current os.read function can also exit the function.

4383:
Would you like to propose a PR to fix it?

PR #886, opened 2024-01-16
Asyncio hub support for Python 3.7 to 3.9

itamarst:


eventlet/hubs/asyncio.py
```diff
@@ -24,11 +24,11 @@ class Hub(hub.BaseHub):
 
     def __init__(self):
         super().__init__()
-        self.sleep_event = asyncio.Event()
         # The presumption is that eventlet is driving the event loop, so we
         # want a new one we control.
         self.loop = asyncio.new_event_loop()
         asyncio.set_event_loop(self.loop)
+        self.sleep_event = asyncio.Event()
```
4383, line 31: Why we need to move this line?

4383:
All the new jobs looks successful

eventlet/hubs/asyncio.py
itamarst: In older Python versions the Event() would've been created with the wrong loop if it ran before `set_event_loop()`, and then it complained about mismatching loops.

4383: ack

4383 (approved):
LGTM

PR #889, opened 2024-01-16
Support awaiting GreenThread in an `async def` context

itamarst:
* [x] Basic awaiting works
* [x] Cancelling the Future kills the GreenThread


4383 (approved):
Changes LGTM, good job!

PR #894, opened 2024-01-16
Add very minimal linting

itamarst:
Doesn't lint `eventlet/green/` and has a very restricted set of lints, but better than nothing.


eventlet/backdoor.py
```diff
@@ -2,8 +2,6 @@
 import errno
 import socket
 import sys
-import errno
```
4383, line 5: errno is used many times in this module, we can't remove it. For an example, please, see line 37.

eventlet/wsgi.py
```diff
@@ -474,7 +474,7 @@ def handle_one_request(self):
                 return
 
             if transfer_encoding is not None:
-                if reject_bad_requests:
+                if self.reject_bad_requests:
```
4383, line 477: good catch

eventlet/zipkin/api.py
```diff
@@ -155,10 +155,11 @@ def build_span(name, trace_id, span_id, parent_id,
 
     @staticmethod
     def build_annotation(value, endpoint=None):
-        if isinstance(value, unicode):
+        if isinstance(value, str):
             value = value.encode('utf-8')
+        assert isinstance(value, bytes)
```
4383, line 160: please avoid using `assert` in runtime code. Could be replaced by something like:

```python
if not isinstance(value, bytes):
    raise RuntimeError("Only bytes are accepted...")
```

4383: Also, why adding this check?
`ttypes.Annotation` expect a string: https://github.com/eventlet/eventlet/blob/master/eventlet/zipkin/_thrift/zipkinCore/ttypes.py#L143

eventlet/zipkin/api.py
```diff
@@ -155,10 +155,11 @@ def build_span(name, trace_id, span_id, parent_id,
...
         return ttypes.Annotation(time.time() * 1000 * 1000,
-                                 str(value), endpoint)
```
4383, line 161: I don't think we should remove that cast.
ttypes.Annotation expect a string: https://github.com/eventlet/eventlet/blob/master/eventlet/zipkin/_thrift/zipkinCore/ttypes.py#L143 and your previous check want to force for type bytes.

tests/greenpool_test.py
```diff
@@ -7,11 +7,6 @@
 import tests
 
 
-def passthru(a):
```
4383, line 10: this function is used below, see line 29.

tests/ssl_test.py
```diff
@@ -1,6 +1,4 @@
-import contextlib
 import random
-import socket
```
4383, line 3: As `eventlet.green.socket` is imported below I think it is safe to remove this one.

4383 (changes requested):
Thanks for this patch, I think some parts needs changes, see my inline comments.

eventlet/backdoor.py
itamarst: It's still imported, see line 2 right above it. The issue was it was being imported twice.

eventlet/zipkin/api.py
itamarst: That's UTF-8 encoded string, i.e. bytes. AFAICT this code was never ported to Python 3, so when it says 'str()` in the original code it actually meant `bytes`. Notice I'm deleting a usage of `unicode()`... AFAICT from other uses of the code it expects bytes.

This is, to be fair, a guess though. This code is probably completely busted honestly given it wasn't ported to Python 3.

eventlet/zipkin/api.py
itamarst: See above.

tests/greenpool_test.py
itamarst: There were two `passthru` functions, there's another one defined later on.

eventlet/zipkin/api.py
4383: ok thanks for your explanation.

eventlet/backdoor.py
4383: Oh sorry, I missed the previous import.

eventlet/zipkin/api.py
4383: thanks

tests/greenpool_test.py
4383: lol :)

As the other is defined later, I suppose that this is this other one which is loaded on memory at runtime.

4383 (approved):

PR #895, opened 2024-01-17
install eventlet at readthedocs build

4383:
Else the build fails due to an eventlet import in the doc config [1].

[1] https://github.com/eventlet/eventlet/blob/master/doc/source/conf.py#L59


4383:
Example of fails https://readthedocs.org/projects/eventlet/builds/23135483/

itamarst (approved):

Issue #897, opened 2024-01-18
[release] eventlet 0.35.0


4383:
Will release a couple of new features and bugfixes:

* 3b4c683e (origin/master, origin/HEAD) fix link in readme (#902)
* ee76d8b5 update changelog for version 0.35.0 (#898)
* 4f0a90b8 Move docs to readthedocs (#899)
* b2142b79 [doc] fix a missing word (#901)
* 387ca4ba Basic documentation for asyncio migration (#892)
* c36286f1 Add very minimal linting (#894)
* 8d781c2e install eventlet at readthedocs build (#895)
* 6ee471d8 fix truncate size nullable (#789)
* 8f5ebf13 Handle transport endpoint shutdown in conditions (#884)
* 881dd62f Rework reject_bad_requests option
* 3f1242b1 Fix NameError from #826
* 963a1c17 Support awaiting GreenThread in an `async def` context (#889)
* e33c6dec Extend test cert to 2049 (#643)
* 7a804272 Asyncio hub support for Python 3.7 to 3.9 (#886)
* 7fa158ca fix bad directory name in readthedocs config file (#888)
* b9c72161 Modernize doc generation (#880)
* 81d43592 Fix bad exceptions handlings (#883)
* 131651af Reword maintenance policy sentences (#882)
* 3f8c1dad Support using asyncio coroutines from inside greenlets (#877)
* 4cf31905 Bugfix/837 (#881)
* 7662654f Update maintenance goals (#850)
* df81d5d4 An asyncio hub for eventlet (#870)

Would help with https://review.opendev.org/c/openstack/swift/+/905796

4383:
Tag: https://github.com/eventlet/eventlet/releases/tag/v0.35.0
Pypi: https://pypi.org/project/eventlet/0.35.0/

PR #898, opened 2024-01-18
update changelog for version 0.35.0

4383:
Related to #897

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,26 @@
 Unreleased
 ==========
 
+0.35.0
+======
+
+* [doc] Basic documentation for asyncio migration https://github.com/eventlet/eventlet/pull/892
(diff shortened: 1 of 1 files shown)


itamarst (approved):

4383:
@itamarst I'd suggest to add https://github.com/eventlet/eventlet/pull/899
Hence 0.35.0 would have its own docs page

4383:
> @itamarst I'd suggest to add #899 Hence 0.35.0 would have its own docs page

No docs pages are modified by #899, so would only impact pypi description and github home page, though.
But could worth adding it before pushing a new version to pypi, hence, the modified links would be propagated with the last release (0.35.0)

itamarst:
That seems fine, yeah.

PR #899, opened 2024-01-18
Move docs to readthedocs

4383:
Official docs is now hosted on the readthedocs https://eventlet.readthedocs.io/en/latest/.
Only branches or tags containing the previous readthedocs setup are able to been displayed.

For now only the master branch contains the right setup.

Historically, the docs was hosted on gh-pages. This previous setup only host the last built version.
No parallel versions of the docs were possible.

This new way of documenting eventlet allow us to parallelize versions. 
That will be better to reflect deprecations, removal, or additions.

Now remain a bit of work:
- make other maintainers readthedocs admin
- transfer eventlet.net to readthedocs
- disable gh-pages


4383:
Worth merging it before releasing 0.35.0, as we could benefit from this patch to update pypi's docs links with the latest published version (0.35.0)

itamarst (approved):

PR #905, opened 2024-01-26
[Fix] shrinks window before connecting

### LINKED ISSUE #821: tests fail with kernel 6.6
Downstream report: https://bugzilla.suse.com/show_bug.cgi?id=1216858
Kernel report: https://lore.kernel.org/all/738bb6a1-4e99-4113-9345-48eea11e2108@kernel.org/

tl;dr
https://github.com/torvalds/linux/commit/dfa2f0483360d4d6f2324405464c9f281156bd87 causes https://github.com/eventlet/eventlet/blob/dd2f0ea32d6b6d064be52fe8e27fca275985cdc4/tests/greenio_test.py#L389
to fail due to timeout -- the test takes now 1.7 s instead of the previous 300 ms. 

So far the result from the above is:

> It seems the test had some expectations.
> 
> Setting a small (1 byte) RCVBUF/SNDBUF, and yet expecting to send
> 46080 bytes fast enough was not reasonable.
> It might have relied on the fact that tcp sendmsg() can cook large GSO
> packets, even if sk->sk_sndbuf is small.
> 
> With tight memory settings, it is possible TCP has to resort on RTO
> timers (200ms by default) to recover from dropped packets.

and

> Retracting TCP windows has always been problematic.
> 
> If we really want to be very gentle, this could add more logic,
> shorter timer events for pathological cases like that,
> I am not sure this is really worth it, especially if dealing with one
> million TCP sockets in this state.

Michal proposes:

> How about my suggestion to set SO_SNDBUF and SO_RCVBUF before listen() and
> connect()? Would that be an acceptable solution or does the test depend on
> setting the buffer sizes this late?

4383:
Close #821

For further details please see:
- https://lore.kernel.org/all/20231103092706.6rw2ehuigxfdvhlc@lion.mk-sys.cz/
- https://bugzilla.suse.com/show_bug.cgi?id=1216858
- https://github.com/eventlet/eventlet/issues/821

### DIFF
--- a/tests/greenio_test.py
+++ b/tests/greenio_test.py
@@ -417,8 +417,8 @@ class TestGreenSocket(tests.LimitedTestCase):
 
         server_evt = eventlet.spawn(server)
         client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        client.connect(('127.0.0.1', listener.getsockname()[1]))
         bufsized(client)
+        client.connect(('127.0.0.1', listener.getsockname()[1]))
         large_evt = eventlet.spawn(read_large, client)
         eventlet.sleep(0)
         client.sendall(b'hello world')


4383:
@jirislaby & @mkubecek: Please, let me know if that patch looks ok for you and if it fix your issue. Thanks.

jirislaby:
@4383 Note that the original bug was worked around by https://github.com/eventlet/eventlet/commit/5f39d559d23b633fa1c0b374f6d895b0333fe317 in the meantime. If I revert that (I see the failure again) and apply this, it works too.

jirislaby:
And also the time is from:
```
real    0m1.442s
user    0m0.168s
sys     0m0.037s
```
to
```
real    0m0.214s
user    0m0.153s
sys     0m0.061s
```

4383:
@jirislaby: Cool. Thanks for your feedbacks

itamarst:
I am on kernel 6.6.6. `master` runs the test in 1.3 seconds, the PR runs the test in 0.1 seconds.

PR #907, opened 2024-01-30
Do not allow failed patching to stop execution

jayofdoom:
There are some cases where monkey_patching cannot always be performed early enough; such as sphinx autodoc importing. To help handle these cases, exceptions during patching are logged and execution is allowed to continue.

This fixed an issue in OpenStack Manila docs generation caused by an upgrade to eventlet 0.35.0.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,11 @@
 Unreleased
 ==========
 
+0.35.1
+======
+
+* [fix] Do not allow failed patching to stop execution https://github.com/eventlet/eventlet/pull/907
+
 0.35.0
 ======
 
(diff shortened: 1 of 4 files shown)


jayofdoom:
Before:
image

After:
image

Example from https://zuul.opendev.org/t/openstack/build/63914f9dc9e24e1180eea22bfd92e13c but screenshot posted since that link will bitrot :)

4383 (approved):
LGTM

jayofdoom:
Pushed something that should make the coverage job a little happier :)

4383 (approved):

PR #910, opened 2024-02-02
redirect eventlet.net to readthedoc

4383:
Remove outdated documentation


index.html
```diff
@@ -1,162 +1,10 @@
...
-<li><a class="reference external" target="_blank" href="irc://chat.freenode.net/#eventlet">IRC channel</a></li>
-<li><a class="reference external" target="_blank" href="https://www.openhub.net/p/eventlet">Open Hub stats</a></li>
-</ul>
-        </div>
-      </div>
-    </div>
-    <script src="//yandex.st/highlightjs/7.3/highlight.min.js"></script>
-    <script>hljs.initHighlightingOnLoad();</script>
-  </body>
+<html>
```
itamarst, line 2: Maybe also some text with a link, for people with JavaScript disabled?

itamarst (approved):
Feel free to merge with or without fix.

index.html
itamarst: Although I guess this is temporary until DNS gets updated so not a permanent solution anyway.

4383: Will submit a follow up patch.

PR #921, opened 2024-02-20
Fix tool.setuptools/packages list

thomasgoirand:
In distros, when building the package and attempting to install it, there may be a "debian" folder. This creates this issue:

error: Multiple top-level packages discovered in a flat-layout: ['debian', 'eventlet'].

which is fixed by this patch. Please merge it to make it easier for package maintainers.


4383 (approved):
Building machinery seems happy with that, let's merge this patch

PR #912, opened 2024-02-05
Fix wsgi.server shutdown for in-flight requests

tipabu:
Cloning from #590 (which still has some relevant discussion) -- the source repo got deleted, which prevents me from resolving the conflicts on the original PR. Original description (mostly) follows:

---

While doing some work on OpenStack Swift, I noticed that in-flight
requests were getting their sockets closed by eventlet when the listen
socket was closed. In that case, there is a server process which bound
the listen socket and N fork'ed off worker processes which all run the
wsgi.server loop calling accept() and handling requests.

I noticed that nothing was ever setting connection state to
STATE_REQUEST (i.e. unused constant in previous patch). Upon closer
inspection, it turns out that the socket cleanup code for STATE_IDLE
sockets got applied to the socket with an in-progress request. This
caused code using the socket to get an EPIPE and exit the request's
greenthread without completing the request.

With some instrumentation in the code, without this patch's fix, the new
test observed the code doing this:

```
(7544) wsgi starting up on http://127.0.0.1:36745/
(7544) calling accept
(7544) accept got <eventlet.greenio.base.GreenSocket object at 0x7f7650d73290>, ('127.0.0.1', 50194)
(7544) accepted ('127.0.0.1', 50194)
(7544) completed req ('127.0.0.1', 50194)
(7544) calling accept
(7544) got exception 22
(7544) re-raising!
(7544) wsgi exiting, calling pool.waitall()
Traceback (most recent call last):
File "/tmp/swiftstack-eventlet/eventlet/wsgi.py", line 602, in handle_one_response
write(b'')
File "/tmp/swiftstack-eventlet/eventlet/wsgi.py", line 541, in write
wfile.flush()
File "/usr/lib64/python2.7/socket.py", line 303, in flush
self._sock.sendall(view[write_offset:write_offset+buffer_size])
File "/tmp/swiftstack-eventlet/eventlet/greenio/base.py", line 403, in sendall
tail = self.send(data, flags)
File "/tmp/swiftstack-eventlet/eventlet/greenio/base.py", line 397, in send
return self._send_loop(self.fd.send, data, flags)
File "/tmp/swiftstack-eventlet/eventlet/greenio/base.py", line 384, in _send_loop
return send_method(data, *args)
error: [Errno 32] Broken pipe

127.0.0.1 - - [31/Oct/2019 01:08:05] "PUT /foo-bar HTTP/1.1" 200 0 0.000696
(7544) wsgi exited, is_accepting=True
```

[...]

PR #924, opened 2024-03-01
[WIP] Implement file descriptor removal and waiting in asyncio hub

4383:
We recently observed (#912) difference in the behaviors between the asyncio hub and the other eventlet hubs.

The asyncio hub behaved differenly because bad file descriptor were not properly removed from the list of readers and writers listeners.

To clean these bad file descriptors we have to implement an awaitable wait method specific to the asyncio hub

This patch should fix #912.

I prefer to propose this patch as a separated patch to isolate topics and simplify reviews.

https://github.com/eventlet/eventlet/pull/912

### DIFF
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -3,10 +3,19 @@ Asyncio-based hub, originally implemented by Miguel Grinberg.
 """
 
 import asyncio
+import errno
 import os
 import sys
 
+from eventlet import patcher, support
 from eventlet.hubs import hub
+select = patcher.original('select')
+
+
+try:
+    BAD_SOCK = {errno.EBADF, errno.WSAENOTSOCK}
+except AttributeError:
+    BAD_SOCK = {errno.EBADF}
 
 
 def is_available():
@@ -29,6 +38,18 @@ class Hub(hub.BaseHub):
         self.loop = asyncio.new_event_loop()
         asyncio.set_event_loop(self.loop)
         self.sleep_event = asyncio.Event()
+
+    def _remove_bad_fds(self):
+        """ Iterate through fds, removing the ones that are bad per the
+        operating system.
+        """
+        all_fds = list(self.listeners[self.READ]) + list(self.listeners[self.WRITE])
+        for fd in all_fds:
+            try:
+                select.select([fd], [], [], 0)
+            except OSError as e:
+                if support.get_errno(e) in BAD_SOCK:
+                    self.remove_descriptor(fd)
 
     def add_timer(self, timer):
         """
@@ -134,7 +155,7 @@ class Hub(hub.BaseHub):
                             pass
                         self.sleep_event.clear()
                     else:
-                        await asyncio.sleep(0)
+                        await self.wait(0)
                 else:
                     self.timers_canceled = 0
                     del self.timers[:]
@@ -144,3 +165,37 @@ class Hub(hub.BaseHub):
                 self.stopping = False
 
         self.loop.run_until_complete(async_run())
+
+    async def wait(self, seconds=None):
+        readers = self.listeners[self.READ]
+        writers = self.listeners[self.WRITE]
+        if not readers and not writers:
+            if seconds:
+                await asyncio.sleep(seconds)
+            return
+        reader_fds = list(readers)
+        writer_fds = list(writers)
+        all_fds = reader_fds + writer_fds
+        try:
+            r, w, er = select.select(reader_fds, writer_fds, all_fds, seconds)
+        except OSError as e:
+            if support.get_errno(e) == errno.EINTR:
+                return
+            elif support.get_errno(e) in BAD_SOCK:
+                self._remove_bad_fds()
+                return
+            else:
+                raise
+
+        for fileno in er:
+            readers.get(fileno, hub.noop).cb(fileno)
+            writers.get(fileno, hub.noop).cb(fileno)
+
+        for listeners, events in ((readers, r), (writers, w)):
+            for fileno in events:
+                try:
+                    listeners.get(fileno, hub.noop).cb(fileno)
+                except self.SYSTEM_EXCEPTIONS:
+                    raise
+                except:
+                    self.squelch_exception(fileno, sys.exc_info())


4383:
I need to address the job errors. Will have a look soon.
Do not hesitate to share your feeling/suggestions.

4383:
The test in error (`test_cancelling_future_kills_greenthread`) only fails if it is launched between other tests.

If I run this failing test in isolated way, it is successful each time I run it...

```
.tox/py312-asyncio/bin/py.test \
tests/asyncio_test.py::test_cancelling_future_kills_greenthread
============================================================================================================= test session starts =============================================================================================================
platform linux -- Python 3.12.2, pytest-8.0.2, pluggy-1.4.0
rootdir: /home/dev/app
plugins: cov-4.1.0
collected 1 item                                                                                                                                                                                                                              

tests/asyncio_test.py .                                                                                                                                                                                                                 [100%]

============================================================================================================== 1 passed in 0.21s ==============================================================================================================
```

4383:
Again I observes two different behaviors depending on if I run the test (`test_cancelling_future_kills_greenthread``) alone [1] or not [2].

If I run the problematic test between other tests, the test do not seems to run the `go` inner method, and, so the CancelledError thrown by the greenthread is not properly catched up, and then the test is broken...

  File "/home/dev/app/eventlet/event.py", line 173, in _do_send
    waiter.switch(result)
  File "/home/dev/app/eventlet/greenthread.py", line 267, in main
    self._resolve_links()
  File "/home/dev/app/eventlet/greenthread.py", line 283, in _resolve_links
    f(self, *ca, **ckw)
  File "/home/dev/app/eventlet/asyncio.py", line 42, in _got_result
    gthread.wait()
  File "/home/dev/app/eventlet/greenthread.py", line 224, in wait
    return self._exit_event.wait()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/app/eventlet/event.py", line 131, in wait
    current.throw(*self._exc)
  File "/home/dev/app/eventlet/greenthread.py", line 264, in main
    result = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/app/eventlet/asyncio.py", line 54, in _run
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/dev/app/tests/asyncio_test.py", line 236, in go
    async def go():
asyncio.exceptions.CancelledError

https://github.com/eventlet/eventlet/blob/0a130e2077ae7f5fd4f250b0bf8a9037000fc6bd/tests/asyncio_test.py#L236

https://github.com/eventlet/eventlet/blob/0a130e2077ae7f5fd4f250b0bf8a9037000fc6bd/tests/asyncio_test.py#L229

This strange behavior seems introduced by the changes I proposed in this pull request.

[1] `.tox/py312-asyncio/bin/py.test tests/asyncio_test.py::test_cancelling_future_kills_greenthread`
[2] `.tox/py312-asyncio/bin/py.test tests/asyncio_test.py`

4383:
Here is a reproducer by entering a PDB post mortem on the broken test.
We can observe that the go task have been canceled. However we can also observe that a task remains pending. The pending task is related to the clock used by hubs (default to monotonic).

https://github.com/eventlet/eventlet/blob/0a130e2077ae7f5fd4f250b0bf8a9037000fc6bd/eventlet/hubs/hub.py#L124C12-L126

And which is called in the asyncio hub module:

https://github.com/eventlet/eventlet/blob/b13a262eb2f12a021870a31f56f09af2fa0ad99d/eventlet/hubs/asyncio.py#L141

```
$ .tox/py312-asyncio/bin/py.test --pdb  tests/asyncio_test.py
...
        phases = []
        def green():
            phases.append(1)
            future.cancel()
            eventlet.sleep(1)
            # This should never be reached:
            phases.append(2)
    
        gthread = eventlet.spawn(green)
    
        async def go():
            try:
                await gthread
            except asyncio.CancelledError:
                return "good"
            else:
                return "bad"
    
        future = asyncio.ensure_future(go())
>       assert spawn_for_awaitable(future).wait() == "good"

tests/asyncio_test.py:244: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
eventlet/greenthread.py:224: in wait
    return self._exit_event.wait()
eventlet/event.py:124: in wait
    result = hub.switch()
eventlet/hubs/hub.py:310: in switch
    return self.greenlet.switch()
eventlet/greenthread.py:264: in main
    result = function(*args, **kwargs)
eventlet/asyncio.py:56: in _run
    return future.result()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   async def go():
E   asyncio.exceptions.CancelledError

tests/asyncio_test.py:235: CancelledError
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> entering PDB >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>

>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /home/dev/app/tests/asyncio_test.py(235)go()
-> async def go():
(Pdb) asyncio.all_tasks()
{<Task pending name='Task-1' coro=<Hub.run.<locals>.async_run() running at /home/dev/app/eventlet/hubs/asyncio.py:141> cb=[_run_until_complete_cb() at /usr/local/lib/python3.12/asyncio/base_events.py:181] created at /usr/local/lib/python3.12/asyncio/tasks.py:695>}
(Pdb) future
*** NameError: name 'future' is not defined
(Pdb) u
> /home/dev/app/eventlet/asyncio.py(56)_run()
-> return future.result()
(Pdb) future
<Task cancelled name='Task-17' coro=<test_cancelling_future_kills_greenthread.<locals>.go() done, defined at /home/dev/app/tests/asyncio_test.py:235> created at /usr/local/lib/python3.12/asyncio/tasks.py:695>

```

eventlet/hubs/asyncio.py
```diff
@@ -30,6 +39,18 @@ def __init__(self):
...
         self.sleep_event = asyncio.Event()
 
+    def _remove_bad_fds(self):
+        """ Iterate through fds, removing the ones that are bad per the
+        operating system.
+        """
+        all_fds = list(self.listeners[self.READ]) + list(self.listeners[self.WRITE])
+        for fd in all_fds:
+            try:
+                select.select([fd], [], [], 0)
```
itamarst, line 49: Using `select()` is problematic. It has a fixed max numeric value for file descriptors. So if the max is e.g. 512 and you pass in file descriptor 637, it won't be able to do anything with that descriptor.

To avoid this you want to use `poll()` if available, and only fall back to `select()` if `poll()` is not available (which effectively means Windows).

eventlet/hubs/asyncio.py
```diff
@@ -144,3 +165,37 @@ async def async_run():
                 self.stopping = False
 
         self.loop.run_until_complete(async_run())
+
+    async def wait(self, seconds=None):
```
itamarst, line 169: You're duplicating a whole bunch of `asyncio` logic, and moreover doing it in a way that makes it much less scalable because e.g. you're not using `epoll` on Linux. And the issues with `select()`'s max file descriptor limit. This does not seem like the way to solve this problem.

itamarst (commented):
I think this approach is fundamentally not the way to do it.

itamarst:
I can see if I can implement this a different way.

eventlet/hubs/asyncio.py
itamarst: Or, let me rephrase. You are completely breaking the `asyncio` event loop by blocking. So now any normal `asyncio` code will only be run intermittently.

eventlet/hubs/asyncio.py
4383: oh interesting, I was not aware of that threshold. Indeed poll seems more appropriate.
To be honest, I copy/pasted this logic from the other hubs and I tried to adapt it for asyncio. That mean that all the other hubs are impacted by the limitation of `select`.

eventlet/hubs/asyncio.py
4383: That could explain the failure I observed on the asyncio tests you originally implemented. See the job failure.

4383:
> I can see if I can implement this a different way.

Sure, do not hesitate, you are well far well skilled than me on asyncio things. I think we need this file descriptor management, but I surely implemented it in a bad way. So do not hesitate to go wild! ;)

itamarst:
I think we decided not to go down this route.

PR #926, opened 2024-03-06
Allow creating eventlet backdoor with any eventlet based running process

4383:
This patch propose to introduce a new way to create an eventlet backdoor on any eventlet based running process.

This patch bind the UNIX SIGUSR1 signal with an internal signal handler who, if called, will create an eventlet backdoor.

After this creation, the freshly created backdoor is listening on localhost:3000, and then accessible by using telnet.

Example:

```
$ telnet localhost 3000
Trying 127.0.0.1...
Connection failed: Connection refused
Trying ::1...
telnet: Unable to connect to remote host: Cannot assign requested address
$ kill -SIGUSR1 <pid-of-our-running-process>
$ telnet localhost 3000
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
Python 3.12.2 (main, Feb 13 2024, 09:38:19) [GCC 10.2.1 20210110] on linux
(InteractiveConsole)
>>>
```


4383:
Can be tested by using the following Python script:

```python
import eventlet

eventlet.monkey_patch()


def main():
    c = 0
    while True:
        eventlet.sleep(1)
        print(c)
        c += 1


if __name__ == "__main__":
    main()
```

And then running it by using tox:

```
$ .tox/py312-asyncio/bin/python test.py &
$ kill -SIGUSR1 $! # $! the last pid or replace it manually by the pid of our running process
$ telnet localhost 3000
```

4383:
I added a couple of unit tests and I also added the capability to disabling backdoor by sending `SIGUSR2` signal to the process where the backdoor was created.

The unit tests reflects backdoor enabling and backdoor disabling.

4383:
I think these changes could be a good addition in conjunction of epbf/bcc and other debugging tools, especially to debug our new asyncio based hub, and also older ones.

itamarst:
I am uncomfortable with this being on by default. It makes it much easier for anyone who gains access to a server to get control of the process and e.g. steal credentials.

4383:
Well, you need to be able to send signal to user process to do launch the backdoor, so, I if your able to send system signal to a running process, you already have lot of way to do what sniff or steal things.

Only users with the right system privileges would be able to trigger that backdoor, an attacker would have to escalate to is system level to trigger that backdoor and steal things. At this point it would be already possible for him to run gdb or even some ebpf kernel tracing depending on the user right running the process (sudoers etc).

itamarst:
You can lock down the system a lot (disable ptrace(), which means no gdb), and this undoes some of that locking down. In particular I am thinking of credentials that are in the process memory.

4383:
> You can lock down the system a lot (disable ptrace(), which means no gdb), and this undoes some of that locking down. In particular I am thinking of credentials that are in the process memory.

Indeed that's seems not possible to lock down calling signal to some context as other process may use it to communicate between them. In parallel, the backdoor would be useful in a bug context, and if the process need to be restarted to launch this backdoor, then the bug would be lost. Users would have to pray to face this bug again and catch up this bug again...

Another option would be to implement some sort of port knocking with a protected sequence, who if emitted, would trigger the backdoor. This kind of feature would surely require a parallel thread or something like that and would be more secure than calling a signal. However, if we run a thread to catch port knocking, then a backdoor could be ran at start too...

Let's abandon this idea. Thanks for your previous thoughts.

PR #934, opened 2024-03-13
Run tests on macOS in CI, and some fixes to get it in reasonable state

itamarst:
I filed #935 as eventual follow-up to get a few tests actually passing, but this PR is an improvement on the status quo, at least.


itamarst:
This should probably be merged before #930

PR #938, opened 2024-03-14
Make sure asyncio hub doesn't use greendns for asyncio DNS APIs

### LINKED ISSUE #936: Make sure asyncio hub + dnspython work correctly together
asyncio will try to run DNS lookups in a thread pool, probably, which won't work well if DNS has been monkeypatched. Possibly DNS monkeypatching should just be disabled if the asyncio hub is in use.

itamarst:
Fixes #936

### DIFF
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -41,6 +41,11 @@ class Hub(hub.BaseHub):
         import selectors
         selectors.select = original("select")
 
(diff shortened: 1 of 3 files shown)


itamarst:
Lacking the monkeypatching it does infinite recursion, but apparently that doesn't always happen given we have examples that work :shrug: Anyway this should fix it.

4383 (approved):

PR #939, opened 2024-03-14
Add asyncio API docs

itamarst:
Apparently I forgot to check this file in...


4383 (approved):
LGTM

This page was automatically referenced inside the reference API section [1], but adding it manually into our handcrafted table of contents also make sense.

[1] https://eventlet.readthedocs.io/en/latest/reference/api/eventlet.html#module-eventlet.asyncio

Issue #868, opened 2024-01-03
Asyncio integration for eventlet


itamarst:
### A migration path off of eventlet for libraries/applications that exclusively rely on eventlet

Eventlet has different users with different use cases. This issue covers one specific subset of users: those who deliberately choose eventlet as their exclusive networking/event loop solution. A notable and relevant example is OpenStack.

My basic assumption is that the monkey-patching approach used by eventlet is fundamentally non-viable without massive engineering resources. For example, consider all the effort being put into patching RLock, when patching it actually _introduces_ bugs (e.g. https://bugs.python.org/issue13697). A correct solution would involve maintaining a custom C RLock, and C Queue, and so on. And the more the Python stdlib changes, the harder eventlet maintenance becomes.

As such, I believe eventlet-only users are likely better off switching to event-driven frameworks like asyncio.

### Access to asyncio libraries

Even if you disagree with the above, or you're a user who doesn't fit the above criteria, access to asyncio libraries in eventlet code is useful. Eventlet used to have Twisted support for this reason.

In general I don't think new features should be added to eventlet (see #835), but given that this aids with migration I think it's worth it.

I also believe it's easier to do this inside of eventlet rather than as an external project (see https://github.com/eventlet/aiohub/issues/2).

## The work involved

Prior art for gevent suggests this all should be possible. Each of these should be done as its own issue.

* [ ] Support running eventlet on the asyncio event loop.
* [ ] Support `await`ing GreenThread in an `async def` context.
* [ ] Support calling `async def` functions from inside greenlets.

## Questions

@temoto @tipabu what is your opinion on this approach of integrating asyncio and eventlet?

Issue #835, opened 2023-12-14
Decide on maintenance goals, and document status for users


itamarst:
I can imagine different maintenance goals for the project going forward:

* **Legacy:** Maintenance for stability and bug fixing, no new features, discourage usage in new projects.
* **Maintenance:** Maintenance for stability and bug fixing, no new features, new projects by new users are fine.
* **Active:** Also add new features and enhancements, not just maintenance.

My _personal_ feeling is that eventlet should be in legacy mode, but the maintainers as a group should decide on a level and then document it so users know what to expect.

4383:
As discussed on the [Openstack side](https://review.opendev.org/c/openstack/governance/+/902585) I proposed to smoothly led eventlet to retirement in the coming years, so my personal is:

- move to legacy mode
- introduce aiohub as an official repo of the eventlet orga https://github.com/4383/aiohub/ . I was think propose to move aiohub into the eventlet orga early in 2024 and focus first on urgent things on eventlet. aiohub would be an official guideline to all eventlet users to transition smoothly to asyncio.
- officially retire eventlet in 2-3 years, unless in the meantime someone wants to continue eventlet.

4383:
The aiohub repo is ready for development and I was thinking inspire from [aioeventlet](https://pypi.org/project/aioeventlet/#files) and from gevent-asyncio to implement aiohub. I think that both projects are solid rock foundations for aiohub.

IMO aiohub should be an official response, from the eventlet orga, concerning the future of eventlet.

4383:
aiohub would how we discourage the usage of eventlet in new projects and how we allow to migrate the existing ones.

4383:
I moved aiohub into the eventlet orga https://github.com/eventlet/aiohub

As we are closes from the end of year celebrations, I propose to let this topic opened until early next year (2024), to allow people to react. If nobody else will replied, then, we could consider moving officially to the legacy mode.

Issue #824, opened 2023-12-08
Eventlet is at a dead end. We want to help you get out of it!


4383:
Hi Eventlet maintainers,

We are aware that, currently, eventlet face a cruel lack of human resources. This is a growing issue since 2-3 years.

Lack of resources is a recurring problem in the economy of the open source projects [1]. Maintainers of open source projects have bills to pay and a private life. We, users of open source projects, must thanks hundred of time these maintainers, who, on their free time,  carry out all these small projects which make all technology around us run on a daily basis.

The result of all this is a significant drop in activity or even a abandonment of maintenance of certain projects, due to lack of time, money, or motivation. All these reasons are legit.

Eventlet is currently in this specific case.

Eventlet is in danger. Its health and its future are threatened. The health and future of projects that rely on eventlet are threatened.

A significant part of the python ecosystem is currently in a cold sweat about its own future. Many developers see a cloud of sleepless nights looming on the horizon of their evening. Others have a painful keyboard from biting their fingers until they bleed.

Almost all Python users all around the world already used eventlet at least once, probably without even knowing it. Indeed, Eventlet alone represents more than 62 million downloads. Eventlet is downloaded more than 1,300 million times every month.

https://www.pepy.tech/projects/eventlet

More than 32 thousand projects are based on eventlet, and among them we can cite Openstack, Celery, PgAdmin, numerous Flask and Django add-ons, etc...

https://github.com/eventlet/eventlet/network/dependents

If Eventlet collapses then entire parts of new technologies could collapse with it.

Growing compatibility issues between eventlet and recent versions of CPython are reaching a critical point. de facto, these problems threaten all users of this library. 

- https://github.com/eventlet/eventlet/pull/817
- https://github.com/eventlet/eventlet/pull/820
- https://github.com/eventlet/eventlet/pull/823

Openstack is the first victim of these issues:

https://lists.openstack.org/archives/list/openstack-discuss@lists.openstack.org/thread/YO5CZDVAJ6QSF734ALWSGNOQDDAIOXKI/#RZFUTBFTUCSHKVA6SOCWWIXEH3QJHMYT

We, developers from the Openstack community, offer to provide you with our help and support. We propose to come with you and strengthen your workforce and help you get eventlet back up and running for the next versions of Python to come.

[...]

PR #945, opened 2024-03-25
[DNM][deprecation] deprecate non asyncio hubs

4383:
In accordance with our recent maintenance goals we sets [1][2] and in accordance with the recent addition of the asyncio hub [3], this patch proposes to deprecate all the other existing hubs, the non asyncio hubs, to encourage users to start their migration to asyncio.

[1] https://github.com/eventlet/eventlet/issues/835
[2] https://github.com/eventlet/eventlet/issues/824
[3] https://github.com/eventlet/eventlet/issues/868

### DIFF
--- a/doc/source/hubs.rst
+++ b/doc/source/hubs.rst
@@ -12,15 +12,16 @@ Eventlet has multiple hub implementations, and when you start using it, it tries
     | By using this hub, Asyncio and Eventlet can be run the same thread in the same process.
     | We discourage new Eventlet projects.
     | We encourage existing Eventlet projects to migrate from Eventlet to Asyncio.
+    | We encourage using this hub at runtime.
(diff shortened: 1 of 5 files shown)


itamarst:
My concern is whether we think the asyncio hub is actually in working condition, since we've just fixed a bunch of major bugs in it.

4383:
I was thinking to propose to holds this PR. Maybe we can hold it until as long as the work on the devstack integration is not completed.

https://review.opendev.org/c/openstack/devstack/+/914108

We'll see more clearly once this work is finished.

4383:
Updated the title with "do not merge".

4383:
Related to:

- https://github.com/eventlet/eventlet/issues/824
- https://github.com/eventlet/eventlet/issues/835

PR #949, opened 2024-03-26
eventlet.websocket is not always used from eventlet.wsgi, so do not assume `eventlet.set_idle` exists

### LINKED ISSUE #946: eventlet 0.36.0 incompatible with flask-socketio running under gunicorn
Haven't figured out a minimal test case for this yet, but our CI pipelines for https://github.com/roedoejet/g2p/ have all stopped working because of the 0.36.0 release.  We are running a background flask-socketio server with gunicorn and eventlet, and trying to access it via Playwright, which hangs forever when trying to connect to the socket.

The documentation seems to indicate that we should not use eventlet anymore, so we'll do that :) but just to note that the most recent eventlet seems to be particularly ill-behaved.

dhdaines:
In the case where Eventlet's WebSocket implementation is used, for instance, in a Gunicorn worker, it ends up being the case that `WebSocketWSGI` is not running under an Eventlet server, and thus shouldn't assume that the environment contains `eventlet.set_idle` (which is an implementation detail of the Eventlet server and thus quite meaniningless).

This is notably the case when using Flask-SocketIO with Gunicorn as suggested in the documentation: https://flask-socketio.readthedocs.io/en/latest/deployment.html#gunicorn-web-server

It seems quite safe to just avoid trying to call it in this case.  If, for instance, the engineio wrapper (https://github.com/miguelgrinberg/python-engineio/blob/main/src/engineio/async_drivers/eventlet.py#L39) wishes to provide `eventlet.set_idle` for some reason, then it could do so.

Fixes #946

### DIFF
--- a/eventlet/websocket.py
+++ b/eventlet/websocket.py
@@ -132,10 +132,11 @@ class WebSocketWSGI:
                            [('Connection', 'close'), ] + headers)
             return [body]
 
-        # We're ready to switch protocols; flag the connection
-        # as idle to play well with a graceful stop
-        environ['eventlet.set_idle']()
-
+        # We're ready to switch protocols; if running under Eventlet
+        # (this is not always the case) then flag the connection as
+        # idle to play well with a graceful stop
+        if 'eventlet.set_idle' in environ:
+            environ['eventlet.set_idle']()
         try:
             self.handler(ws)
         except OSError as e:


4383 (approved):
Concerning myself, this patch LGTM.

I'd appreciate to have a secondary opinion from Tim.

4383:
Well, I think these changes are well isolated and should not trigger side effects. Let's go ahead to fix the latest release.
I'm going to prepare a new release.

Thanks @dhdaines for your patch,for your time and for the interesting discussion. Much appreciated.

PR #958, opened 2024-04-24
doc: Repair changelog link

skieffer:
Currently, the link to the changelog found in the Eventlet docs at https://eventlet.readthedocs.io/en/latest/index.html#changelog links to the *Python* changelog, at https://docs.python.org/3/whatsnew/changelog.html#changelog.

This PR changes it to point to https://github.com/eventlet/eventlet/blob/master/NEWS

### Other possibilities

I noticed there's a reference to a `changelog.html` here:

https://github.com/eventlet/eventlet/blob/c22b896473fc38864a3ace488ddc4b0514177cae/doc/source/real_index.html#L151

so maybe there was some intention to generate a true changelog page, belonging to the docs? Don't know.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -113,7 +113,7 @@ Changelog
 =========
 
 For further details about released versions of Eventlet please take a
-look at the :ref:`changelog`.
+look at the `changelog`_.
 
 Authors & History
 =================
@@ -130,4 +130,7 @@ Indices and tables
 * :ref:`genindex`
 * :ref:`modindex`
 * :ref:`search`
-* :ref:`changelog`
+* `changelog`_
+
+
+.. _changelog: https://github.com/eventlet/eventlet/blob/master/NEWS


4383:
Good catch, thanks.

I wonder if rather than pointing out an url, if it would be possible to refer to a file, I know this possible but I always forget the syntax to use... a kink of thing like this:

```
:ref:`../../NEWS`
```

skieffer:
> I wonder if rather than pointing out an url, if it would be possible to refer to a file

There's
```
:doc:`../../some/other/file`
```

for referring to other pages within the Sphinx project, but the `NEWS` file lives up outside of `doc/source`, so as far as I know that wouldn't work.

Unless your build script involves a step that copies the `NEWS` file into `doc/source` before building, it seems the URL might be the easiest solution.

4383:
> > I wonder if rather than pointing out an url, if it would be possible to refer to a file
> 
> There's
> 
> ```
> :doc:`../../some/other/file`
> ```
> 
> for referring to other pages within the Sphinx project, but the `NEWS` file lives up outside of `doc/source`, so as far as I know that wouldn't work.

Indeed, I just tested it locally and it finished with a display of the passed url...
Let's drop that idea.

> 
> Unless your build script involves a step that copies the `NEWS` file into `doc/source` before building, it seems the URL might be the easiest solution.

No I don't think we want to spend more time on this topic, using the url will do the job.

> 
> Also: Linking directly to the github URL means people get to see the most up to date version.

Indeed, thanks.

4383 (approved):

4383:
Thank you for your contribution. Much appreciated

skieffer:
Thanks for the review, and for your work on this project!

PR #959, opened 2024-05-02
drop header keys with underscores

kevin-mizu:
Headers which contain `_` shouldn't be allowed as they could create confusion in incoming requests, allowing, for example, the usage of `CONTENT_LENGTH` instead of `Content-Length`.

This has already been fixed in various WSGI/proxy servers:

- [Nginx](http://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers)
- [Django](https://www.djangoproject.com/weblog/2015/jan/13/security/)
- [Gunicorn](https://github.com/benoitc/gunicorn/commit/72b8970dbf2bf3444eb2e8b12aeff1a3d5922a9a)
- [Werkzeug](https://github.com/pallets/werkzeug/commit/5ee439a692dc4474e0311de2496b567eed2d02cf)
- ...

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -748,6 +748,9 @@ class HttpProtocol(BaseHTTPServer.BaseHTTPRequestHandler):
 
         env['headers_raw'] = headers_raw = tuple((k, v.strip(' \t\n\r')) for k, v in headers)
         for k, v in headers_raw:
+            if "_" in k:
+                continue
+
             k = k.replace('-', '_').upper()
             if k in ('CONTENT_TYPE', 'CONTENT_LENGTH'):
                 # These do not get the HTTP_ prefix and were handled above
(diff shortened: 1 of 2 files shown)


4383:
Thank you @kevin-mizu

eventlet/wsgi.py
```diff
@@ -748,6 +748,9 @@ def get_environ(self):
 
         env['headers_raw'] = headers_raw = tuple((k, v.strip(' \t\n\r')) for k, v in headers)
         for k, v in headers_raw:
+            if "_" in k:
```
4383, line 751: I'd suggest either to internalize the treatment of this condition and of the following replacement (from line 751 to line  754) into a dedicated function with a an explicit naming (e.g `formalize_key_naming`) and with an explicit docstring explain the purpose of doing this, or to add a comment just before the condition you added here.

I think that without more explanations than that, at the implementation level, many people could wonders why this condition exist.

4383:
Also, this patch brokes unit tests. I think unit testing need some adjustment to become compatible with this new rule.

4383 (approved):

4383:
Do you plan to declare a new CVE?

kevin-mizu:
I will try to start the process this afternoon :D

4383:
Awesome, thank you.
Lets merge this patch.

PR #967, opened 2024-06-10
Drop support of Python 3.7

4383:
Python 3.7 is not supported since almost 1 year, Debian Buster LTS is ending on 30th June 2024, and github actions only support versions of Python which are not EOL [3].

For all these reasons, I think we can drop the support of Python 3.7.

[1] https://devguide.python.org/versions/
[2] https://www.debian.org/releases/buster/
[3] https://github.com/actions/python-versions?tab=readme-ov-file#support-policy


4383:
Lets drop it at the end of June

.github/workflows/test.yaml
```diff
@@ -29,7 +29,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { py: 3.7, toxenv: py37-epolls, ignore-error: false, os: ubuntu-latest }
```
kajinamit, line 32: L47 needs to be removed, too

kajinamit (changes requested):
We can probably even remove Python 3.8 because it already reached its EOL, but we can do it later.

.github/workflows/test.yaml
4383: Good catch, done. Thanks

4383:
> We can probably even remove Python 3.8 because it already reached its EOL, but we can do it later.

I'm not against removing python 3.8, maybe with a follow up patch early next year.

itamarst (approved):

PR #970, opened 2024-07-02
Upgrade RLocks as last thing we do

### LINKED ISSUE #969: monkey_patch() fails to green all rlocks
Right now greening of locks happens _before_ modules are patched.

Unfortunately the patching process can trigger imports of additional modules, which can then create more RLocks that never get greened.

This is can be observed in OpenStack's ironic when running on Python 3.12; the logging module is somehow imported during the module-patching process, so its lock never gets greened.

itamarst:
Fixes #969 

Also hide an upgradable RLock in some edge cases.

I tried writing a test but completely failed at getting it to reproduce when run via pytest (it always passes regardless), even though it does reproduce when run manually :cry: So I didn't check it in.

### DIFF
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -274,8 +274,8 @@ def monkey_patch(**on):
             on.setdefault(modname, False)
         on.setdefault(modname, default_on)
 
-    if on['thread'] and not already_patched.get('thread'):
-        _green_existing_locks()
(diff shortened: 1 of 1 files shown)


4383 (approved):

4383:
Feel free to merge it if you think you are done

PR #971, opened 2024-07-03
Describes Eventlet's Alternatives Into the Migration Guide

4383:
This is the beginning of the implementation of the migration guide proposed here:
    
https://review.opendev.org/c/openstack/governance/+/902585

The goal is to centralize all the materials we have to help users to safely migrate off of Eventlet.
It document patterns and methodology.

It will be followed by other pull requests that will reflect the other points made into the proposal.

### DIFF
--- a/doc/source/asyncio/migration.rst
+++ b/doc/source/asyncio/migration.rst
@@ -104,12 +104,15 @@ your application or library to ``asyncio``.
 Calls to blocking APIs like ``urlopen()`` or ``requests.get()`` can get
 replaced with calls to ``aiohttp``, for example.
 
+Depending on your Eventlet usage, during your migration, you may have to
+deprecate CLI options that are related to Eventlet, we invite the reader
+to take a look to :ref:`manage-your-deprecations`.
+
 The `awesome-asyncio <https://github.com/timofurrer/awesome-asyncio>`_ github
-repository propose a curated list of awesome
-Python asyncio frameworks, libraries, software and resources. Do not hesitate
-to take a look at it. You may find
-candidates compatible with asyncio that can allow you to replace some of your
-actual underlying libraries.
+repository propose a curated list of awesome Python asyncio frameworks,
+libraries, software and resources. Do not hesitate to take a look at it.
+You may find candidates compatible with asyncio that can allow you to replace
+some of your actual underlying libraries.
 
 Step 3. Drop Eventlet altogether
 --------------------------------
--- a/doc/source/asyncio/migration.rst
+++ b/doc/source/asyncio/migration.rst
@@ -5,29 +5,43 @@ Migrating off of Eventlet
 
 There are two main use cases for Eventlet:
 
-1. As a required networking framework, much like one would use ``asyncio``, ``trio``, or older frameworks like ``Twisted`` and ``tornado``.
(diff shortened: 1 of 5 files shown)


4383:
related to #972

doc/source/asyncio/guide/pattern.rst
```diff
@@ -0,0 +1,252 @@
...
+
+You can observe that we switched the `User-Agent` header depending on the
+client used by the end users. It will allow you to observe the difference if
+you decide to run this example.
+
+If people want to use the asynchronous client, then their code will be
+dotted with `async` and `await` keywords, making their asynchronous calls
+explicit, not like with Eventlet.
+
+If people want to use the synchronous client, then their code will remains
```
itamarst, line 157: Yes, but, this is _different_ than using Eventlet, because it blocks, which means you need to use regular threads. The blocking version is no longer usable in Eventlet.

doc/source/asyncio/guide/pattern.rst
```diff
@@ -0,0 +1,252 @@
+.. _asyncio-facade-pattern:
+
+AsyncIO Facade Pattern
+======================
+
+This section present a code pattern that you can use to migrate your
```
itamarst, line 6: Is the implied assumption is that you drop Eventlet altogether? If so you need to say so.

Also you switch from "migrate your application" to talking about libraries, those are very different use cases.

itamarst (changes requested):
Much of this is fine, but the facade pattern document seems either incomplete or very misleading, it's unclear what it's proposing and it's unclear if people reading it will interpret it consistently.

I have a vague sense of what it's aiming at, _maybe_ ("here's how to make a library that works with both asyncio and blocking threads"), but I could be wrong! And even if I'm right, it leaves out key points ("you'll have to drop eventlet completely" and "this is for libraries only").

4383:
Thanks @itamarst for your previous review, I'll rework my copy

4383:
I also amended our glossary with new definitions. 
I think their places are legit in this document and I think they can be useful in the eventlet context.

itamarst:
Explaining that the facade is useful _after_ completely migrating off of Eventlet makes sense, yeah. "You're no longer using Eventlet at all, how do you have libraries that provide both async and non-async APIs"

doc/source/asyncio/migration.rst
```diff
@@ -5,40 +5,57 @@ Migrating off of Eventlet
...
-**Most of this document will focus on the first use case: Eventlet as the sole networking framework.**
-For this use case, we recommend migrating to Python's ``asyncio``, and we are providing infrastructure that will make this much easier, and allow for *gradual* migration.
+**Most of this document will focus on the first use case: Eventlet as the sole
+networking framework.**
+For this use case, we recommend migrating to Python's ``asyncio``, and we are
+providing infrastructure that will make this much easier, and allow for
+*gradual* migration.
 
-For the second use case, we believe this is a fundamentally unsustainable approach and encourage the upstream frameworks to come up with different solutions.
+For the second use case, we believe this is a fundamentally unsustainable
```
itamarst, line 27: The existence of `awaitlet` suggests we should direct people there, maybe?

itamarst (approved):

PR #975, opened 2024-07-13
os.read/write waits until file descriptor is ready.

### LINKED ISSUE #634: eventlet with Flask/SocketIO in debug mode with watchdog installed hangs on startup
It appears that having the `watchdog` package installed caused eventlet to hang while starting a Flask/Flask-SocketIO server.

I am not referencing this package in the source code for the server. It's only installed for another process.

Turning off Flask debug in `socketio.run` allows the server to start.
Not calling `eventlet.monkey_patch()` allows the server to start.
Uninstalling `watchdog` allows the server to start.

I don't know how watchdog causes the issue, and I'm totally unsure of where to report such a defect. In theory, if the defect is simple, anybody with a Flask/Flask-SocketIO server using eventlet can install `watchdog` and check for the same behaviour.

The behaviour is that the server locks, doesn't server requests, and doesn't reload.  I suspect it's something with the file watching, since watchdog also watches files. But how it interferes, I don't know.

### LINKED ISSUE #973: os.read() blocks even after monkey patching.
This is likely the root cause behind #634.

I thought that monkey patching the os module means all blocking functions are converted into non-blocking functions. However, while using Watchdog, I found that `os.read()` still block. Simple example using watchdog:
```py
import eventlet
from eventlet.greenthread import spawn, sleep

eventlet.monkey_patch()

import logging
from watchdog.observers import Observer
from watchdog.events import LoggingEventHandler

increment = 0


def heartbeat():
    global increment
    while True:
        print('second:', increment)
        increment += 1
        sleep(1)


if __name__ == "__main__":
    spawn(heartbeat)
    sleep(2)

    logging.basicConfig(level=logging.INFO,
                        format='%(asctime)s - %(message)s',
                        datefmt='%Y-%m-%d %H:%M:%S')
    event_handler = LoggingEventHandler()
    observer = Observer(generate_full_events=True)
    observer.schedule(event_handler, '/home/ethan/test', recursive=True)
    observer.start()
    try:
        while observer.is_alive():
            observer.join(1)
    finally:
        observer.stop()
        observer.join()
```

I expect it to print out time stamps forever ("seconds: 0", "seconds: 1", "seconds: 2", "seconds: 3" ...), but once Watchdog starts, it stops printing because Watchdog blocks on the main thread.

The issue is that Watchdog spawns a new thread and reads from the INotify API. The read calls are executed through `os.read(fd, size)` here:
https://github.com/gorakhargosh/watchdog/blob/9f23b599f369cdd873c0cf3dd9e69567234d78fb/src/watchdog/observers/inotify_c.py#L302

Of course the read calls are redirected to the monkey patched version:
https://github.com/eventlet/eventlet/blob/06ec82896ebb9a26edaf6e1ad4d63393990f15b7/eventlet/green/os.py#L36-L51

I believe the issue is that the monkey patched method immediately calls the original read function, which blocks because no output is available yet, before it calls `trampoline`. If I change the order of method calls to call `trampoline` first, it works fine:
```py
def read(fd, n):
    while True:
        try:
            hubs.trampoline(fd, read=True)
        except hubs.IOClosed:
            return ''

        try:
            return __original_read__(fd, n)
        except OSError as e:
            if get_errno(e) == errno.EPIPE:
                return ''
            if get_errno(e) != errno.EAGAIN:
                raise
```

It makes more sense to me that Eventlet waits until data is available before calling a potentially blocking function. Is this a bug with Eventlet? If not, is `os.read` designed to be blocking (even after being monkey patched)?

ethan-vanderheijden:
Fixes #973 
Fixes #634 

Currently, calling `os.read(fd, size)` will attempt to read the file descriptor immediately. If there is no data to be read, the thread will block until data is available, causing the program to hang. With this pull request, `os.read` will wait until the file descriptor is ready to read before attempting a read. It will not wait if the file descriptor is a regular file (according to select/poll, regular files are always ready to read).

My fear is that `os.write` has the same bug:
https://github.com/eventlet/eventlet/blob/06ec82896ebb9a26edaf6e1ad4d63393990f15b7/eventlet/green/os.py#L57-L68

It's not common for a write call to block, but it can happen, and if it does happen, the entire program will hang. Do you want me to investigate and fix this as well?

### DIFF
--- a/eventlet/green/os.py
+++ b/eventlet/green/os.py
@@ -1,6 +1,7 @@
 os_orig = __import__("os")
 import errno
 socket = __import__("socket")
+from stat import S_ISREG
 
 from eventlet import greenio
 from eventlet.support import get_errno
@@ -38,6 +39,15 @@ def read(fd, n):
 
     Read a file descriptor."""
     while True:
+        # don't wait to read for regular files
+        # select/poll will always return True while epoll will simply crash
+        st_mode = os_orig.stat(fd).st_mode
+        if not S_ISREG(st_mode):
+            try:
+                hubs.trampoline(fd, read=True)
+            except hubs.IOClosed:
+                return ''
+
         try:
             return __original_read__(fd, n)
         except OSError as e:
@@ -45,10 +55,6 @@ def read(fd, n):
                 return ''
             if get_errno(e) != errno.EAGAIN:
                 raise
-        try:
-            hubs.trampoline(fd, read=True)
-        except hubs.IOClosed:
-            return ''
 
 
 __original_write__ = os_orig.write
--- a/eventlet/green/os.py
+++ b/eventlet/green/os.py
@@ -66,12 +66,20 @@ def write(fd, st):
     Write a string to a file descriptor.
     """
     while True:
+        # don't wait to write for regular files
+        # select/poll will always return True while epoll will simply crash
+        st_mode = os_orig.stat(fd).st_mode
+        if not S_ISREG(st_mode):
+            try:
+                hubs.trampoline(fd, write=True)
+            except hubs.IOClosed:
+                return 0
+
         try:
             return __original_write__(fd, st)
         except OSError as e:
             if get_errno(e) not in [errno.EAGAIN, errno.EPIPE]:
                 raise
-        hubs.trampoline(fd, write=True)
 
 
 def wait():
--- a/tests/isolated/os_read_nonblocking.py
+++ b/tests/isolated/os_read_nonblocking.py
@@ -1,0 +1,30 @@
+if __name__ == '__main__':
+    import eventlet
+    from eventlet.greenthread import sleep, spawn
+
+    eventlet.monkey_patch()
+
+    import signal
+    import os
+
+    thread = None
+    timed_out = False
+
+    def on_timeout(signum, frame):
+        global timed_out
+        timed_out = True
+        thread.kill()
+
+    def blocking_read(fd):
+        os.read(fd, 4096)
+
+    signal.signal(signal.SIGALRM, on_timeout)
+    signal.alarm(3)
+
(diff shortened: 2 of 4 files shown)


4383:
Thanks for this patch.

> It's not common for a write call to block, but it can happen, and if it does happen, the entire program will hang. Do you want me to investigate and fix this as well?

Yes please

4383 (approved):

ethan-vanderheijden:
Ok, I made the same change with `os.write()`. Now, it waits until the file descriptor is ready before attempting a write.

According to the manual (for select, poll, epoll):

> However, even if a file descriptor indicates as writable, a large write may still block.

If someone calls `os.write()` with a single large chunk of data, the program may still hang. The only solution is to mark the file descriptor as non-blocking prior to the write, but that is something the user has to do, not us.

4383 (approved):

4383:
I just  re-enqueued all jobs as some jobs failed during the previous run. No logs were available so I've no clue. I want to be sure than nothing is broken.

4383:
Ok, it was just a flaky CI, lets move on

4383:
@ethan-vanderheijden thanks for your investigations and for your work. Good job!

ethan-vanderheijden:
@4383 Thanks for the quick review!

Sorry to bug you, but when is the next release of Eventlet scheduled?

4383:
@ethan-vanderheijden I'll generate a new version early next week. I need to do an inventory of the latest changes first.

4383:
@ethan-vanderheijden I'm preparing the next release, you can follow its progress there => https://github.com/eventlet/eventlet/issues/979

4383:
@ethan-vanderheijden eventlet 0.37.0 is out! https://pypi.org/project/eventlet/0.37.0/

PR #985, opened 2024-10-03
wsgi: No request Content-Length nor Transfer-Encoding implies no body

tipabu:
Previously, if an application tried to read some set length from such a request (because, say, it always tried to read every request's body), the server could hang waiting on the client to send more bytes.

The spec, however, is clear -- even going back to HTTP/1.0 -- if `Content-Length` (nor `Transfer-Encoding`, in HTTP/1.1) is missing, the request has no body, which is equivalent to having a body of length 0.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -142,8 +142,7 @@ class Input:
             # 100 Continue response
             self.send_hundred_continue_response()
             self.is_hundred_continue_response_sent = True
-        if (self.content_length is not None) and (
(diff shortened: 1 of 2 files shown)


4383:
Please could you put in your commit message a link toward the RFC section that speak about that?
I think it will be helpful in some manner.

tipabu:
Sure thing! Linked to all four that seemed relevant.

4383 (approved):
Thanks

PR #990, opened 2024-11-15
Correct line lookup from inspect.getsourcelines()

stefanor:
I can't write a test for this, as I don't know how to trigger this part of Spew, without also falling into the OSError branch (which `test_line_nofile` tests).

However, it has been observed to hit this code when running the tests under pytest-xdist (execnext), which pokes around in `linecache`, and the code is obviously wrong. This corrects it.

### DIFF
--- a/eventlet/debug.py
+++ b/eventlet/debug.py
@@ -35,8 +35,12 @@ class Spew:
             else:
                 name = '[unknown]'
                 try:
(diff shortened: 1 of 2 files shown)


4383 (approved):

4383:
Thank you, will be land by the next release

PR #995, opened 2024-11-19
Fix bug where asyncio hub didn't support multiple os threads

### LINKED ISSUE #994: asyncio hub doesn't support multiple OS threads
There are no tests for this or we would've noticed by now, but it seems `eventlet` expects different OS threads to be able to run greenlets. There are e.g. tests in `oslo.log` that rely on this.

Unfortunately the `asyncio` hub breaks when this is done, and seems to implicitly assume just one thread.

E.g. this runs on epolls hub:

```python
import time

import eventlet
import eventlet.patcher
eventlet.monkey_patch()

threading_orig = eventlet.patcher.original("threading")

def os_thread_2():
    print("THREAD 2 STARTED", time.time())
    eventlet.sleep(1)
    print("THREAD 2", time.time())
    eventlet.sleep(1)
    print("THREAD 2", time.time())

threading_orig.Thread(target=os_thread_2).start()
print("THREAD 1", time.time())
eventlet.sleep(0.99)
print("THREAD 1", time.time())
eventlet.sleep(3)
```

But when run with asyncio hub:

```
Exception in thread Thread-1 (os_thread_2):
  File "/home/itamarst/devel/eventlet/eventlet/hubs/__init__.py", line 104, in get_hub
    hub = _threadlocal.hub
          ^^^^^^^^^^^^^^^^
AttributeError: '_thread._local' object has no attribute 'hub'. Did you mean: 'Hub'?

During handling of the above exception, another exception occurred:

  File "/home/itamarst/devel/eventlet/eventlet/hubs/__init__.py", line 104, in get_hub
    hub = _threadlocal.hub
          ^^^^^^^^^^^^^^^^
AttributeError: '_thread._local' object has no attribute 'hub'. Did you mean: 'Hub'?

...

Exception ignored in: <function BaseEventLoop.__del__ at 0x7197b4a77880>
  File "/usr/lib/python3.11/asyncio/base_events.py", line 695, in __del__
    self.close()
  File "/usr/lib/python3.11/asyncio/unix_events.py", line 68, in close
    super().close()
  File "/usr/lib/python3.11/asyncio/selector_events.py", line 91, in close
    self._close_self_pipe()
  File "/usr/lib/python3.11/asyncio/selector_events.py", line 98, in _close_self_pipe
    self._remove_reader(self._ssock.fileno())
                        ^^^^^^^^^^^
AttributeError: '_UnixSelectorEventLoop' object has no attribute '_ssock'
```

and other garbage.

itamarst:
Fixes #994 

Apparently we didn't unpatch sufficiently.

### DIFF
--- a/eventlet/hubs/asyncio.py
+++ b/eventlet/hubs/asyncio.py
@@ -2,6 +2,14 @@
 Asyncio-based hub, originally implemented by Miguel Grinberg.
 """
 
+# The various modules involved in asyncio need to call the original, unpatched
+# standard library APIs to work: socket, select, threading, and so on. We
+# therefore don't import them on the module level, since that would involve
+# their imports getting patched, and instead delay importing them as much as
+# possible. Then, we do a little song and dance in Hub.__init__ below so that
+# when they're imported they import the original modules (select, socket, etc)
+# rather than the patched ones.
+
 import os
 import sys
 


eventlet/hubs/asyncio.py
```diff
@@ -32,22 +24,46 @@ class Hub(hub.BaseHub):
...
+            "select",
+            "socket",
+            "os",
+            "threading",
+            "queue"
+        )()
 
         # The presumption is that eventlet is driving the event loop, so we
         # want a new one we control.
+        import asyncio
```
4383, line 71: Cannot we import asyncio at the global level rather than importing it multiple time in this module?

4383: My understanding (if it is correct) is that we have to tweak some stdlib module like you did in this patch, but I wonder if the import of asyncio is by itself required multiple times

itamarst: The important thing is not importing it at the top level. I'll add some docs.

4383: ok, thanks

4383 (approved):

PR #996, opened 2024-11-20
Update Actions

tinyboxvk:
- Bump actions/checkout to v4
- Bump actions/setup-python to v5
- Bump actions/cache to v4
- Fixes `The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3, actions/cache@v3, actions/setup-python@v3.`
- Switch to the stable release of 3.13 in `test.yaml`

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
     - name: Set up Python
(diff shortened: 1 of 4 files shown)


4383 (approved):

4383:
Thank you @tinyboxvk

tinyboxvk:
Mostly this is to update their default runtime to node20 so that they don't generate deprecation warnings. If you go into the details of each workflow run you shouldn't see anything under Annotations. There are of course other fixes and improvements which probably aren't immediately apparent.

4383:
Thanks for your answer

Issue #998, opened 2024-12-02
Python 3.13: threading.Thread.is_alive() wrongly returns True


CyrilRoelandteNovance:
Consider the following piece of code:

```
#import eventlet; eventlet.patcher.monkey_patch()
import queue
import threading


class Worker(threading.Thread):
    EXIT_SENTINEL = object()

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.q = queue.Queue(maxsize=-1)
        self.daemon = True

    def run(self):
        while True:
            task = self.q.get()
            if task == self.EXIT_SENTINEL:
                break
            print(f'Treating task {task}')
            # Pretend to work

    def submit(self, job):
        self.q.put(job)

    def terminate(self):
        self.q.put(self.EXIT_SENTINEL)
        self.join()

worker = Worker()
assert not worker.is_alive()
worker.start()
assert worker.is_alive()
worker.submit(1)
worker.terminate()
assert not worker.is_alive()
```

Here is the output in Python 3.13:

```
$ python rep.py 
Treating task 1
```

If we uncomment the first line in order to use eventlet, this is what we get:

```
$ python rep.py 
Treating task 1
Traceback (most recent call last):
  File "/usr/src/app/rep.py", line 35, in <module>
    assert not worker.is_alive()
           ^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

This is because worker._handle.is_done() still returns False when it should be returning True.

This worked properly in Python 3.12 and started failing in Python 3.13.

4383:
Hey Cyril,

Thanks for opening this issue.

We will have a look ASAP.

4383:
I wonder if these [oslo.messaging errors observed on debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1089050) are not the same than the problem described here.

4383:
@stefanor FYI ^

stefanor:
Doesn't suprise me at all! :)

CyrilRoelandteNovance:
It's definitely the same failure. There are a few different OpenStack projects that use this pattern, so it's going to cause this exact failure multiple times.

4383:
Sorry, I've problems with my laptop so I cannot launch things like I'd love, can you tell me if you still face the error even when `self.daemon = False`?
I wonder if daemon do not need some additional adjustments.

PR #1000, opened 2024-12-09
Python 3.13: Use greenthread's dead state where possible

stefanor:
Fixes: #998

### DIFF
--- a/tests/isolated/patcher_threading_subclass_done.py
+++ b/tests/isolated/patcher_threading_subclass_done.py
@@ -1,0 +1,40 @@
+import queue
+import threading
+
+
+class Worker(threading.Thread):
+    EXIT_SENTINEL = object()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.q = queue.Queue(maxsize=-1)
+        self.daemon = True
+
+    def run(self):
+        while True:
+            task = self.q.get()
+            if task == self.EXIT_SENTINEL:
+                break
+            print(f"Treating task {task}")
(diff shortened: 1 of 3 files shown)


tests/isolated/patcher_threading_subclass_done.py
```diff
@@ -0,0 +1,40 @@
+import queue
+import threading
+
+
+class Worker(threading.Thread):
```
itamarst, line 5: This is defined _before_ monkeypatching, so it creates an OS thread, is that what you intended?

tests/patcher_test.py
```diff
@@ -536,3 +536,7 @@ def test_patcher_existing_locks():
 
 def test_patcher_existing_locks_exception():
     tests.run_isolated("patcher_existing_locks_exception.py")
+
+
+def test_patcher_threading_subclass_done():
```
itamarst, line 541: This is failing in CI.

itamarst (changes requested):
Thank you for submitting this, could you take a look at the comments above?

tests/isolated/patcher_threading_subclass_done.py
stefanor: The Thread class is never monkey-patched. It's the underlying `_thread` methods that are.

tests/patcher_test.py
stefanor: Tsk. That's because dead greenlets evaluate falseish.

I had the code structured differently earlier, when I was testing.

itamarst (approved):
Thank you!

Issue #1003, opened 2024-12-10
Asyncio hub should error out if asyncio has already been imported


itamarst:
In this situation, you get mysterious errors (recursive AttributeError when running a thread, say) because monkeypatching is happening too late, so e.g. asyncio.selector_events ends up with monkeypatched socket and #994 starts happening again, with errors like:

```
Exception in thread Thread-1 (pthread1):
Traceback (most recent call last):
  File "/home/itamarst/devel/oslo.log/.tox/py311/lib/python3.11/site-packages/eventlet/hubs/__init__.py", line 104, in get_hub
    hub = _threadlocal.hub
          ^^^^^^^^^^^^^^^^
AttributeError: '_thread._local' object has no attribute 'hub'. Did you mean: 'Hub'?

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/itamarst/devel/oslo.log/.tox/py311/lib/python3.11/site-packages/eventlet/hubs/__init__.py", line 104, in get_hub
    hub = _threadlocal.hub
          ^^^^^^^^^^^^^^^^
AttributeError: '_thread._local' object has no attribute 'hub'. Did you mean: 'Hub'?

During handling of the above exception, another exception occurred:
...
```

On startup, the hub should check if asyncio etc are already imported, and if they are, error out saying "monkey_patch has to happen earlier."

itamarst:
Alternatively, maybe it should try to monkey patch them anyway, but that's difficult.

PR #1005, opened 2024-12-10
EVENTLET_MONKEYPATCH mechanism

### LINKED ISSUE #1004: It is very difficult to monkey patch early enough, both in applications and testing
My scenario: I am testing `oslo.log`, and I need to monkey patch to ensure that happens early enough. If you don't, `asyncio` gets imported before monkey_patch() is called and everything breaks (#1003). 

Other common scenarios is application calling monkey_patch() too late, and something doesn't get patched.

Proposed solution (user perspective): if user sets `export EVENTLET_MONKEYPATCH=1`, monkey patching happens as early as possible.

Implementation: Use a .pth file. It still might not suffice in all cases (other `.pth` files might also import enough stuff to trigger e.g. #1003 situation) but it'll hopefully help.

itamarst:
Fixes #1004

### DIFF
--- a/doc/source/patching.rst
+++ b/doc/source/patching.rst
@@ -61,6 +61,8 @@ library.  This has the disadvantage of appearing quite magical, but the advantag
          
     It is important to call :func:`~eventlet.patcher.monkey_patch` as early in the lifetime of the application as possible.  Try to do it as one of the first lines in the main module.  The reason for this is that sometimes there is a class that inherits from a class that needs to be greened -- e.g. a class that inherits from socket.socket -- and inheritance is done at import time, so therefore the monkeypatching should happen before the derived class is defined.      It's safe to call monkey_patch multiple times.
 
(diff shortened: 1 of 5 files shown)


itamarst:
(I will submit a fix for #1003 after this is merged.)

doc/source/patching.rst
```diff
@@ -61,6 +61,8 @@ library.  This has the disadvantage of appearing quite magical, but the advantag
          
     It is important to call :func:`~eventlet.patcher.monkey_patch` as early in the lifetime of the application as possible.  Try to do it as one of the first lines in the main module.  The reason for this is that sometimes there is a class that inherits from a class that needs to be greened -- e.g. a class that inherits from socket.socket -- and inheritance is done at import time, so therefore the monkeypatching should happen before the derived class is defined.      It's safe to call monkey_patch multiple times.
 
+    Starting in Eventlet 0.39, you can also set an environment variable ``EVENTLET_MONKEYPATCH=1`` to get patching as early as possible in Python startup, which may avoid some problems.
```
4383, line 64: good idea

4383 (approved):

Issue #1007, opened 2024-12-11
[release] eventlet 0.38.2


4383:
There is a new features, a 1 fix. I propose a minor version (0.39.0).

List of commits:

- 0d5f1208 asyncio hub should error out if asyncio has already been imported (#1006)
- b754135b EVENTLET_MONKEYPATCH mechanism. (#1005)

Global diff between 0.38.1 and master (the coming release 0.39.0):

https://github.com/eventlet/eventlet/compare/0.38.1..master

PR #1008, opened 2024-12-11
Update changelog for version 0.38.2

4383:
Related to #1007

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,12 @@
 Unreleased
 ==========
 
+0.38.2
+======
+
+* [fix] fix the monkey patching with the asyncio hub
+* [feature] introduce the unmonkeypatching feature
+
 0.38.1
 ======
 


itamarst (approved):
Thank you!

4383:
Ok, then let's pause the release process.

NEWS
```diff
@@ -1,6 +1,13 @@
 Unreleased
 ==========
 
+0.39.0
+======
+
+* [fix] fix the monkey patching with the asyncio hub
```
itamarst, line 7: Really it's only the first bullet item, I think.

itamarst: Second one is implementation detail, third one isn't true anymore because I came up with better unmonkeypatching.

NEWS
```diff
@@ -1,6 +1,13 @@
 Unreleased
 ==========
 
+0.39.0
```
itamarst, line 4: Yeah 0.38.1 is more accurate

itamarst (approved):
Couple of comments to address, and then can merge, thank you!

NEWS
4383: 0.38.2 because 0.38.1 is already out

4383:
done

4383:
(I didn't changed the name of the branch used to create the pull request, it still reflect the previous version proposal, but, eh, who care? :smile: )

itamarst:
Looks like 3.7 can't run on Ubuntu 24.04.

PR #1011, opened 2024-12-11
Replace EVENTLET_MONKEYPATCH with more robust un-monkeypatching for asyncio hub

### LINKED ISSUE #1009: New 3.8-specific failures due to detection of asyncio pre-import in asyncio hub
In `tox38-asyncio` run in one PR we've seen:

```
ERROR tests/asyncio_test.py - RuntimeError: asyncio has already been imported before hub creation and monkey patching. Try calling eventlet.monkey_patch() earlier. If that is not possible, use the EVENTLET_MONKEYPATCH=1 env variable instead of eventlet.monkey_patch().
ERROR tests/hub_test.py - RuntimeError: asyncio has already been imported before hub creation and monkey patching. Try calling eventlet.monkey_patch() earlier. If that is not possible, use the EVENTLET_MONKEYPATCH=1 env variable instead of eventlet.monkey_patch().
ERROR tests/not_asyncio_test.py - RuntimeError: asyncio has already been imported before hub creation and monkey patching. Try calling eventlet.monkey_patch() earlier. If that is not possible, use the EVENTLET_MONKEYPATCH=1 env variable instead of eventlet.monkey_patch().
```

### LINKED ISSUE #1010: Maybe don't recommend EVENTLET_MONKEYPATCH except for testing?
The mechanism applies to _any_ Python process you run while the environment variable is set.

So this can include:

1. Random non-eventlet using application. Probably not what you want.
2. Subprocesses. Again probably not what you want.

In addition, it makes applications that rely on it broken if it's _not_ set.

As such:

1. It should probably be unset so subprocesses don't inherit it.
2. It should probably only be used for testing purposes, and not real world applications, which is a doc update.

### LINKED ISSUE #1012: The un-monkeypatching in AsyncioHub might be too aggressive
Context:
1. We want `asyncio` module to use blocking APIs.
2. Therefore, we have a song-and-dance in `AsyncioHub.__init__` to make sure they're loaded with `original("socket")` and friends instead of `socket` and friends.

The problem is that may impact transitive modules (modules imported by `asyncio`) where we want to stick to greened APIs. So maybe some sort of more fine grained unmonkeypatching mechanism should be used, in which case we can maybe:

1. Move it to `monkey_patch()`.
2. Maybe rip out the whole check-for-early-asyncio import thing?

itamarst:
Fixes #1009 
Fixes #1010 
Fixes #1012 

This makes monkeypatching/asyncio hub interact together better: less aggressive and more accurate unmonkeypatching, less frustration with timing of monkey_patch(). E.g. with this you can just use normal slightly-too-late monkey_patch() in oslo.log tests and still have stestr tests run correctly.

Unfortunately I ran black on the file by mistake so there's an extra-large diff, sorry, noticed too late.


itamarst:
I'm wondering if #1012 might be a better approach.

itamarst:
Need to figure out what's up with `kqueue` on macOS and then I think this is good to go for review.

itamarst:
I think it might actually be working again?! seems suspicuous.

eventlet/patcher.py
```diff
@@ -223,6 +230,99 @@ def original(modname):
 already_patched = {}
 
 
+def _unmonkey_patch_asyncio(unmonkeypatch_refs_to_this_module):
```
itamarst, line 233: This is a meaningful change in patcher.py

eventlet/patcher.py
```diff
@@ -355,16 +467,20 @@ def after_fork():
...
     # Issue #508: Since Python 3.7 queue.SimpleQueue is implemented in C,
     # causing a deadlock.  Replace the C implementation with the Python one.
     import queue
+
     queue.SimpleQueue = queue._PySimpleQueue
 
     # Green existing locks _after_ patching modules, since patching modules
     # might involve imports that create new locks:
-    _green_existing_locks(original_rlock_type)
+    for name, _ in modules_to_patch:
```
itamarst, line 481: And this is a meaningful change, and everything else is syntax tweaks, sorry again.

4383 (approved):

PR #1018, opened 2025-01-20
Remove monotonic from requirements

kajinamit:
... because it is required only in Python <= 3.5 which has no overlap with the python versions currently supported.


4383 (approved):

4383:
Thanks

PR #1019, opened 2025-01-20
Remove Python 3.8 support

kajinamit:
Python 3.8 already reached its eol on 2024-10-07 .

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,12 +29,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { py: 3.8, toxenv: py38-epolls, ignore-error: false, os: ubuntu-latest }
-          - { py: 3.8, toxenv: py38-openssl, ignore-error: false, os: ubuntu-latest }
-          - { py: 3.8, toxenv: py38-poll, ignore-error: false, os: ubuntu-latest }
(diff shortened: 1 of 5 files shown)


4383 (approved):

4383 (approved):

4383:
MacOS jobs are failing but I don't think the failure is related to this patch. Can be safely ignored.

4383:
hm, I think there is a problem with gates... py38 jobs are pending and they block this patch, but the problem is that this patch drop these jobs so... the snake bites its tails

kajinamit:
Hmm I tried force-push but these jobs are still required for some reason. I wonder if there are some definitions left in github side (within repository setting ?)

4383:
> Hmm I tried force-push but these jobs are still required for some reason. I wonder if there are some definitions left in github side (within repository setting ?)

No, normally everything is centralized here

4383:
That's not the first time we remove old jobs

4383:
> That's not the first time we remove old jobs

Example https://github.com/eventlet/eventlet/commit/fcc5ce42d979757396e222d2823d27d57985caec

4383:
I wonder if this is a github bug...
I'm not able to merge this patch while the jobs are stuck.
Can you try to create a new PR by duplicating your patch? I think you just have to create and push a new branch on your fork.

PR #1031, opened 2025-03-28
[WIP][Workaround] Skip ident comparison to avoid crash on Python 3.13+

4383:
Python 3.13 introduced major internal changes to thread management as part of PEP 703 (GIL removal), including native thread handles and joinable threads. These changes modified how thread identifiers are defined and used in the standard library, breaking compatibility with Eventlet's monkey-patched threading model.

The issue decribed in #1030 is related to the GIL and PEP 703 because Python 3.13 introduced internal thread handles and joinable threads to support the no-GIL mode, changing how thread identifiers are managed: instead of simple values, they are now internal native identifiers handled at the C level, tied to OS-level threads or new CPython thread structures.

Since Eventlet is deeply tied to green threads and emulates standard threads using greenlet-based constructs, properly adapting to Python 3.13 would require a near-complete rewrite of its threading logic.

As a temporary workaround, this commit skips the comparison of thread identifiers to avoid AttributeErrors and restore partial functionality under Python 3.13+.

A full fix will need a deeper architectural review.

Related to #1030

### DIFF
--- a/eventlet/green/thread.py
+++ b/eventlet/green/thread.py
@@ -127,13 +127,6 @@ def start_new_thread(function, args=(), kwargs=None):
 start_new = start_new_thread
 
 
-def _get_main_thread_ident():
-    greenthread = greenlet.getcurrent()
-    while greenthread.parent is not None:
-        greenthread = greenthread.parent
-    return get_ident(greenthread)
-
-
 def allocate_lock(*a):
     return LockType(1)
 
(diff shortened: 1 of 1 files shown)


4383:
Sure, will do that.

itamarst:
I have come up with a more fundamental solution for #1030, at least.

PR #1034, opened 2025-04-22
Fix ssl test when linking against openssl 3.5+

yankov-pt:
Openssl 3.5 changed some error propagations. As a result, sometimes instead of SSL_Error we get a Broken Pipe. Update the ssl_test to handle that.

### DIFF
--- a/tests/ssl_test.py
+++ b/tests/ssl_test.py
@@ -80,7 +80,8 @@ class SSLTest(tests.LimitedTestCase):
             sock.recv(8192)
             try:
                 self.assertEqual(b'', sock.recv(8192))
-            except greenio.SSL.ZeroReturnError:
+            except (greenio.SSL.ZeroReturnError,
+                    BrokenPipeError):
                 pass
 
         sock = listen_ssl_socket()


4383:
Hello,

Thanks for reporting this issue.
Is there a way to reproduce the behaviour you observed?

yankov-pt:
We discovered this when trying to upgrade openssl to 3.5 in openSUSE Tubmleweed. 

Here's a a build log without the patch: [
[eventlet-build-fail.log](https://github.com/user-attachments/files/19848416/eventlet-build-fail.log)
](url)

Here's a log with the patch applied and all tests passing: https://build.opensuse.org/package/live_build_log/devel:languages:python/python-eventlet/openSUSE_Tumbleweed/x86_64

4383:
Thanks for details.

4383 (approved):

4383:
Hm, apparently we face problems with the pypy jobs. This jobs seems canceled, I don't know why and no logs are available. I need to dig further to see what happen, because this job is required, and so it block our merging.

paulwouters:
confirmed on our end as well. This fixes the tests on Fedora Rawhide that uses openssl-3.5.0

4383:
I made several jobs relaunch attempts and the pypy jobs is still failing without real reason.
I created a merge request to make the pypy job optional https://github.com/eventlet/eventlet/pull/1035 and hence being able to merge your patch.

4383:
Unfortunatelly #1035 is not enough...
We have to disable this canceled job at the repository settings level, but for now we are not granted to do that.

PR #1036, opened 2025-05-07
Move pypy job to ubuntu-24.04

gboutry:
This instance of the job matrix is running on ubuntu 20.04 which has been actively discontinued by github: actions/runner-images#11101

PR #1035, opened 2025-04-28
make pypy jobs optional

4383:
Pypy job is failing without real reasons and logs are not useful. The CI just say: " The operation was canceled."

Lets ignore that job as we are going to abandon eventlet and we have urgent fix to merge before.

PR #1021, opened 2025-01-20
Remove Python 3.8 support

kajinamit:
Python 3.8 already reached its eol on 2024-10-07 .

PR #1037, opened 2025-05-12
Discontinues PyPy support

4383:
PyPy broke our CI and avoid us merging important patches. See:
- #1036 
- #1035 
- #1034
- #1031 
- etc

We are not PyPy users, and we have no enough resources to spend days maintaining compatibility with PyPy. We want to focus our time on our use cases.

Persons interested to continue maintaining the PyPy support are welcomed, and we could surely consider reintroducing its support if people volunteer to maintain it.

For now our priority is to ensure that we are able to offer a stable migration to our users, so we first have to ensure that we remain safe.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -81,19 +81,6 @@ jobs:
       - name: codecov
         run: codecov --flags=$(echo ${{ matrix.toxenv }} |tr -d -- '-.')
 
-  # Optional tests
-  optional-tests:
-    name: "Optional Test: ${{ matrix.toxenv }}"
-    continue-on-error: true
-    runs-on: ${{ matrix.os }}
-    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != 'eventlet/eventlet'
-    timeout-minutes: 10
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - { py: pypy3.9, toxenv: pypy3-epolls, os: ubuntu-20.04 }
-
     steps:
       - name: install system packages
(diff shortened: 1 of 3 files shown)


itamarst (approved):

itamarst:
Yeah seems fine. Note you can change required CI builds in Settings for the repo. I don't have access but maybe you do @4383 ?

4383:
No, unfortunately I do not have the rights to access and configure that, this is why I proposed to drop this job. 

We tried to repair it through #1036, but then SSL errors appeared with PyPy, and I do not think we want to waste more time with this topic.

itamarst:
Dropping the job won't suffice, it'll still be required in the Settings.

4383:
Are you sure? 
This job is not anymore listed in the jobs ran within this pull request: https://github.com/eventlet/eventlet/actions/runs/14977253183

4383:
My understanding is that github actions only run the jobs that are defined in the files under the `.github/workflows` directory.

4383:
> It won't run the job but I expect it to still _require_ it (see the "1 expected" in the UI above).

Yeah sorry, I was on my mobile and the mobile version did not rendered that the same way than on the desktop version.
I see Sergey forced the push of my previous proposal, I rebased this patch on that, in all the cases I would suggest to continue with the removal of that job, as said Jay:
> It's hard to imagine anyone who cares enough about performance to be using pypy but not enough to migrate to something more modern than eventlet."

@itamarst Please can you approve again, my latest rebase make stale your previous vote.

4383:
I propose to merge first #1021 and then, once done, I'll update the branch and manage the coming conflicts and then we will be able to review/approve this patch again

itamarst (approved):

4383:
Thanks you very much

PR #1038, opened 2025-05-13
Bump mysql version in tests

kajinamit:
... because 5.7 reached its eol in 2023.


4383 (approved):
Thank you Takashi, much appreciated

Issue #1039, opened 2025-05-13
New release 0.40.0


4383:
There is no new features, 1 fix, few docs updates, and the removal of Python 3.8' supports. I propose a minor version (0.40.0) to try to reflect the removal of Python 3.8 as far as possible.

List of commits:

- https://github.com/eventlet/eventlet/commit/0f59c833678c770dffd9ef89d5f924fa6860ef3a
- https://github.com/eventlet/eventlet/commit/bd1cabb65d4da07c8964e17255cda6e22233d38
- https://github.com/eventlet/eventlet/commit/5092d4e0e4768bab97776f0fbdc77b1fe1976522
- https://github.com/eventlet/eventlet/commit/da893c86d9695e5a7aa0cab7f287a9a7bf0a359a
- https://github.com/eventlet/eventlet/commit/3f0bb49b3a1aded1534de3c9ca9a0e94036f3684

Global diff between 0.39.1 and master (the coming release 0.40.0):

https://github.com/eventlet/eventlet/compare/0.39.1..master

4383:
Released: 
- https://pypi.org/project/eventlet/0.40.0/
- https://github.com/eventlet/eventlet/releases/tag/0.40.0

Thank you everyone!

PR #1041, opened 2025-05-13
Update changelog for version 0.40.0

4383:
Related to #1039

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,14 @@
 Unreleased
 ==========
 
+0.40.0
+======
+
+* [fix] Fix ssl test when linking against openssl 3.5 (#1034)
+* Drop support Python 3.8 (#1021)
+* [doc] Various doc updates (#981, #1033)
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,6 +1,14 @@
 Unreleased
 ==========
 
+0.40.0
+======
+
+* [fix] Fix ssl test when linking against openssl 3.5 (#1034)
+* Drop support Python 3.8 (#1021)
+* [doc] Various doc updates (#981, #1033)
+* [env] Make PyPy job optional (#1035)
```
itamarst, line 10: Should probably be "dropped PyPy support"?

itamarst (changes requested):

NEWS
4383: Indeed, I can rebase over it now that the patch is merged. Let's do that

4383: done

itamarst (approved):

4383:
@itamarst Thank you

PR #1046, opened 2025-06-09
Replace deprecated datetime.datetime.utcnow

kajinamit:
The utcnow function was deprecated in Python 3.12[1].

Note that all datetime instances are kept non-timezone-aware to keep backword-compatibility.

[1] https://docs.python.org/3.13/library/datetime.html#datetime.datetime.utcnow


4383 (approved):
LGTM, thanks Takashi

PR #1050, opened 2025-06-17
Replace deprecated datetime.utcfromtimestamp

kajinamit:
It was deprecated in Python 3.12 in favor of datetime.fromtimestamp[1].

[1] https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp


4383 (approved):
LGTM, thanks Takashi

PR #1053, opened 2025-06-18
Fix patching of removed URLopener class in Python 3.14

allhailwesttexas:
The class `urllib.request.URLopener` has been removed in Python 3.14: https://github.com/python/cpython/pull/125739

Because of this, trying to run the tests in `test__greenness.py` fails with the following on Python 3.14:

```
ERROR tests/test__greenness.py - NameError: name 'URLopener' is not defined
```

Update the patching so that we only do it for Python versions lower than 3.14.


4383 (approved):
Good catch, thanks

PR #1054, opened 2025-06-30
Prepare jobs and CI/CD for python 3.14

4383:
Jobs are non-voting. The goal is simply to prepare the machinery to be ready for python 3.14, and hence, allowing us to test eventlet against this version, observe issues, and fix them early if possible.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,6 +9,7 @@ jobs:
   required-tests:
     name: "Required Tests: ${{ matrix.toxenv }}"
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.ignore-error || false }}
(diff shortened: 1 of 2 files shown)


itamarst:
I've updated it so the tests actually run on 3.14 on Linux (macOS is still a mystery to me, it seems like it _ought_ to work?!). So the CI errors are now more meaningful for 3.14.

itamarst (approved):
Would be nice to figure out why macOS is failing to run, but not strictly necessary for progress on 3.14 support.

itamarst:
Nope :cry:  Maybe you can review it?

4383:
Nope, unfortunately I'm still blocked. Does not matter, I'll recreate this PR from a new branch tomorrow, it should fix our problem.

4383:
I mean, even my review is blocked, I can just comment, that's all.

PR #1055, opened 2025-07-02
Prepare jobs and CI/CD for python 3.14

4383:
Jobs are non-voting. The goal is simply to prepare the machinery to be ready for python 3.14, and hence, allowing us to test eventlet against this version, observe issues, and fix them early if possible.

Also updates versions of pyzmq and pyscopg-binary to work with 3.13+


4383:
New version of #1054

itamarst (approved):

4383:
@itamarst weird, notice that I'm now able to merge... you are surely right, we should probably drop that restriction.

sins921 (approved):

PR #1056, opened 2025-07-07
Make database removal safer with IF EXISTS

4383:
Use the SQL clause `DROP DATABASE IF EXISTS` instead of `DROP DATABASE` to avoid errors when the target database does not exist. This change improves robustness in test cleanup for MySQL and PostgreSQL connection pools.


itamarst (approved):

4383:
Thanks @itamarst

PR #1058, opened 2025-07-14
Fix tests on Python 3.14 on Linux

### LINKED ISSUE #1057: Tests fail on Python 3.14 on Linux

itamarst:
Fixes #1057

### DIFF
--- a/eventlet/greenpool.py
+++ b/eventlet/greenpool.py
@@ -91,9 +91,7 @@ class GreenPool:
                 if DEBUG:
                     traceback.print_exc()
         finally:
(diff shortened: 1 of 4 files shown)


eventlet/greenpool.py
```diff
@@ -91,9 +91,7 @@ def _spawn_n_impl(self, func, args, kwargs, coro):
                 if DEBUG:
                     traceback.print_exc()
         finally:
-            if coro is None:
```
itamarst, line 94: `return` in `finally` is a problem, apparently.

Issue #1063, opened 2025-08-11
New release 0.40.3


4383:
There is no new features, and only 1 security fixes. I propose a bug fix version (0.40.3).

List of commits:

- https://github.com/eventlet/eventlet/commit/0bfebd1117d392559e25b4bfbfcc941754de88fb

Global diff between 0.40.2 and master (the coming release 0.40.3):

https://github.com/eventlet/eventlet/compare/0.40.2..master

4383:
Git tag: https://github.com/eventlet/eventlet/releases/tag/0.40.3
Pypi release: https://pypi.org/project/eventlet/0.40.3/

Thanks everyone

PR #1064, opened 2025-08-11
Update changelog for version 0.40.3

4383:
Related to #1063


4383:
@itamarst no problem I was on vacation too. I do not want to waste your time with useless reviews, but on the hand sometime that's useful to have a double check just in case, so, as you prefer....

PR #1067, opened 2025-09-08
Fix Python 3.14 on macOS

itamarst:
Tests being green is blocked on #1065 (just needs to be merged) and #1068 (someone needs to figure what is going on). But this does fix the problem with 3.14 specifically on macOS.


4383:
> The failure on 3.14 has switched from "not running at all" to running and failing one test (see #1068), so this is an improvement.

\o/

4383:
Weird

Issue #1071, opened 2025-10-09
Missing sub-modules when building w/ setuptools


clayg:
When I build an sdist from master with a clean .egg-info using the latest setuptools in a venv I get a very anemic `.tar.gz`

https://gist.github.com/clayg/694ec219ab3a0cecdd99bb5de3cf47d2

prior to !921 when I would run `python setup.py sdist` all of the eventlet sub-modules would be in the package.

I'm not 100% what @thomasgoirand  was going for with !921 - in my read of the [relevant setuptools docs](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html) I would have expected something more like:

```
[tool.setuptools]
packages = ['eventlet', 'eventlet.zipkin', 'eventlet.support', 'eventlet.greenio', 'eventlet.green', 'eventlet.hubs', ...]
```

???

I *suspect* eventlet isn't actually trying to support setuptools builds anymore !845 - since the pyproject.toml clearly states the build-system is hatch and setuptools doesn't seem to understand the `dynamic = ['version']` directive correctly.  But I guess I'm not the only packager who's build system is still stuck creating new packages to fix CVEs for old distros and isn't ready to let go of setuptools yet?

Two options that would be so great for me:

1) revert !921 - not telling setuptools anything beyond what's already declared in generic/default `[project]` and `options.*` sections in pyproject.toml seems to work totally fine for building a working sdist .tar.gz files with old and new setuptools alike!  Even if setuptools is "best effort" - why ruin the party!?

2) better understand the original problem with the debian directories getting mixed up with the source pkg_dir !???  I use rpm/spec files and my builds' setuptools do not seem to reproduce that problem/confusion.  I wonder if an alternative spelling would have have worked for debian:

```
-[tool.setuptools]
-packages = ['eventlet']
+[tool.setuptools.package-dir]
+eventlet = "eventlet"
```

... but also like how is the debian builds even working?  Are they doing something like:

```
python eventlet/setup.py sdist
```

and then somehow digging the resulting `eventlet/eventlet` out of the `unknown-0.0.0` tarball!?!  So confused...

PR #1072, opened 2025-10-10
Remove legacy setuptools configuration files

4383:
Removes setup.py and setup.cfg as they contain redundant configuration that duplicates what's already properly defined in pyproject.toml with hatchling build backend. The setup.py only contained an empty setuptools.setup() call, and setup.cfg only specified description_file which is already configured as readme in pyproject.toml.

Also removes obsolete [tool.setuptools] and [options.packages.find] sections from pyproject.toml since hatchling handles package discovery automatically for standard project layouts.

This modernizes the packaging setup to use only pyproject.toml standards.

Related to: #1071


4383:
@itamarst: Hey, would you mind reviewing this patch? Thanks in advance.

PR #1077, opened 2025-12-08
More visible deprecation

### LINKED ISSUE #1076: More visible deprecation
https://sethmlarson.dev/deprecations-via-warnings-dont-work-for-python-libraries suggests DeprecationWarning doesn't work because it's hidden by default for any module other than `__main__`, and suggests a custom warning class.

itamarst:
Fixes #1076 

Turns out DeprecationWarning is only visible if eventlet is imported in `__main__`. If it's imported in a secondary module, it won't show up. This makes sure it will show up in both cases.

### DIFF
--- a/eventlet/__init__.py
+++ b/eventlet/__init__.py
@@ -98,6 +98,11 @@ framework.  For more detail see
 https://eventlet.readthedocs.io/en/latest/asyncio/migration.html
 """
 
+class EventletDeprecationWarning(Warning):
+    """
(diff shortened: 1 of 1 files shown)


itamarst:
Apparently I'm not authorized to merge this? I don't understand.

4383:
> Apparently I'm not authorized to merge this? I don't understand.

I think you face the same problem than me weeks ago... I merged it.

PR #1079, opened 2026-02-26
fix: replace 45 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.

### DIFF
--- a/benchmarks/context.py
+++ b/benchmarks/context.py
@@ -67,7 +67,7 @@ try:
     start = time.time()
     test_eventlet()
     print("epoll:     %.02f seconds" % (time.time() - start))
-except:
+except Exception:
     print("epoll hub unavailable")
 
 try:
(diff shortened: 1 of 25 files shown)


4383:
Hey, thanks for this patch. At first glance I've no objection, but I'd love to heard a bit more about your reasons that motive this change, have you experienced some issues?

haosenwang1018:
Formatting fix for my previous comment (shell escaped badly) — re-posting clearly:\n\nThanks for taking a look and great question. The motivation is safety/debuggability: bare except also catches KeyboardInterrupt, SystemExit, and GeneratorExit, which can hide control-flow signals and make failures harder to diagnose.\n\nThis patch narrows those handlers to except Exception so we still catch normal runtime errors while letting system-level exceptions propagate as intended. I targeted only bare-except sites and kept behavior otherwise unchanged.\n\nIf you prefer, I can further reduce scope to a smaller subset (for example one module per PR) to make review/rollback even easier.

haosenwang1018:
😄 Nice one. Back to the PR: if it helps review, I can split this into smaller follow-up PRs by module.\n\nFor this change, the concrete goal is to replace bare except with except Exception so runtime errors are still handled but control-flow/system exceptions (KeyboardInterrupt/SystemExit/GeneratorExit) are not swallowed.

Issue #1087, opened 2026-03-16
New release 0.41.0


itamarst:
In particular #1083 is required for the poor unfortunate souls who are using `fork()` without exec on 3.14.1+.

4383:
- tag created: https://github.com/eventlet/eventlet/releases/tag/0.41.0
- pypi package created: https://pypi.org/project/eventlet/
Thanks

Issue #1083, opened 2026-03-11
fork() breaks eventlet in the child process, as of Python 3.14.1


itamarst:
Probably due to https://github.com/python/cpython/pull/139897/changes#diff-b5b7e4f5599916b2018e911ef423f9d3fd46cb71268e1a73c74a1c9ae2cb6791 some of our tests for "working" post-fork() have started failing.

PR #1088, opened 2026-04-01
Update changelog for version 0.41.0

4383:
related to #1087


itamarst (approved):
You should also add #1083

PR #1090, opened 2026-05-10
Reduce warning in import

kajinamit:
... and add migrate the "noisy" warning to actual patching.

Showing warning during import is problematic, rather than beneficial. Any projects importing eventlet indirectly now sees these warnings. We can suppress the warning by adding a filter, but we can't filter the specific warning because the warning class requires us to import the 'eventlet' module, which emits warning.

### DIFF
--- a/eventlet/__init__.py
+++ b/eventlet/__init__.py
@@ -105,4 +105,4 @@ class EventletDeprecationWarning(Warning):
 
 # If we're running tests this adds extra output that messes up some assertions.
 if os.environ.get("EVENTLET_TESTS") is None:
-    warnings.warn(_DEPRECATED, EventletDeprecationWarning, stacklevel=2)
+    warnings.warn(_DEPRECATED, DeprecationWarning, stacklevel=2)
--- a/eventlet/patcher.py
+++ b/eventlet/patcher.py
@@ -6,8 +6,10 @@ except ImportError:
     import imp
 import asyncio as asyncio_module
 import importlib
+import os
 import pkgutil
 import sys
+import warnings
 
 try:
     # Only for this purpose, it's irrelevant if `os` was already patched.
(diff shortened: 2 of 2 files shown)


itamarst:
Why is it a problem to have this warning? We want people to see it and stop using eventlet.

itamarst:
Possibly we should make the warning more strongly-worded, something about Eventlet being "an architectural dead-end and unfortunately a likely source of bugs"?

kajinamit:
I understand your point, though I've seen multiple job failures in OpenStack, due to the warning output breaking json output and the parse logic. This behavior is caused by python-openstackclient which indirectly import eventlet (via oslo.concurrency, which is a common lock library in openstack).

Because we can't drop that eventlet import logic immediately and still have to maintain eventlet support for a while, this requires us to add warning filters to any possible import logic (in oslo.concurrency and any other common libraries which can import eventlet if exists) but I find it causing undesired effort honestly speaking.

itamarst:
Another option is something like `os.environ["I_KNOW_EVENTLET_IS_BROKEN"] = "1"; import eventlet`, i.e. add another user-facing env variable like `EVENTLET_TESTS`?

kajinamit:
I'm afraid that's not still a good solution. I'm unsure if it can make any sense to ask people using openstackclient/openstacksdk to export I_KNOW_EVENTLET_IS_BROKEN to silence the warning.

itamarst:
I was thinking `oslo.concurrency` could set that? Flip is that hides the warning for all libraries/apps that use `oslo.concurrency`.

kajinamit:
> I was thinking `oslo.concurrency` could set that? Flip is that hides the warning for all libraries/apps that use `oslo.concurrency`.

Technically yes, but I'm more concerned with immediately breaking software which "indirectly" imports eventlet (The initial reason I'm looking into this issue is broken ansible playbooks failing to parse openstackclient output due to that warning garbage). openstacksdk and python-openstackclient are just examples and there could be more software affected by the same mechanism.

PR #1091, opened 2026-05-10
Bump actions dependent on nodejs 20

kajinamit:
... and replace these by the new versions dependent on nodejs 24.


4383 (approved):
LGTM, we can safely ignore the MacOS jobs

4383:
Thanks

PR #1092, opened 2026-05-21
wsgi: Handle ENOTCONN

### LINKED ISSUE #943: wsgi server should handle ENOTCONN error
cPython's ssl.py has merged a [fix](https://github.com/python/cpython/commit/264b1dacc67346efa0933d1e63f622676e0ed96b) for [CVE-2023-40217](https://github.com/advisories/GHSA-4j9r-82g6-9mj3) several months ago. When a client closes the connection before SSL handshake is finished, and there's still data in the buffer, the server would raise an SSLError whose errno is ENOTCONN to prevent any potential malicious action from the client.

However, this can happend on some normal curcumstances (for example, client disconnects on connection timeout). When it happens, sock.accept() would raise an error that is not handled by wsgi server, thus cause the server to exit.

```python
try:
    serv.log.info('({}) wsgi starting up on {}'.format(serv.pid, socket_repr(sock)))
    while is_accepting:
        try:
            client_socket, client_addr = sock.accept()           #  1. an SSLError(errno.ENOTCONN) will be raised
            client_socket.settimeout(serv.socket_timeout)
            serv.log.debug('({}) accepted {!r}'.format(serv.pid, client_addr))
            connections[client_addr] = connection = [client_addr, client_socket, STATE_IDLE]
            (pool.spawn(serv.process_request, connection)
                .link(_clean_connection, connection))
        except ACCEPT_EXCEPTIONS as e:
            if support.get_errno(e) not in ACCEPT_ERRNO:      # 2. the errno is not handled here
                raise
            else:
                break
finally:
    for cs in connections.values():
        prev_state = cs[2]
        cs[2] = STATE_CLOSE
        if prev_state == STATE_IDLE:
            greenio.shutdown_safe(cs[1])
    pool.waitall()
    serv.log.info('({}) wsgi exited, is_accepting={}'.format(serv.pid, is_accepting))     # 3. wsgi server is exited
```

Callum027:
`ENOTCONN` is sometimes raised when client connections are unexpectedly closed. The fix for [CVE-2023-40217](https://github.com/python/cpython/issues/108310) included in more recent versions of CPython also results in `ENOTCONN` being raised when an SSL connection is closed by the client while data remains on the buffer. These errors cause the eventlet WSGI server to terminate unexpectedly.

Add `ENOTCONN` to the list of accepted errnos to ensure this does not happen.

Fixes #943.

### DIFF
--- a/eventlet/wsgi.py
+++ b/eventlet/wsgi.py
@@ -924,11 +924,12 @@ Please use server.log.info instead.''')
 try:
     import ssl
     ACCEPT_EXCEPTIONS = (socket.error, ssl.SSLError)
-    ACCEPT_ERRNO = {errno.EPIPE, errno.ECONNRESET,
+    ACCEPT_ERRNO = {errno.EPIPE, errno.ECONNRESET, errno.ENOTCONN,
                     errno.ESHUTDOWN, ssl.SSL_ERROR_EOF, ssl.SSL_ERROR_SSL}
 except ImportError:
     ACCEPT_EXCEPTIONS = (socket.error,)
-    ACCEPT_ERRNO = {errno.EPIPE, errno.ECONNRESET, errno.ESHUTDOWN}
+    ACCEPT_ERRNO = {errno.EPIPE, errno.ECONNRESET, errno.ENOTCONN,
+                    errno.ESHUTDOWN}
 
 
 def socket_repr(sock):


itamarst:
Seems reasonable at first glance.

Callum027:
The failure for `py314-epolls` *appears* to be unrelated, is that failing intermittently?

itamarst:
On macOS we were getting failures in general; on Linux, we shouldn't be (but might be unrelated, e.g. change in  a CPython patch release, haven't looked).

4383 (approved):
At first glance LGTM, now we need to see why it is failing only with the Py314 epoll job

4383:
The py314-epolls failure is unrelated to this PR's ENOTCONN change.

I reproduced the failure locally, and both hubs (asyncio & epolls) produce the same error, but CI shows py314-asyncio passed.

The issue is in tests/isolated/patcher_existing_logging_module_lock.py. It asserts that before monkey patching, logging._lock and handler.lock are NOT threading._PyRLock, and after patching they ARE. Something changed in Python 3.14 that breaks this assumption.

The root cause seems to be: During Python interpreter shutdown, logging's _removeHandlerRef weakref callback fires and tries to acquire a lock via with lock:. Since monkey patching converted that lock to a green lock, the RLock.acquire() calls get_ident() which calls greenlet.getcurrent(). At that point, greenlet is already being finalized by the interpreter, so it raises RuntimeError: greenlet is being finalized.

The traceback output appears after "pass" in the subprocess, so lines[-1] is no longer b'pass' and the test assertion fails.

The call chain:
logging._removeHandlerRef (weakref callback during shutdown)
  → threading.RLock.acquire (green lock)
    → eventlet.green.thread.get_ident
      → greenlet.getcurrent()  ← RuntimeError: greenlet is being finalized

The fix should go in `eventlet/green/thread.py:get_ident`, it needs to catch `RuntimeError` during greenlet finalization and fall back to `_thread.get_ident()`.

I'm gonna to merge this patch and to propose a fix to this rlock problem.

PR #1093, opened 2026-06-08
Fix 4 DNS tests that fail without /etc/resolv.conf

### LINKED ISSUE #963: 4 unit tests depend on the presence of /etc/resolv.conf
Hi,

Under Debian buildd servers, the /etc/resolv.conf isn't present anymore since a few days. Therefore, the below unit tests are failing. What should be done, IMO, is mocking the /etc/resolv.conf instead of depending on the system version.

See this bug for details:
https://bugs.debian.org/1070333

In the Debian package, I've blacklisted these tests to solve the problem, however, I'd prefer to run them:
    - test_clear
    - test_noraise_dns_tcp
    - test_raise_dns_tcp
    - test_dns_methods_are_green

Cheers,

Thomas Goirand (zigo)

4383:
## Summary
- Fix 4 unit tests (`test_clear`, `test_raise_dns_tcp`, `test_noraise_dns_tcp`, `test_dns_methods_are_green`) that depend on `/etc/resolv.conf` being present
- On systems without `/etc/resolv.conf` (e.g. Debian buildd servers), `dns.resolver.Resolver()` raises `NoResolverConfiguration`, causing these tests to fail

## Changes
- **`test_clear`**: use a temporary resolv.conf instead of the system default
- **`test_raise_dns_tcp` / `test_noraise_dns_tcp`**: use `Resolver(configure=False)` since these tests already set their own nameservers manually
- **`socket_resolve_green.py`**: provide a fallback temporary resolv.conf when the system file is missing

## Test plan
- [x] All 4 affected tests pass with `tox -e py314-selects`

Closes #963

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/greendns_test.py
+++ b/tests/greendns_test.py
@@ -299,13 +299,17 @@ class TestUdp(tests.LimitedTestCase):
 class TestProxyResolver(tests.LimitedTestCase):
 
     def test_clear(self):
-        rp = greendns.ResolverProxy()
+        tmp = tempfile.NamedTemporaryFile(mode='w', suffix='.conf')
+        tmp.write('nameserver 127.0.0.1\n')
+        tmp.flush()
+        rp = greendns.ResolverProxy(filename=tmp.name)
         assert rp._cached_resolver is None
         resolver = rp._resolver
         assert resolver is not None
         rp.clear()
         assert rp._resolver is not None
         assert rp._resolver != resolver
+        tmp.close()
 
     def _make_mock_hostsresolver(self):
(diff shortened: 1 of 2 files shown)


4383:
Require merging https://github.com/eventlet/eventlet/pull/1096 first

4383:
@itamarst o/ I think we can now continue with this patch

4383:
@itamarst Please can you have a look to this patch? Sorry to harass you with that.

itamarst (approved):

PR #1098, opened 2026-06-26
pin greenlet to 3.4 to temporary unlock the CI

4383:


itamarst:
Trying less than 3.5

itamarst:
Seems to work.

4383:
indeed. I cannot approve this patch, and I think you too, so we won't be able to merge it. Do you mind proposing your change via a new PR? I'll approve it

4383:
@itamarst see my previous comment

PR #1100, opened 2026-07-15
Fix bare except clauses in eventlet/greenpool.py

koteshyelamati:
Bare `except:` clauses catch `BaseException` (including `SystemExit`, `KeyboardInterrupt`, and `GreenletExit`), which is almost never the intent. This PR fixes two such clauses in `eventlet/greenpool.py`:

1. **Exception logger in `_spawn_n_impl`** (after the explicit `except (KeyboardInterrupt, SystemExit, greenlet.GreenletExit): raise` guard): changed `except:` → `except Exception:` so only non-fatal exceptions are logged.
2. **Spawn cleanup in `GreenPile`**: the counter-decrement and re-raise handler changed `except:` → `except Exception:` to avoid silently swallowing fatal signals.

### DIFF
--- a/eventlet/greenpool.py
+++ b/eventlet/greenpool.py
@@ -87,7 +87,7 @@ class GreenPool:
                 func(*args, **kwargs)
             except (KeyboardInterrupt, SystemExit, greenlet.GreenletExit):
                 raise
-            except:
+            except Exception:
                 if DEBUG:
                     traceback.print_exc()
         finally:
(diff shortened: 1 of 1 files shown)


koteshyelamati:
Hi @itamarst — thanks for looking. Could you share why this was closed? I'm guessing that `GreenletExit` (a `BaseException` subclass) needs to propagate through the greenpool handlers to properly terminate greenlets, similar to how `huey`'s gevent backend needs bare `except:` for greenlet lifecycle management. If the bare `except:` here is intentional for that reason, happy to close this formally and note the rationale for the record.

itamarst:
It didn't seem like you bothered to run the tests, the tests failed in CI, and this project is in maintenance-only mode so we're only applying minimal effort to maintain it.

itamarst:
Oh and the AI vibes from your user account in general.

PR #1101, opened 2026-07-29
Fix compatibility with greenlet >= 3.4

4383:
Since greenlet 3.5.1, greenlet.getcurrent() raises RuntimeError during interpreter shutdown instead of returning None. This broke eventlet's monkeypatched code paths that run during finalization: Python's logging module triggers weakref callbacks during shutdown that acquire locks, which flow through eventlet's patched threading.Lock (Semaphore) and thread.get_ident — both of which call greenlet.getcurrent().

Guard get_ident() and Semaphore.acquire() against this RuntimeError. In get_ident(), fall back to the real _thread.get_ident() captured at import time (before monkeypatching). In Semaphore.acquire(), do a simple non-blocking counter decrement since interpreter finalization is single-threaded.

Remove the greenlet < 3.4 version pin that was working around this.

### DIFF
--- a/eventlet/green/thread.py
+++ b/eventlet/green/thread.py
@@ -6,6 +6,8 @@ from eventlet.timeout import with_timeout
 from eventlet.lock import Lock
 import sys
 
+_real_get_ident = __thread.get_ident
+
 
 __patched__ = ['Lock', 'LockType', '_ThreadHandle', '_count',
                '_get_main_thread_ident', '_local', '_make_thread_handle',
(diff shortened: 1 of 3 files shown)


4383:
Would require merging first #1102

4383:
I confirm that #1102 is fixing the gates

4383:
@itamarst as the CI is now fixed and this patch rebased over #1102 I think we can continue with this one.

itamarst (approved):

itamarst:
Thank you!

PR #1105, opened 2026-09-08
Raise a clear error for non-latin-1 WSGI response headers

### LINKED ISSUE #1061: UTF-8 in headers is not handled correctly
It seems like UTF-8 content in response headers causes an error due to the [following code](https://github.com/eventlet/eventlet/blob/master/eventlet/wsgi.py#L529):

```python3
for header in response_headers:
    towrite.append(('%s: %s\r\n' % header).encode('latin-1'))
```

I can't find a spec saying that response header **values** have to be Latin-1, I'm not the only one for whom it [does not seem to be specified](https://www.jmix.io/blog/utf-8-in-http-headers/). In the case of H2 [the RFC says](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2): "Just as in HTTP/1.x, header field names are strings of ASCII characters".

This leads me to believe that the values should not be forcefully encoded as `Latin-1`, valid UTF-8 should pass through unmodified. Alternatively, for ease-of-use, eventlet should %-encode any UTF-8 values instead of erroring out like this.

gyanu2507:
WSGI response headers were encoded as latin-1, so anything outside that set raised a bare `UnicodeEncodeError`.

This still encodes name and value as latin-1 (PEP 3333). They are encoded separately, so a non-latin-1 value cannot drag the name out as UTF-8. If the value is outside latin-1, the error now says to use RFC 8187 or percent-encoding instead of sending raw UTF-8 (RFC 9110 field values are visible ASCII).

Fixes #1061


4383:
This tackles a real problem: the `UnicodeEncodeError` from non-latin-1 header values is genuinely unhelpful and confusing for developers. But silently falling back to UTF-8 introduces a different kind of problem: it sends bytes on the wire that violate both the WSGI spec (PEP 3333 explicitly mandates latin-1 encoding for response headers on Python 3) and the `HTTP/1.1` spec (RFC 9110 restricts header values to visible ASCII plus SP/HTAB). The developer won't know anything went wrong until a proxy, CDN, or client downstream silently drops or garbles the header in production.

There's also a subtlety in how the fallback is structured. ` _encode_header_line` encodes the entire Name: `Value\r\n` line as a single unit, so if a header name happens to contain non-latin-1 characters, that  would also silently get sent as UTF-8 rather than raising an error. Header names must be ASCII under every version of the spec.

The test validates the raw bytes on the wire using a socket, which confirms the encoding works mechanically, but it doesn't tell us whether an actual HTTP client can parse that response correctly. Many won't.

I think a better path forward would be either raising a clearer error that explains what went wrong and points developers toward standard solutions (RFC 8187 encoding or percent-encoding), or automatically percent-encoding the value so it stays within the ASCII-safe range. If the team does want a UTF-8 fallback for pragmatic reasons, it should at minimum emit a warning log so operators and developers can see it happening and fix their application code rather than unknowingly depending on non-standard behavior.

gyanu2507:
Dropped the UTF-8 fallback.

Name and value encode as latin-1 on their own. A value outside that set raises `UnicodeEncodeError` and points at RFC 8187 / percent-encoding. A header name that is not latin-1 fails the same way instead of leaving as UTF-8.

4383 (approved):

4383:
That's much better, thank you!

Avamander:
No, HTTP/1.1 does not limit the value to ASCII. The BEP might.

For HTTP/1.1 the ABNF allows `VCHAR / obs-text`.

